PR #6, opened 2023-04-27
User types

goens:
I translated `Kind`s into parametric `UserType` inductives as we discussed; Lean gets pretty picky about which instance you're talking about, even with `variable`s, so the code at points becomes more messy, but in principle we should get the same just parametric on this `BaseType` with a typeclass that says we can translate to a Lean `Type`

This essentially competes with the `TypeSemantics` of Chris, where we encode the user types concretely as `Nat`s. With the instances we have in the end we could translate everything into these semantics only "under the hood".

There's an issue remaining where I couldn't figure out what was happening in the `op` case of `inferType` and sorried it now. If we agree that this approach makes sense, we can try and figure out what happened here.

### DIFF
--- a/SSA/WellTypedFramework.lean
+++ b/SSA/WellTypedFramework.lean
@@ -1,10 +1,14 @@
 import SSA.Framework
 import Mathlib.Data.Option.Basic
 
-class BaseTypeClass (BaseType : Type)  : Type 1 where
+/-
+Typeclass for a `baseType` which is a godel code of 
+Lean types.
+-/
+class Godel (BaseType : Type)  : Type 1 where
   toType : BaseType → Type
 
-instance : BaseTypeClass Unit where toType := fun _ => Unit
+instance : Godel Unit where toType := fun _ => Unit
 
 inductive UserType (T : Type) : Type where
   | base : T → UserType T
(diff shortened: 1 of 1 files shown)


SSA/WellTypedFramework.lean
```diff
@@ -1,91 +1,128 @@
...
+def SSAIndex.ieval (Op : Type)
+    [TypedUserSemantics Op BaseType] : SSAIndex → Type
+  | .TERMINATOR => TypeData BaseType
+  | .EXPR => TypeData BaseType
+  | .REGION => TypeData BaseType × TypeData BaseType
   | .STMT => List Var
 
-abbrev K : Type → Type := StateT ((Var → TypeData) × (Var → (TypeData × TypeData))) Option
-
+abbrev K : Type → Type := StateT ((Var → TypeData BaseType) × (Var → (TypeData  BaseType × TypeData BaseType))) Option
```
bollu, line 105: I suggest parameterizing K by `BaseType`. That should allow Lean to figure out what we mean, as then `BaseType` becomes an explicit argument (instead of an implicit as it is right now)

bollu:
@goens review, please

SSA/WellTypedFramework.lean
goens: no octopus emoji in github :(

SSA/WellTypedFramework.lean
```diff
@@ -1,57 +1,60 @@
...
-structure UserData {BaseType : Type} [DecidableEq BaseType] [BaseTypeClass BaseType] where
+structure UserData {BaseType : Type} [BaseTypeClass BaseType] where
   type : UserType BaseType
   value : type.toType
 
 
-inductive  TypeData (BaseType : Type) [DecidableEq BaseType] [BaseTypeClass BaseType] : Type
+-- @goens: I suggest removing [DecidableEq ...] [BaseTypeClass ...]
+-- We do not keep typeclass constraints on definitions, only on use sites.
+-- See the Haskell rationale:
```
goens: Interesting! My rationale was that while `DecidableEq` is technically unnecessary for `BaseTypeClass`, it's morally what we want, same for `TypeData`, we morally want this only for user types, that's why I put them here. I think you forgot to put the link to the Haskell rationale? I'd be curious to understand why.

SSA/WellTypedFramework.lean
```diff
@@ -1,10 +1,14 @@
 import SSA.Framework
 import Mathlib.Data.Option.Basic
 
-class BaseTypeClass (BaseType : Type)  : Type 1 where
+/-
+Typeclass for a `baseType` which is a godel code of
```
goens, line 5: I'm happy with having a better name than `BaseTypeClass` for this! Honestly, almost any name does it. I'm curious why Gödel though? It's not quite the same as the Gödel encoding of statements as numbers, which is the only analogy that fits here I know of.

SSA/WellTypedFramework.lean
```diff
@@ -5,8 +5,11 @@ import Mathlib.Data.Option.Basic
 Typeclass for a `baseType` which is a godel code of 
 Lean types.
 -/
-class Godel (BaseType : Type)  : Type 1 where
-  toType : BaseType → Type
+class Godel (β : Type)  : Type 1 where
+  toType : β → Type
+open Godel /- make toType publically visible in module. -/
+
+notation "⟦" x "⟧" => Godel.toType x
```
goens, line 12: that's a cool notation, why not call the typeclass `DenoteLean` for example, in allusion to this?

SSA/WellTypedFramework.lean
```diff
@@ -18,118 +21,121 @@ inductive UserType (T : Type) : Type where
...
-def toType [Godel BaseType] : UserType BaseType → Type
-  | .base t => Godel.toType t
+def toType [Godel β] : UserType β → Type
+  | .base t =>  ⟦t⟧
   | .pair k₁ k₂ => (toType k₁) × (toType k₂)
   | .triple k₁ k₂ k₃ => toType k₁ × toType k₂ × toType k₃
   | .unit => Unit
 
-def mkPair {BaseType : Type} [Godel BaseType] {k₁ k₂ : UserType BaseType}: toType k₁ → toType k₂ → toType (.pair k₁ k₂)
+instance [Godel β] : Godel (UserType β) where
```
goens, line 32: nice!

SSA/WellTypedFramework.lean
```diff
@@ -18,118 +21,121 @@ inductive UserType (T : Type) : Type where
...
 
-structure UserData {BaseType : Type} [Godel BaseType] where
-  type : UserType BaseType
-  value : type.toType
+structure UserData [Godel β] where
+  type : UserType β
+  value : toType type
 
 
--- @goens: I suggest removing [DecidableEq ...] [Godel ...]
```
goens: so @bollu style is removing comments :D? I was curious about the rationale (see comment on other commit)

SSA/WellTypedFramework.lean
bollu: Ah, as a generalization of godel coding, where we reflect the metatheory (PA) into the theory using natural numbers. I've seen this type of thing be called `Code`, though I am unsure where --- either in ATAPL, or in one of Andrej Bauer's code.

SSA/WellTypedFramework.lean
bollu: Heh, right, sorry for deleting the comment :) 

the rationale is from abstract algebra: you usually need no typeclass data to construct an object. It is thus evil to prevent users from constructing the object, since then your constructor is not at the maximum level of generality.

In terms of pure engineering, it reduces  the number of typeclasses to be deduced and makes error messages (and code) more concise :)

PR #11, opened 2023-05-16
Simplify tactics in instcombine proof to better understand the proof evolution

tobiasgrosser:


SSA/Examples/InstCombinePeepholeRewrites.lean
```diff
@@ -45,16 +45,36 @@ theorem InstCombineShift279 : ∀ w : Width, ∀ C : BitVector w,
...
+      simp only [outUserType]
+      simp only [eval]
+      simp only [if_true]
+      simp only [argUserType]
+      simp only [Option.some_eq_pure]
+      simp only [EnvC.toEnvU]
+      simp only [uncurry]
+      simp only [Option.elim]
+      simp only [BitVector.width]
+      simp only [pure_bind]
```
goens, line 69: Why not do both at once here instead? as in `simp only [pure_bind, dite_true]`, instead of iterating them multiple times?

PR #15, opened 2023-05-24
feat: implement evalM for monadic evaluation

bollu:
This allows us to express undefined behaviour such as division by zero in a way that represents program sequencing.
- Implementation is straightforward, since I had thought of the type signatures beforehand for ITrees. Sprinkle an `M` at the appropriate locations to make it all work.
- We still keep around the old `eval`. This gives us the version with the cleaner unfolding properties when we want to reason about pure programs, while retaining the monadic version for the cases we really want to unfold through the layers of `Option`s this will create.

### DIFF
--- a/SSA/Core/WellTypedFramework.lean
+++ b/SSA/Core/WellTypedFramework.lean
@@ -226,7 +226,6 @@ def TSSA.evalM {Op β : Type} {M : Type → Type} [Goedel β] [TUSM : TypedUserS
 def TypeSemantics : Type 1 :=
   ℕ → Type
 
-#check Id
 inductive NatBaseType (TS : TypeSemantics) : Type
   | ofNat : ℕ → NatBaseType TS
 deriving DecidableEq


goens:
this is really nice! one thing that strikes me here is that there's a lot of duplication between `evalM` and `eval`. Couldn't we get `eval` as an `abbrev` of `evalM` for `Id`?

SSA/Core/WellTypedFramework.lean
```diff
@@ -175,6 +175,52 @@ def TSSA.eval {Op β : Type} [Goedel β] [TUS : TypedUserSemantics Op β] :
...
+  | _, _, TSSA.op o arg rg => fun e => do 
+    let rgv := rg.evalM e
+    TypedUserSemanticsM.evalM o (e arg) rgv
+  | _, _, .rgn _arg body => fun e arg => do
+      body.evalM (fun _ v =>
+        match v with
+        | Context.Var.prev v' => e v'
+        | Context.Var.last => arg)
+  | _, _, .rgn0 => fun _ => fun x => return x
+  | _, _, .rgnvar v => sorry -- TODO: what to do about rgnvar?
```
goens, line 220: help me out here, why would the obvious `return (e v)` not work?

bollu:
@goens I wrote the `abbrev eval := evalM Id` as an idea on chat. Unfortunately, while semantically they are the same, the proof terms will look different. We will litter the terms with `Id.bind`, `Id`, `Id.map` and so forth. While all of these unfold to no-ops, the unfolding would still need to be done, since `Id` is defined as `def Id (type : Type u) : Type u := type`, and thus has real existence . It's not just an `abbrev` that is removed during elaboration. 

---

This is why I was talking about partial evaluation. If we could partially evaluate the `evalM` at `Id`, then the partial evaluator would eliminate away all this overhead, leaving us with the term that was (syntactically!) isomorphic to `eval`.

SSA/Core/WellTypedFramework.lean
goens: @bollu oh I see what you meant! that makes a lot of sense and is really cool, it's a shame github doesn't allow me to react with an octopus.

bollu:
@goens @ChrisHughes24 I improved the definition of `rgnvar` to be a weaker definition that what we would ideally want, but suffices to merge in, I feel. 

Feel invited to merge if you feel the patch is good.

PR #16, opened 2023-05-26
Tensor1D in TSSA

bollu:
This tests our ability to work with regions.

### DIFF
--- a/SSA/Core/WellTypedFramework.lean
+++ b/SSA/Core/WellTypedFramework.lean
@@ -262,7 +262,7 @@ scoped syntax "[dsl_op|" dsl_op "]" : term
 scoped syntax "%v" num : dsl_var
 
 
-scoped syntax "op:" dsl_op dsl_var ("{" dsl_region "}")? : dsl_expr
+scoped syntax "op:" dsl_op dsl_var ("," dsl_region)? : dsl_expr
 scoped syntax "unit:"  : dsl_expr
 scoped syntax "pair:"  dsl_var dsl_var : dsl_expr
 scoped syntax "triple:"  dsl_var dsl_var dsl_var : dsl_expr
@@ -273,6 +273,8 @@ scoped syntax sepBy(dsl_assign, ";") : dsl_stmt
 -- scoped syntax "dsl_ret " dsl_var : dsl_terminator
 scoped syntax "rgn{" dsl_var "=>" dsl_bb "}" : dsl_region
 scoped syntax "^bb"  (dsl_stmt)?  "dsl_ret " dsl_var : dsl_bb
+
+scoped syntax "rgn$(" term ")" : dsl_region
 
 open Lean Elab Macro in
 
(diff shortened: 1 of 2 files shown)


bollu:
I already discovered something interesting. Naively, one might try and write:

```lean
open SSA EDSL in
theorem extract_map (r0 : TSSA Op _ _) :
  ∀ (e : Env Val) (re : Env (Val → Val)),  -- for metavariable in typeclass
  SSA.eval e re [dsl_region| rgn{ %v0 =>
    ^bb
      %v42 := unit: ;
      %v1 := op:map1d %v0, rgn$(r0) ; -- TODO: add syntax for DSL regions.
      %v2 := op:const(v) %v42;
      %v3 := op:const(43) %v42;
      %v4 := triple:%v1 %v2 %v3;
      %v5 := op:extract1d %v4
      dsl_ret %v5
  }] =
  SSA.eval e re [dsl_region| rgn{ %v0 =>
    ^bb
      %v42 := unit: ;
      %v1 := op:const(v) %v42;
      %v2 := op:const(43) %v42;
      %v3 := triple: %v0 %v1 %v2;
      %v4 := op:extract1d %v3;
      -- jeez, so having intrinsically well typed terms means that I
      -- cannot reuse the same variable r0 as they occur in different
      -- contexts. crazy!
      %v5 := op:map1d %v4, rgn$(r0) -- TODO: add syntax for region variables
      dsl_ret %v5
  }] := by {
    sorry
    -- @chris, @andres: Can I have some help rewriting this to eliminate all the overhead?
  }
```

where `r0` is a _lean level_ region variable (ie, a metavariable from the metatheory). However, the above program does not even type check! `r0` in the first program has a different context than the `r0` in the second program, and thus the above program simply refuses to type check. I had not considered this, it's very interesting.

bollu:
I now made it such that regions are `isolatedFromAbove` --- @ChrisHughes24 , that is to say, a region always runs in a `EnvC.empty`. It would be nice if we can lift this restriction, but I am not sure how. 

@tobiasgrosser we now have a verified tensor theorem that uses regions.

tobiasgrosser:
Lovely. It's this enough to turn the text black?

goens:
That change in `eval` seems very strong. Could SSA be hiding there somewhere? In a sense that it sounds like a type of freshness requirement that could maybe be also given by a weaker condition like SSA?

bollu:
@goens I feel like SSA isn't hiding there, but rather a coercion of contexts. We can always use a smaller context within a larger context.


I don't know how to do it nicely from within Lean, since we could have a region that uses `r := { return b + c }` from within two "different" contexts:

```
b = 1
c = 2
region.run r
```

versus

```
c = 2
b = 1
region.run r
```

The two `r`s will have two different contexts, since the variable declarations occur in different orders.

Morally  I believe we need to quotient the type of contexts by structural rules. I still don't think this will play well with type inference? That is, we can't "just" write the original program with the region occurring in two different contexts. @ChrisHughes24: how would you design this so that the original program fragment works?

bollu:
I chatted with @ChrisHughes24 about this. What we will need is the machinery to define a morphism of contexts according to the structural rules, and to also transport programs along morphisms of contexts. Given this machinery we can transport the region `r` to work in both the original context `(a : nat), (b : nat)` as well as the permuted context `(b : nat), (a : nat)`. This will be quite a bit of code to write down [IIRC, @ChrisHughes24 had ~1000 lines of code for this? Is that correct, chris?] and furthermore, we would need to build the transports in the metaprogramming level. While this is fun, this does not seem well scoped for the POPL deadline.

ChrisHughes24:
This might actually be a good motivation for keeping the variable names as well as the de Bruijn indices, because now we know that the canonical embedding of one context into another is the one that preserves variable names.

PR #18, opened 2023-05-26
Add some of the UB semantics to InstCombine and integrate with Mathlib's Bitvec

goens:


goens:
This is WIP but we can merge it to unblock others (e.g. @ChrisHughes24 or @bollu) to work with this, as suggested on Zulip by @bollu and @tobiasgrosser

bollu:
@goens could you please nuke the `Tensor1D` file before merging into main? That should fix CI, correct?

PR #19, opened 2023-05-28
feat: Add auto generated examples from Alive

bollu:
We can now auto generate Alive examples. I changed the API of `InstCombineBase` in two or three places and left notes.

Tl;DR

1. We can't have two types of bitvec, `bitvec` and `bitvec_option`, since LLVM has only one type of bitvectors. I suggest we model this by mapping `bitvec` into `Option (Bitvector w)`, and doing the wrapping and unwrapping ourselves in the semantics of the operation. 
2. We need to define a new `Bitvector.ofInt'`, that takes a bitvector, and an integer, and converts the integer into a bitvector of the given width. The Lean definition `Bitvector.ofInt` adds one extra bit (to the width of the bitvector) which makes it a no-go, unfortunately. 

The interesting changes to Alive are at:

https://github.com/bollu/alive/compare/a481bfaa574a65a12b0e87b9b6773271a4395d00...e55b33ed712a743bb65531cf149b71e4e430abd7#diff-398f74351065d8eabd4a829028f092764eeca4cdacef31e2b57d8b2b1855e3f6

which adds a custom Lean IR into the alive codebase (which represents our SSA IR), and then performs the appropriate translation from `Alive IR -> Lean IR -> string output`. This code is fairly clean, and ought to be easy to extend. We produce errors for operations we do not cover yet. 

```
error: cannot convert value '(-1 - C)' (type: 'constants.CnstBinaryOp')
error: cannot convert value '(C1 & C2)' (type: 'constants.CnstBinaryOp')
error: cannot convert value '(C1 & ~C)' (type: 'constants.CnstBinaryOp')
error: cannot convert value '(C + 1)' (type: 'constants.CnstBinaryOp')
error: cannot convert value '(C - 1)' (type: 'constants.CnstBinaryOp')
error: cannot convert value '(C - C2)' (type: 'constants.CnstBinaryOp')
error: unknown binop 'ashr i31 %X, C' ; bop.op = '8'
error: unknown binop 'ashr %nx, %y' ; bop.op = '8'
error: unknown binop 'lshr 1, %op0RHS' ; bop.op = '9'
error: unknown binop 'lshr %e1, C2' ; bop.op = '9'
error: unknown binop 'mul %a, %b' ; bop.op = '2'
error: unknown binop 'mul %x, -1' ; bop.op = '2'
error: unknown binop 'sdiv exact %X, %Y' ; bop.op = '4'
error: unknown binop 'sdiv i5 %X, %Y' ; bop.op = '4'
error: unknown binop 'sdiv %X, -1' ; bop.op = '4'
error: unknown binop 'shl 1, %op0RHS' ; bop.op = '7'
error: unknown binop 'srem i9 %X, %Op1' ; bop.op = '6'
error: unknown binop 'udiv 1, %X' ; bop.op = '3'
error: unknown binop 'urem 1, %Op1' ; bop.op = '5'
error: unknown instruction 'icmp eq %a, 0'
error: unknown instruction 'select i1 %a, %a, %b'
error: unknown instruction 'sext %a'
error: unknown instruction 'trunc %a'
error: unknown instruction 'zext %a'
```


goens:
cool, thanks a lot!

1. makes a lot of sense
2. I don't really understand the problem? Why can't we then just call it with `Nat.pred width`? (and handle the zero case especially). Or is that what you want to do with `ofInt'`?

bollu:
2. Yes, exactly, that's what `ofInt'` is for. Hide the implementation details of building a bitvector.

bollu:
@goens if we are happy the patch, I'd suggest merging, and then I can offload the work of fixing the definitions of `Bitvector` to the code owner of the InstCombine semantics,ie, you ;)

SSA/Projects/InstCombine/InstCombineBase.lean
```diff
@@ -109,11 +113,11 @@ def eval (o : Op)
     | Op.ashr _ => match arg with
       | (arg₁,arg₂) => Bitvec.sshr arg₁ arg₂.toNat
     | Op.const c => c
-    | Op.sub _ => uncurry Bitvec.sub arg
-    | Op.add _ => uncurry Bitvec.add? arg
-    | Op.mul _ => uncurry Bitvec.mul? arg
-    | Op.sdiv _ => uncurry Bitvec.sdiv? arg
-    | Op.udiv _ => uncurry Bitvec.udiv? arg
+    | Op.sub _ => (uncurry Bitvec.sub arg)
+    | Op.add _ => (uncurry Bitvec.add? arg).get!
```
goens, line 117: I'd rather have a sorries than `get!`s here to be frank

bollu: @goens sure, SGTM. I'll make the change.

bollu: BTW, for reference, I am happy with you pushing to this branch and then merging once you're happy with the `get! -> sorry` conversion.

PR #23, opened 2023-05-31
Define a Refinement relation

goens:
Allows us to properly state the theorems for UB

### DIFF
--- a/SSA/Projects/InstCombine/ForMathlib.lean
+++ b/SSA/Projects/InstCombine/ForMathlib.lean
@@ -4,6 +4,7 @@ import Mathlib.Algebra.Group.InjSurj
 import Mathlib.Tactic.Ring
 import Mathlib.Data.Int.Cast.Lemmas
 import Mathlib.Data.ZMod.Basic
+import Mathlib.Order.Basic
 
 namespace Vector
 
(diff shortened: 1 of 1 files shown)


bollu:
Overall, looks good to me! I would personally develop the theory in full generality, talking about a refinement for `Option a`. However, the specialization we have here may help with general type inference.

goens:
There's actually an instance of `LT` for option for an arbitrary relation, I thought using it, and I could probably refactor it to be `Option.lt Eq`:

https://github.com/leanprover/lean4/blob/28538fc74889ea2dedbc30e236ce26492e5823ca/src/Init/Data/Option/Basic.lean#L65

I just found out this existed when the instances clashed actually, so I wasn't sure it was worth it, but I'd be happy to do that refactoring if you think it is.

SSA/Projects/InstCombine/ForMathlib.lean
```diff
@@ -424,6 +425,45 @@ theorem shl_ushr_eq_and_shl {w : Nat} {x C : Bitvec w.succ} :
...
+  { apply isTrue; exact Refinement.noneAny }
+  { apply isFalse; intro h; cases h }
+  { rename_i val val'
+    cases (decEq val val')
+    { apply isFalse; intro h; cases h; contradiction } 
+    { apply isTrue; apply Refinement.bothSome; assumption }
+  }
+  
+
+instance {w : Nat} : LE (Option (Bitvec w)) := ⟨Refinement⟩
```
ChrisHughes24, line 459: My mathlib brain says not to make these global instances because there are maybe other orders it could be and it's not clear what `LE` is. I would just define a different notation for it and generalize `refinement` to `Option A`

PR #24, opened 2023-05-31
WIP: Bitfid

goens:
This is to try and get the bitfid decision procedure working in an IR using the (T)SSA framework.


bollu:
@goens am I right in understanding that this is now super-ceded by the development upstream, and if so, it's safe to close the PR?

goens:
@bollu I'm not sure if I would agree with the term "superseeded" (I think it's an interesting question, whether Chris' tactic is actually subsumed by the upstream development), but I think it is safe to close this PR regardless.

PR #25, opened 2023-06-02
Option bitvec

goens:
Push the `Option` into the denotation of the UserType for UB in Alive

### DIFF
--- a/SSA/Core/Util.lean
+++ b/SSA/Core/Util.lean
@@ -3,6 +3,24 @@ import Mathlib.Data.Fin.Basic
 @[simp]
 def uncurry (f : α → β → γ) (pair : α × β) : γ := f pair.fst pair.snd
 
+@[simp]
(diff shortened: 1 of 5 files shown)


goens:
@bollu I feel there's going to be some Kleisli whatnot names for the functions I added in `Util`, maybe you can suggest the proper names for them :)

bollu:
@goens don't know the name, we would call it `liftA2 (,)` or `liftA3 (,,)`, wouldn't give them names :D

goens:
@bollu so are you fine with the current names? Or should we rename them (and what about `pairBind`)? does the rest look okay to you?

bollu:
I am happy with the current names :)

bollu:
@goens Can we refactor `(fun x y => (x == y) ::ᵥ Vector.nil)` and similar into a common function `Bool.toBitVector: Bool -> Bitvec`

goens:
@bollu I'll do you one better and just use `Coe`s :)

PR #27, opened 2023-06-07
add tensor2d theorems

bollu:
- [x] Implement pure theory for https://github.com/bollu/ssa/issues/26
- [x] Write DSL code to express above example.

### DIFF
--- a/SSA/Core/WellTypedFramework.lean
+++ b/SSA/Core/WellTypedFramework.lean
@@ -334,7 +334,7 @@ scoped syntax "()" : dsl_expr2
 scoped syntax "(" dsl_expr2 ")" : dsl_expr2
 scoped syntax "(" dsl_expr2 "," dsl_expr2 ")" : dsl_expr2
 scoped syntax "(" dsl_expr2 "," dsl_expr2 "," dsl_expr2 ")" : dsl_expr2
-scoped syntax "op:" dsl_op2 dsl_expr2 ("," dsl_region2)? : dsl_expr2
+scoped syntax "op:" dsl_op2 ws dsl_expr2 (ws dsl_region2)? : dsl_expr2
 syntax dsl_assign2 := dsl_var2 ":= " dsl_expr2 ";"
 syntax dsl_bb2 := (dsl_assign2)* "return " dsl_expr2
 
@@ -518,7 +518,7 @@ partial def elabStxExpr : TSyntax `dsl_expr2 → SSAElabM (ElabWithTemporaries E
 | `(dsl_expr2| $v:dsl_var2) => do
     let var : ElabVar ← dslVarToElabVar v
     return ElabWithTemporaries.ofVal var
-| `(dsl_expr2| op: $o:dsl_op2 $arg:dsl_expr2 $[, $r?:dsl_region2 ]? ) => do
+| `(dsl_expr2| op: $o:dsl_op2 $arg:dsl_expr2 $[ $r?:dsl_region2 ]? ) => do
   let argelab ← elabStxExpr arg
   let rgn ← match r? with
     | none => `(SSA.TSSA.rgn0)
@@ -618,6 +618,7 @@ def elabStxVar : TSyntax `dsl_var → SSAElabM (TSyntax `term)
   match ← SSAElabContext.getIndex? var.getNat with
   | .some ix => idxToContextVar ix
   | .none => Macro.throwErrorAt var s!"variable '{var}' not in scope"
+| `(dsl_var| $$($x)) => return x
 | stx => Macro.throwErrorAt stx s!"expected variable, found {stx}"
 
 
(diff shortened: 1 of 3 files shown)


bollu:
@tobiasgrosser @goens @ChrisHughes24 , there is an interesting dilemma here that increases the scope of what I need to prove to integrate with our framework.

- First, see that we have the "pure mathematical theorems" about `map`, `extract`, and `fill` already proven.
- Now see that `extract` can potentially fail if we make an out of bounds access. At this point, we have three options:
1. Return an empty tensor of dimension zero to indicate failure
2. Increase our value domain, stick with pure semantics, but allow values to be `Option (Tensor2d' Int)`. This is similar to our compromise for `poison` in the LLVM semantics. It is not very aesthetic in my view
3. Use the monadic semantics. 

If we use the monadic semantics, then an interesting thing happens: the denotation of regions is no longer `Int -> Int`, but is forced to become `Int -> Option Int` , according to our type system! So, we can no longer write theorems about `map`, we will instead have to prove more complex theorems about `mapM`, and potentially reason about side effects.

@tobiasgrosser , how do you suggest we handle failure in the tensor dialect?

bollu:
@goens , review requested.

SSA/Core/WellTypedFramework.lean
```diff
@@ -66,6 +97,36 @@ def trdTriple [Goedel β] {k₁ k₂ k₃ : UserType β} : ⟦k₁.triple k₂ k
 end UserType
 
 
+namespace Example
+
+inductive UserType (β : Type) : Type where
```
goens, line 102: I'm not sure I understand the point of this. Is it testing code you forgot to remove?

SSA/Projects/Tensor2D/Tensor2D.lean
```diff
@@ -0,0 +1,225 @@
+import SSA.Core.WellTypedFramework
+import Mathlib.Data.Matrix.Basic
+import Mathlib.Data.Nat.Basic
+import SSA.Core.Util
+
+
+/-- Tensor2d with existential dimension sizes. -/
+structure Tensor2d' (α : Type) where
+  n₀ : ℕ
```
goens, line 9: Can we use more descriptive names here? Like `dim1` and `dim2` or so? The `n` seems confusing enough, but then the names also don't match the count ;) it's called tensor 2d but I don't see any dimension called 2?

SSA/Projects/Tensor2D/Tensor2D.lean
```diff
@@ -0,0 +1,225 @@
...
+  n₁ : ℕ
+  mat : Matrix (Fin n₀) (Fin n₁) α
+
+def Tensor2d'.error (α : Type) : Tensor2d' α where
```
goens, line 13: do you think this is more elegant than using `Option`. It would break reasoning about the dimensions being preserved, for example, in some cases. I like the explicit option more because it allows you to separate the error mode cleanly

SSA/Projects/Tensor2D/Tensor2D.lean
```diff
@@ -0,0 +1,225 @@
...
+-- TODO: generalize 'const'
+| constNat (v: Nat) | constTensor (t : Tensor2d' Int) | constInt (v : Int)
+| sub
+| map2d
+| fill2d
+| extract2d 
+
+inductive BaseType
+| int : BaseType
+| nat : BaseType
```
goens, line 87: why do we need both nat and int?

SSA/Projects/Tensor2D/Tensor2D.lean
```diff
@@ -0,0 +1,225 @@
...
+| .tensor2d => Tensor2d' Int -- TODO: eventually generalize to arbitrary type.
+
+instance : Goedel BaseType where toType := BaseType.toType
+
+abbrev UserType := SSA.UserType BaseType
+
+-- Can we get rid of the code repetition here? (not that copilot has any trouble completing this)
+@[simp]
+def argUserType : Op → UserType
+| Op.add => ↑(BaseType.int, BaseType.int)
```
goens, line 103: I like these `Coe` instances!

goens:
most of my comments above still persist, but I'm happy to merge this (none of them is too important)

bollu:
I find that only the comment about using `Option` to be open. I would prefer to use `tensor 0 0` as the sentinel error value out of mathematical curiosity :) 

We want to have both `nat` and `int` to show the difference between the MLIR types `index` and (say) `i32`.

Are there other comments that are unresolved?

bollu:
@goens pushed the migration from `Nat` to `Index`. Does the patch now good to you?

goens:
@bollu oh that's interesting indeed! I meant just renaming the constructor, but this looks good to me too (probably cleaner by your design standards of isomorphic types that model different things)

PR #36, opened 2023-06-15
Exec: Reference Executable Semantics

goens:
Executable file and infrastructure to get executable semantics out of tests.


SSA/Projects/Dialect.lean
```diff
@@ -0,0 +1,78 @@
+import SSA.Projects.InstCombine.InstCombineBase
+import SSA.Projects.InstCombine.Tests
+
+inductive Dialect
+  | InstCombine : Dialect
+  deriving Repr
+
```
alexkeizer, line 7: `Dialect` looks like it would be very suited as a typeclass (at some point in the future, might be premature generalization right now)

SSA/Core/WellTypedFramework.lean
```diff
@@ -684,3 +686,13 @@ register_simp_attr TypedUserSemantics.argUserType
 register_simp_attr TypedUserSemantics.outUserType
 register_simp_attr TypedUserSemantics.regionDom
 register_simp_attr TypedUserSemantics.regionCod
+
+
+structure SSA.Executable [Goedel β] where
+  (input : SSA.UserType β)
+  (output : SSA.UserType β)
```
alexkeizer, line 693: Maybe call these `inputType` and `outputType`. My first thought is that these fields might be input and output data that are fed to / produced by some executable

SSA/Projects/Dialect.lean
bollu: I believe it should be an `AttribtuteEnvironmentExtension` so it is in fact open world. Users can register dialects by tagging a `structure Dialect ` with `@[dialect]`.

alexkeizer: That's a fair point. My reason for suggesting the typeclass is that the current approach seems like it will become unmanageable when there are more dialects, since the different operations that need to be defined for that dialect become quite spread out, and operation `Dialect.foo` has to be defined for all dialect in one place. Maybe a middle road is to group the functions that define a `Dialect` into a single structure 
```lean
structure DialectInterface where
  baseType : Dialect → Type
  userType : Dialect → Type
  opType : Dialect → Type
  ..
```

And then have a single function `Dialect.toInterface : Dialect -> DialectInterface`. 

Although, again, maybe a concern for later when we actually need more than just the one dialect

goens: Yep, I agree with all of that @alexkeizer! 

I just think it's two separate issues. I think managing all of those is the right thing for a typeclass (and I would use that to do something like what `TUSInstance` is doing in #37 )! The problem is that we need some way of keeping track of the different dialects we have and use the right types/instances for those depending on the dialect, like we do now in `Main.lean`. That's a separate concern.

Maybe I can do a consolidating PR once this and #37 are merged and deal with it there?

bollu:
@goens I am hoping we can refactor to be like other parts of the Lean compiler, which allow information to be added "at runtime" by using attributes. A good example of what I am thinking of is @hargonix's LCNF pass system, which works as follows:

1. We create a [`PersistentEnvExtension`](https://github.com/leanprover/lean4/blob/bff612e59e4860bef5180db08356ab5f256bea86/src/Lean/Compiler/LCNF/Passes.lean#L90-L96) which holds a list of passes.
2. We create a [`cpass`](https://github.com/leanprover/lean4/blob/bff612e59e4860bef5180db08356ab5f256bea86/src/Lean/Compiler/LCNF/Passes.lean#L110-L120) attribute that is used to tag occurrences of `Lean.Compiler.LCNF.Pass`.
3. We can lookup `PassManager` using [`getPassManager`](https://github.com/leanprover/lean4/blob/bff612e59e4860bef5180db08356ab5f256bea86/src/Lean/Compiler/LCNF/Passes.lean#L98-L99), from which we can extract the list of passes and run them, etc. 

I would like for us to do the same thing here, as it seems to be to be the (only?) principled solution to allow end-users data such as "dialect name" with a given `TypedUserSemantics` in our framework. Since we are writing a framework, we should design it so that the open-world assumption is handled.

goens:
@bollu it's an interesting thought; I don't think it should be blocking this PR though, should it?

bollu (changes requested):
@goens I am marking the status to "request changes" to indicate that I'd like this PR refactored to use the attribute machinery. I feel this way because I would like for us to design this such that we can write in the paper that our framework supports open-world addition of new executable semantics. This requires our design to be extensible by an end-user. The only way I see to achieve such a thing is to allow runtime-registration of new information via attributes. 

I would like for us to be develop this in `Core/Dialects.lean`. The goal of `Projects/` having only subfolders is to *force* us to engineer reusable components that ought to live in `Core/` 😄 I did not anticipate the pattern of keeping files in the `Projects/` hieararchy 😅 . 


I really like the idea of having these executable semantics in our framework. I would like to ship a version of it that's well engineered, so we can proudly write in the paper a little bit about how this works, and how our framework is carefully designed!

hargoniX:
> @goens I am hoping we can refactor to be like other parts of the Lean compiler, which allow information to be added "at runtime" by using attributes. A good example of what I am thinking of is @hargoniX's LCNF pass system, which works as follows:
> 
>     1. We create a [`PersistentEnvExtension`](https://github.com/leanprover/lean4/blob/bff612e59e4860bef5180db08356ab5f256bea86/src/Lean/Compiler/LCNF/Passes.lean#L90-L96) which holds a list of passes.
> 
>     2. We create a [`cpass`](https://github.com/leanprover/lean4/blob/bff612e59e4860bef5180db08356ab5f256bea86/src/Lean/Compiler/LCNF/Passes.lean#L110-L120) attribute that is used to tag occurrences of `Lean.Compiler.LCNF.Pass`.
> 
>     3. We can lookup `PassManager` using [`getPassManager`](https://github.com/leanprover/lean4/blob/bff612e59e4860bef5180db08356ab5f256bea86/src/Lean/Compiler/LCNF/Passes.lean#L98-L99), from which we can extract the list of passes and run them, etc.
> 
> 
> I would like for us to do the same thing here, as it seems to be to be the (only?) principled solution to allow end-users data such as "dialect name" with a given `TypedUserSemantics` in our framework. Since we are writing a framework, we should design it so that the open-world assumption is handled.

Note that the actual system is even cooler than that! `cpass` does not tag `Pass`es it tags `PassInstaller`s that get the current list of passes and return a new one. They can have several behaviors like:
1. Just inserting a Pass at the end of some phase
2. Inserting multiple passes
3. Inserting passes after certain other passes
4. Removing passes that are not compatible
5. probably many more I did not think of yet!

goens:
@bollu I'm happy to move this from `Projects/Dialect` to `Core/Dialect`; I didn't really think too carefully about where to put that file, I just put it somewhere.

I think those are separate concerns: this is about having executable semantics, not about having an open-world, extensible data type to manage different dialects. The attribute machinery is really cool, it'd be really cool if you build that in! I just don't think it has anything to do with reference executable semantics. 

I also don't think this is poorly engineered. It's not as generic as we might want it to be at some point, but that's an entirely different question. I think it's a sensible way of development to have concise PRs that address a single issue (e.g. executable semantics or extensible dialect definitions). It's fine to have big monolithic ones too, I guess, but I'd argue not preferable.

goens:
I also think users should not have to write the boilerplate we've been writing so far. In essence, we could get away with just specifying what the `Op`s should encode and we could generate the rest. Maybe names for the basetypes and or ops, but even thouse could be generated.

SSA/Projects/Dialect.lean
alexkeizer: I certainly wouldn't say this should block the current PR, so dealing with it in a consolidating PR seems fine!

PR #37, opened 2023-06-16
Simp mlir

goens:
Just add the `simp_mlir` macro as we describe it in the paper. @bollu are you okay with us merging this?

### DIFF
--- a/SSA/Core/Tactic.lean
+++ b/SSA/Core/Tactic.lean
@@ -1,0 +1,15 @@
+import SSA.Core.WellTypedFramework
+
+open SSA 
+
+variable [TUS : TUSInstance]
+
+macro "simp_mlir": tactic => 
+  `(tactic| 
+      (simp [TSSA.eval, Function.comp, id.def
+        TypedUserSemantics.eval, Context.Var,
+        TypedUserSemantics.outUserType, TypedUserSemantics.argUserType,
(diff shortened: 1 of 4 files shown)


bollu:
@goens I would prefer the issue mentioned in the tactic (the resolution of `TypedUserSemantics`) works well. I'd personally only feel comfortable changing the statement in the paper from pink to black if the tactic lived in `Core/` and was used everywhere in each `Project` to simplify the semantics

goens:
okay... I don't feel like it's a big issue at all, but I'll get that to work then I guess.

bollu:
@goens I pushed to `simp_mlir` just now to unintroduce the `TUSInstance` typeclass. I do not see why it is necessary here for the tactic. The refactored code `simp_mlir`'s correctly.

bollu:
Review requested, @goens

goens:
@bollu I'm surprised that this just works (not at the computer right now), did you test this, does it reduce all the alive cases?

bollu:
@goens locally, the build is green, which means that the tactic reduces the examples correctly, since our file has `simp_mlir` followed by the "pure math theorem" application.

goens:
cool! I think I missed the general `TypedUserSemantics` addition to the `simp` set, that does make sense! Are we happy to merge this @bollu? (I don't want to merge my own PR)

PR #38, opened 2023-06-17
chore: cleanup unused definitions (NatUserType)

bollu:


tobiasgrosser:
That seems to be just dead code? So probably ready to merge.

PR #39, opened 2023-06-18
More alive

goens:
more alive rewrites, and some broken ones (in a separate file)

### DIFF
--- a/SSA/Core/Tactic.lean
+++ b/SSA/Core/Tactic.lean
@@ -7,7 +7,7 @@ macro "simp_mlir": tactic =>
       (simp [TSSA.eval, Function.comp, id.def, TypedUserSemantics,
         TypedUserSemantics.eval, Context.Var,
         TypedUserSemantics.outUserType, TypedUserSemantics.argUserType,
-        UserType.mkPair] ; 
+        UserType.mkPair, UserType.mkTriple] ; 
        simp [bind_assoc, Option.bind_eq_some', Option.some_bind', 
              Option.bind_eq_bind, pure])
    )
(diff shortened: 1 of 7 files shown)


tobiasgrosser:
Wow, this looks great. Thank you @goens.

Just one question. What is the `broken` file about?

goens:
@tobiasgrosser the `Broken` file has all the rewrites (currently 33) that our syntactic translation pipeline did not fail on, so we could syntactically transform, but that still don't work in our semantics. They either give a type error or timeout trying to elaborate. This could be because:

1. A bug in the syntactic translation that yields ill-formed code (e.g. mismatiching bitwidths). This is the arguably case for example in `Select 485-2` (see my comment in https://github.com/bollu/alive/pull/2#issuecomment-1595913070)
2. A bug in the (unverified) macro expansion that converts the MLIR-like EDSL to the intrinsically well-typed syntax in `TSSA`
3. An issue in Lean (performance) that yields timeouts in the elaboration or DefEq (also probably caused partly by us, but maybe not entirely)
4. A fundamental incompatibility between our TSSA encoding and this program.

I believe most of these will fall under 1 or 2, but I don't truly know. Gives us a few good test cases to debug in any case :)

tobiasgrosser:
Lovely. Should we merge this then?

goens:
I'd be happy to, from my side :) I don't necessarily want to merge my own PR for good meassure, but feel free to do it @tobiasgrosser if you're happy with it as well!

@bollu might have some thoughts e.g. about the directory structure on which he is keen (but then again, those might also be addressed after as well)

tobiasgrosser:
Yes, let's open issues if the directory structure needs addressing.

PR #40, opened 2023-06-18
accelerate lake build

tobiasgrosser:
- use the same lean version between our repo and mathlib
- do not precompile modules (it is not yet clear if this buys us anything)

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -2,7 +2,6 @@ import Lake
 open Lake DSL
 
 package «SSA» where
-  precompileModules := true
 
 @[default_target]
 lean_lib SSA {
--- a/lean-toolchain
+++ b/lean-toolchain
@@ -1,1 +1,1 @@
-leanprover/lean4:nightly-2023-05-18
+leanprover/lean4:nightly-2023-05-16


goens:
@tobiasgrosser I'm curious about the `precompileModules` question. It's true that it's not clear if it buys us anything, maybe we can measure it if it does? My hunch is that it might (e.g. since we've had instances where the elaborator timed out in #39)

goens:
For curiosity, I tested this difference on the `InstCombineAlive.lean` file (which was the most painful build time for us). With `precompileModules`:
```
❯ time lake env lean SSA/Projects/InstCombine/InstCombineAlive.lean
lake env lean SSA/Projects/InstCombine/InstCombineAlive.lean  50.76s user 2.17s system 99% cpu 53.259 total
```

Without `precompileModules`:
```
❯ time lake env lean SSA/Projects/InstCombine/InstCombineAlive.lean
lake env lean SSA/Projects/InstCombine/InstCombineAlive.lean  48.42s user 1.44s system 99% cpu 49.996 total
```

tobiasgrosser:
Thank you for testing. I was not sure how to even test it.

PR #44, opened 2023-06-19
Reference Executable Semantics (Alternative)

bollu:
Alternate proposal to https://github.com/bollu/ssa/pull/36, with an aim of minimal change to `Core/`.


SSA/Projects/InstCombine/InstCombineBase.lean
```diff
@@ -13,6 +13,10 @@ inductive BaseType
   | bitvec (w : Nat) : BaseType
   deriving DecidableEq
 
+instance : Repr BaseType where 
+  reprPrec 
+    | .bitvec w, _ => "bitvec<" ++ repr w ++">"
```
goens, line 18: I would either go for the Lean version `"Bitvec " ++ repr w` or the MLIR version `"i" ++ repr w`, but this looks like some C++ inspired stuff
```suggestion
    | .bitvec w, _ => "i" ++ repr w ++">"
```

SSA/Projects/InstCombine/Tests.lean
```diff
@@ -0,0 +1,28 @@
+import SSA.Projects.InstCombine.InstCombineBase
+
+namespace InstCombine
+
+def tIdx (w : Nat) := (SSA.TSSAIndex.TERMINATOR (SSA.UserType.base (BaseType.bitvec w)))
+
+
```
goens, line 7: I guess this is unnecessary now, right?
```suggestion

````

goens (commented):
I like the API change, thanks!

alexkeizer:
So far, this is only a way to execute specific tests that are pre-compiled into the binary, and not a way to execute arbitrary MLIR from a source file, right? Maybe we want to at least add an option for the second mode to the CLI, but stub out the implementation with an "Executing from a file is not supported yet" error.

bollu:
@goens @alexkeizer another round of review / merge if you're happy!

PR #49, opened 2023-06-20
Use the wraparound version of add/mul for InstCombine

tobiasgrosser:


tobiasgrosser:
This requires a rebuild of all alive files. @goens, can you take over here?

goens:
I still have a bug in the alive conversion script which leaves 27 theorems broken (it is more than before because I discovered another bug that made things work but incorrectly), but this should do for now

PR #53, opened 2023-06-21
Tobiasgrosser/alive simplify theorems

tobiasgrosser:
**Before**
```lean
∀ (w : ℕ) (Z C1 RHS : ℤ),
  some
      (Bitvec.add
        (Bitvec.add (Bitvec.xor (Bitvec.and (Bitvec.ofInt w Z) (Bitvec.ofInt w C1)) (Bitvec.ofInt w C1))
          (Bitvec.ofInt w 1))
        (Bitvec.ofInt w RHS)) ⊑
    some (Bitvec.sub (Bitvec.ofInt w RHS) (Bitvec.or (Bitvec.ofInt w Z) (Bitvec.not (Bitvec.ofInt w C1))))
```
**After**
```lean
∀ (w : ℕ) (Z C1 RHS : ℤ),
  (Bitvec.ofInt w Z &&& Bitvec.ofInt w C1 ^^^ Bitvec.ofInt w C1) + Bitvec.ofInt w 1 + Bitvec.ofInt w RHS =
    Bitvec.ofInt w RHS - (Bitvec.ofInt w Z ||| ~~~Bitvec.ofInt w C1)
```

### DIFF
--- a/SSA/Projects/InstCombine/ForMathlib.lean
+++ b/SSA/Projects/InstCombine/ForMathlib.lean
@@ -469,6 +469,10 @@ inductive Refinement {α : Type u} : Option α → Option α → Prop
   | bothSome {x y : α } : x = y → Refinement (some x) (some y)
   | noneAny {x? : Option α} : Refinement none x?
 
+theorem refinement_some_some {α : Type u} {x y : α} :
+  Refinement (some x) (some y) ↔ x = y :=
+  ⟨by intro h; cases h; assumption, Refinement.bothSome⟩
+
 namespace Refinement
 
 theorem Refinement.refl {α: Type u} : ∀ x : Option α, Refinement x x := by 
--- a/SSA/Projects/InstCombine/InstCombineAliveStatements.lean
+++ b/SSA/Projects/InstCombine/InstCombineAliveStatements.lean
@@ -8,36 +8,48 @@ theorem bitvec_AddSub_1043 :
           (Bitvec.ofInt w 1))
         (Bitvec.ofInt w RHS)) ⊑
     some (Bitvec.sub (Bitvec.ofInt w RHS) (Bitvec.or (Bitvec.ofInt w Z) (Bitvec.not (Bitvec.ofInt w C1))))
-:=sorry
+:= by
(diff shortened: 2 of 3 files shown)


tobiasgrosser:
@goens, that should simplify our alive theorems. Would you be able to rebuild them?

SSA/Core/Tactic.lean
```diff
@@ -7,7 +7,7 @@ macro "simp_mlir": tactic =>
       (simp [TSSA.eval, Function.comp, id.def, TypedUserSemantics,
         TypedUserSemantics.eval, Context.Var,
         TypedUserSemantics.outUserType, TypedUserSemantics.argUserType,
-        UserType.mkPair, UserType.mkTriple] ; 
-       simp [bind_assoc, Option.bind_eq_some', Option.some_bind', 
-             Option.bind_eq_bind, pure])
+        UserType.mkPair, UserType.mkTriple,
```
goens, line 10: why do you remove the split here between the two `simp`s? Does this actually simplify more? (I would suspect it takes more time)

SSA/Projects/InstCombine/ForMathlib.lean
```diff
@@ -472,6 +469,11 @@ inductive Refinement {α : Type u} : Option α → Option α → Prop
   | bothSome {x y : α } : x = y → Refinement (some x) (some y)
   | noneAny {x? : Option α} : Refinement none x?
 
+@[simp]
```
goens, line 472: I would not tag this as `simp` here, because it breaks the abstraction of our simp_mlir. I would add this to the `simp` set above in the tactic.

SSA/Core/Tactic.lean
tobiasgrosser: Because the first simp sometimes resolves the goal, such that the second simp than triggers an error.

SSA/Projects/InstCombine/ForMathlib.lean
tobiasgrosser: Sure, feel free to edit this.

tobiasgrosser:
This has meanwhile been pushed independently.

PR #61, opened 2023-06-22
Try using `Fin` instead of inductive type for referring to variables

tobiasgrosser:
This should decrease the term size `simp` has to traverse, but it makes elaboration more expensive and does not seem to work if the context is not fully known

### DIFF
--- a/SSA/Core/WellTypedFramework.lean
+++ b/SSA/Core/WellTypedFramework.lean
@@ -136,11 +136,17 @@ inductive Context (β : Type) : Type
   | empty : Context β
   | snoc : Context β → Var → (UserType β) → Context β
 
-inductive Context.Var {β : Type} : (Γ : Context β) → UserType β → Type
-  | prev {Γ : Context β} :
-      Context.Var Γ a → Context.Var (Context.snoc Γ v' a') a
-  | last {Γ : Context β} {v : SSA.Var} {a : UserType β} :
-      Context.Var (Context.snoc Γ v a) a
+def Context.length : Context β → Nat
+  | .empty => 0
+  | .snoc Γ _ _ => Γ.length + 1
+
+def Context.getType : (Γ : Context β) → Fin Γ.length → UserType β
+  | .snoc _ _ t, ⟨0, _⟩   => t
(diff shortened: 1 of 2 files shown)


SSA/Projects/InstCombine/InstCombineSlow.lean
```diff
@@ -334,28 +335,19 @@ theorem alive_constant_fold_10: forall (w : Nat), TSSA.eval
...
-  %v151 := op:add w %v150;
-  %v160 := pair:%v151 %v001;
-  %v161 := op:add w %v160;
-  %v170 := pair:%v161 %v001;
-  %v171 := op:add w %v170;
-  %v180 := pair:%v171 %v001;
-  %v181 := op:add w %v180;
-  %v190 := pair:%v181 %v001;
-  %v191 := op:add w %v190
-  dsl_ret %v191
```
bollu, line 348: Why were these removed?

SSA/Core/WellTypedFramework.lean
```diff
@@ -207,19 +212,18 @@ def TSSA.eval {Op β : Type} [Goedel β] [TypedUserSemantics Op β] :
   {Γ : Context β} → {i : TSSAIndex β} → TSSA Op Γ i → (e : EnvC Γ) → i.eval
 | Γ, _, .assign lhs rhs rest, e =>
     rest.eval (fun
-      | _, Context.Var.prev v => e v
-      | _, Context.Var.last => rhs.eval e)
+      | _, { idx := ⟨i + 1, h⟩, of_type } => e { idx := ⟨i, Nat.lt_of_succ_lt_succ h⟩, of_type }
```
bollu, line 215: I wonder if we can make this look cleaner using `@[match_pattern]` to create a pattern synonym for what we are matching here. The current match is quite a whirlwind of syntax.

bollu (commented):
Do we have benchmark numbers that tell us how the performance changes?

I would prefer we merge this after we (a) have numbers and (b) try to make the match look nicer. However, since I'm not actively available, I shall not block the merge.

tobiasgrosser:
We are currently still playing with the encoding. This seems -- so far -- to be a failed attempt.

tobiasgrosser:
This is outdated. Close.

PR #62, opened 2023-06-22
Remove copy operation from instcombine

tobiasgrosser:
It is unclear why we would need a copy operation in an SSA-based IR. @goens @bollu, why do we have a copy operation?

### DIFF
--- a/SSA/Projects/InstCombine/InstCombineBase.lean
+++ b/SSA/Projects/InstCombine/InstCombineBase.lean
@@ -73,7 +73,6 @@ inductive Op
 | mul (w : Nat) : Op
 | sub (w : Nat) : Op
 | neg (w: Nat) : Op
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
Our alive proofs (for some reason) use 'copy'. E.g., 

```
-- Name:Select:489-2
-- precondition: true
/-
  %c = icmp ugt i32 %x, -1
  %r = select i1 %c, %A, %B

=>
  %c = icmp ugt i32 %x, -1
  %r = %B
-/

I guess that just means that R can be simplified to B. I feel our alive encoding is a little fishy here. But let's leave this for now.
```

bollu:
Yes. That's the reason alive encodes it using a `copy`, to match SSA values. We mimic the alive IR.

PR #65, opened 2023-06-23
Add alive_auto tactic

tobiasgrosser:
This commit adds a tactic named `alive_auto` that is run on all alive statements. This allows us to apply additional tactics to all alive proofs without having to edit the auto-generated file.

@goens, could you update the generator and finalize this PR?
@alexkeizer, does this make sense from your perspective?


alexkeizer:
I think `alive_auto` coincides a lot with the `bitvec_aesop` tactic I've been working on in mathlib, but it does make sense to factor it out here as well. If it turns out to just always defer to the mathlib tactic (when/if that one eventually gets upstreamed), we can just do a find/replace.

PR #66, opened 2023-06-23
alive: avoid over-eager simp

tobiasgrosser:
This prevents some alive theorems to turn into 'True'. For example, the following theorem:
```
-- Name:AddSub:1152
-- precondition: true
/-
  %r = add i1 %x, %y

=>
  %r = xor %x, %y

-/
```
is currently `True`. After this PR it folds into:
```
∀ (y x : Bitvec 1), x + y = x ^^^ y
```


tobiasgrosser:
@goens, can you rebuild our alive theorems on top of this change?

tobiasgrosser:
I manually updated `InstCombineAliveStatements.lean` to demonstrate the changes.

PR #67, opened 2023-06-23
alive: use 0/1/-1 to simplify notation

tobiasgrosser:
**Before**
```lean
 ∀ (w : ℕ) (a b : Bitvec w), a + (Bitvec.ofInt w 0 - b) = a - b
 ∀ (w : ℕ) (x C : Bitvec w), (x ^^^ Bitvec.ofInt w (-1)) + C = C - Bitvec.ofInt w 1 - x
```
**After**
```
 ∀ (w : ℕ) (a b : Bitvec w), a + (0 - b) = a - b
 ∀ (w : ℕ) (x C : Bitvec w), (x ^^^ -1) + C = C - 1 - x
```
We can already use 0/1/-1 in theorem statements, and the theorems still compile.

The only problems is ` ∀ (w : ℕ) (b : Bitvec w), b + b = Bitvec.shl b (Bitvec.toNat (Bitvec.ofInt w 1))`, which must be turned into ` ∀ (w : ℕ) (b : Bitvec w), b + b = Bitvec.shl b 1`.

Finally, it seems this change starts to break some proofs. This might be interesting to investigate.

### DIFF
--- a/SSA/Projects/InstCombine/Alive.lean
+++ b/SSA/Projects/InstCombine/Alive.lean
@@ -36,7 +36,7 @@ theorem alive_AddSub_1043 : forall (w : Nat) (C1 Z RHS : Bitvec w)
   %v4 := op:and w %v3;
   %v5 := pair:%v4 %v2;
   %v6 := op:xor w %v5;
-  %v7 := op:const (Bitvec.ofInt w (1)) %v0;
+  %v7 := op:const (1) %v0;
   %v8 := pair:%v6 %v7;
   %v9 := op:add w %v8;
   %v10 := op:const (RHS) %v0;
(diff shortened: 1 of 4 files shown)


goens:
We should be careful with constants, they gave us the timeouts before too. The main issue is that `0` and `1` are elaborated to be of type `Nat` and then coerced into `Bitvec w`, which doesn't always work. We could write `(0 : Int)` or `(1 : Int)` here to ensure this doesn't happen. 

Also we probably shouldn't directly edit `InstCombineAliveStatements`, which is a generated file. We should ideally edit the script in `https://github.com/bollu/alive` and regenerate `InstCombineAlive` out of which we generate `InstCombineAlive`.

tobiasgrosser:
@goens, this patch seems to work now.

I modify the files directly to demonstrate that it compiles error free. However, I cannot run the python scripts, so hope that you can figure out how to auto-generate the right files.

tobiasgrosser:
How nice!

PR #69, opened 2023-06-23
remove redundant InstCombine in a bunch of filenames

goens:
there's a bunch of files that are called InstCombine/InstCombineX, rename them to InstCombine/X


tobiasgrosser (approved):
Very good idea. Good to go as soon as CI passes!

PR #74, opened 2023-06-26
Refactor splitting signed/unsigned, add special functions `-> Bitvec 1`

goens:


goens:
Closing after discussion on Zulip where @ChrisHughes24 advocating having 0 instances rather than 2.

PR #75, opened 2023-06-26
Refactor EDSL2 (the EDSL which allows subexpressions) for new AST

bollu:
- This moves the `EDSL2` into a new file for readability as well.
- Needs some more type tetris to get compiling.

The code is kind of janky, since it passes around this bespoke `StmtBuilder`. I feel this should really be state that is folded into `SSAElabM`. The key idea is that `SSAElabM` should maintain a stack of 'statement builders'. It's a stack because we will nest statement builders each time we build a nested region. 
- It would also be nice to unify `EDSL` with `EDSL2`.


bollu:
@Kha @tobiasgrosser @goens @alexkeizer  review requested

alexkeizer (commented):
My intuition has been that macros are for straightforward syntactic rewrites, whereas if a lot of logic is involved, an elaborator (i.e., working in `TermElabM`) is more appropriate. I feel that the edsl fits very much in the latter category, and was a bit surprised to see we're working in `MacroM`.

Then again, that's just my intuition, I don't know if there is anything wrong, per se, with doing complicated logic in a macro.

bollu:
> My intuition has been that macros are for straightforward syntactic rewrites, whereas if a lot of logic is involved, an elaborator (i.e., working in `TermElabM`) is more appropriate. I feel that the edsl fits very much in the latter category, and was a bit surprised to see we're working in `MacroM`.
> 
> Then again, that's just my intuition, I don't know if there is anything wrong, per se, with doing complicated logic in a macro.

I usually use `MacroM` until I need the full power of `TermElabM`. For example, in this case, it _is_ still purely syntax directed (albeit complicated). I would use `TermElabM` if I need the power of e.e. `inferType`. @Kha , I wonder what your feeling is.

SSA/Core/EDSLNested.lean
```diff
@@ -0,0 +1,257 @@
...
+
+/-
+This is a fancier syntax, which allows expressions in the
+intermediate nodes. This is elaborated into SSA, with temporary
+variable names.
+-/
+syntax dsl_var2 : dsl_expr2
+scoped syntax "()" : dsl_expr2
+scoped syntax "(" dsl_expr2 ")" : dsl_expr2
+scoped syntax "(" dsl_expr2 "," dsl_expr2 ")" : dsl_expr2
```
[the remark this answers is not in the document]
goens: I don't think the `syntax` command generates a pretty printer, does it? I thought it was only `notation` or `infix`. Please do correct me if I'm wrong.

SSA/Core/EDSLNested.lean
```diff
@@ -0,0 +1,257 @@
...
+scoped syntax "(" dsl_expr2 "," dsl_expr2 "," dsl_expr2 ")" : dsl_expr2
```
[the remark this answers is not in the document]
goens: @alexkeizer Our expression language only alows for tuples and triples, so if you want to pass four arguments you have to either make it a triple with a tuple somewhere inside, or a tuple of tuples (4 different options for the same 😬), etc..

SSA/Core/EDSLNested.lean
```diff
@@ -0,0 +1,257 @@
...
+scoped syntax "op:" dsl_op2 ws dsl_expr2 (ws dsl_region2)? : dsl_expr2
+syntax dsl_assign2 := dsl_var2 ":= " dsl_expr2
+scoped syntax sepBy(dsl_assign2, ";") : dsl_stmt2
```
goens, line 30: a bit pedantic, but this is technically a collection of statements, no? and an assign a single statement?

SSA/Core/EDSLNested.lean
```diff
@@ -0,0 +1,257 @@
...
+
+syntax dsl_bb2 := (dsl_stmt2)? "return " dsl_expr2
```
goens, line 32: Interesting, I didn't know you could declare and define syntax categories in a single go

SSA/Core/EDSLNested.lean
```diff
@@ -0,0 +1,257 @@
...
+
+def ElabVar.toString : ElabVar → String
+| .User n => s!"%v{n}"
+| .Synthetic n => s!"%vsynth{n}"
+
+instance : ToString ElabVar where toString := ElabVar.toString
+
+def ElabVar.toNat : ElabVar → Nat
+| .User n => n
+| .Synthetic n => n
```
goens, line 51: this feels evil/dangerous

SSA/Core/EDSLNested.lean
```diff
@@ -0,0 +1,257 @@
...
+  vars : Array ElabVar -- list of variables, in order of occurrence
+
+abbrev SSAElabM (α : Type) := StateT SSAElabContext MacroM α
+
+def SSAElabContext.addVar (var : ElabVar) : SSAElabM Unit := do
+  modify fun ctx => { ctx with vars := ctx.vars.push var }
+
+/- generate a fresh variable for temporary use. -/
+def SSAElabContext.addFreshSyntheticVar : SSAElabM ElabVar := do
+  -- This is a hack. We assign fresh variables an index > 99999.
```
goens, line 68: Why not make a `def SynthVarIndex : Nat -> Nat -> Nat := fun prefix var => prefix + var` and at the end of the elaboration you just instantiate them all with the number of `User` vars? (Probably you'd want to make it a bit nicer than this, e.g. my making your `SSAElabContext` carry the two scopes and then the synthetic variables are parametrized by a single prefix there or so, such that your partially generated syntax is all parametrized by the same prefix)`

goens (commented):
I didn't follow through everything minutiously, but it looks good from afar.

SSA/Core/EDSLNested.lean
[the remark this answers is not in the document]
Kha: It does not generate a delaborator, but it still generates a formatter from that whitespace

alexkeizer:
> > My intuition has been that macros are for straightforward syntactic rewrites, whereas if a lot of logic is involved, an elaborator (i.e., working in `TermElabM`) is more appropriate. I feel that the edsl fits very much in the latter category, and was a bit surprised to see we're working in `MacroM`.
> > Then again, that's just my intuition, I don't know if there is anything wrong, per se, with doing complicated logic in a macro.
> 
> I usually use `MacroM` until I need the full power of `TermElabM`. For example, in this case, it _is_ still purely syntax directed (albeit complicated). I would use `TermElabM` if I need the power of e.e. `inferType`. @Kha , I wonder what your feeling is.

I dig up [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/The.20line.20between.20term.20elaboration.20and.20macro) where I think I took that intuition from, but it doesn't seem like there is any consensus on which approach is better

SSA/Core/EDSLNested.lean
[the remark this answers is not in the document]
alexkeizer: Is this a design choice on our part, our determined by some external factor (MLIR design or so)? It feels a bit arbitrary to support pairs and triples, but not arbitrary n-ary tuples.

Mathlib upstreaming is a bit slow, so I wouldn't mind trying to see if I can generalize this expression language, if you think it could be valuable to have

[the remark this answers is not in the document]
tobiasgrosser: Maybe let's discuss this over zulip?

SSA/Core/EDSLNested.lean
bollu: We call it `stmt` to mirror what we call the [`Index`](https://github.com/bollu/ssa/blob/main/SSA/Core/WellTypedFramework.lean#L165) in TSSA

```lean
inductive TSSAIndex (β : Type) : Type
/-- Ways of making an RHS -/
| EXPR : UserType β → TSSAIndex β
/-- A statement eventually returning a value of type `β` -/
| STMT : UserType β → TSSAIndex β
/-- a lambda -/
| REGION : UserType β → UserType β → TSSAIndex β
```

SSA/Core/EDSLNested.lean
bollu: Unfortunately, it does not suffice to have `prefix + var`, as the synthetic variables must not clash with _any_ variable (even those that occur later!). Thus, the macro expansion would need to be two pass (one pass to collected the largest index that a user uses, and a second pass to fixup the index for all synthetic variables).

goens: @bollu yeah I know, that's exactly why I described it like this, as a *parameter* rather than just adding the (constant) prefix up to that point. If it's a parameter, there's no need to do two passes, you can just instantiate the parameter in the end after the first pass with 1 + the greatest user variable index.

bollu: @goens to be frank, I still I don't see how to do this in a single pass. I'm unsure how the "fixup" is supposed to be done. 

If I were in Haskell, I could express it in a single pass via [Tardis](https://hackage.haskell.org/package/tardis-0.4.4.0/docs/Control-Monad-Tardis.html) and laziness to pass the final value "back up". Here, I'm not too sure 😄 

I propose that you take a stab at it once this is merged. Does that sound like a fair compromise?

bollu:
I've addressed stylistic comments. I've left unadressed the feature request of nested parens (I leave that to you @alexkeizer ) and that of synthetic name index generation (I don't quite follow @goens' suggestion). I would be comfortable with this patch being merged, with suggestions addressed by future PRs . Does that work for the others?

PR #76, opened 2023-06-26
experiments: create rewrite proof

tobiasgrosser:


SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -93,29 +93,103 @@ mk_ex 120
...
+  | add (a : Var) (b : Var)
   | nat (n : Nat)
-  deriving Repr
+  deriving Repr, Inhabited
+
+inductive ExprRec : Type
+  | add (a : ExprRec) (b : ExprRec)
+  | nat (n : Nat)
+  | var (idx : Nat)
+  deriving Repr, Inhabited
```
bollu, line 107: We can unify `ExprRec` and `Expr` into a single `ExprF` and a fixpoint of the initial algebra of `ExprF`. Whether this is useful is  a different question :)

SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -93,47 +99,344 @@ mk_ex 120
...
+        simp [matchVar] at h1
+        split at h1
+        rename_i x n heq
+        · split at h1 <;> rw [hm₀] <;> unfold applyMapping <;> simp
+        · erw [hm₀]
+
+theorem rewriteAtCorrect 
+  (p : Com) (pos: Nat) (rwExpr : ExprRec × ExprRec) 
+  (rewriteCorrect : ∀ s : State, rwExpr.1.denote s = rwExpr.2.denote s)
+  (lets : Lets) (successful : rewriteAt' p pos lets rwExpr = some p'):
```
alexkeizer, line 328: I wonder if it might be beneficial to have `pos` be `Fin (p.size)`, where `p.size` is the number of let-bindings. That is a relatively mild addition of dependent typing, but should let us get rid of this `successful` hypothesis, potentially simplifying proofs

SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -93,47 +99,344 @@ mk_ex 120
...
+def rewrite (inputProg : Com) (depth : Nat) (rewrite: ExprRec × ExprRec) : Com :=
+    let x := rewriteAt inputProg depth rewrite 
+    match x with
+      | none => inputProg
+      | some y => y
 
 def Expr.denote : Expr → State → Value
-| .nat n, _ => .nat n
+| .cst n, _ => .nat n
 | .add a b, l => let a_val := get_nat (l.get! a)
```
alexkeizer, line 257: I wonder if it might be better to use `getD` here, instead of `get!`. For the theorems, I think `get!` with an out of bounds index will default to `0` anyway, so we might as well lean into it and state that all uninitialized variables get implicitly initialized to `0`.

I would also be a fan of indexing `Expr` with an upper-bound on the number of free variables, and making `State` be backed by a `Vector`. Mostly because I like modelling these things exactly, I don't know for sure if it would actually make our lives easier.

alexkeizer:
This PR has become outdated, now that a different version of it has been merged

PR #77, opened 2023-06-27
Fix issue where rewrites with flags would silently ignore them

goens:


bollu:
LGTM! thanks for bumping up the alive commit along with the project commit. I find this to be somewhat painful. We should, perhaps, consider moving the alive code into our project via `git subtree`.

bollu:
@goens may I merge this?

goens:
I still have not understood the point of `git subtree`, I'd be interested in hearing more!

@bollu re merge, yes please! (was waiting for CI)

bollu:
> I still have not understood the point of `git subtree`, I'd be interested in hearing more!

The history of the (sub)repo is stored as part of the main repo. This prevents things like "the link to the submodule breaking because the submodule no longer exists". It also allows for more natural commits which edit the "full repo" as a "single repo", while still allowing the changes in the subfolder to be pushed to the remote if need be. In our case, since `alive` is no longer developed, it may make sense to import it as a `subtree`. That way, during e.g. artifact evaluation, we have just one repo to take care of, and so forth.

PR #78, opened 2023-06-27
Add more notation (for comparisons)

goens:


SSA/Projects/InstCombine/ForMathlib.lean
```diff
@@ -557,11 +557,51 @@ def beq {w : Nat} (x y : Bitvec w) : Bool := x = y
...
+
+
 infixl:75 ">>>ₛ" => fun x y => Bitvec.sshr x (Bitvec.toNat y)
 
+def ult (x y : Bitvec w) : Prop := x.toNat < y.toNat
+def ule (x y : Bitvec w) : Prop := x.toNat ≤ y.toNat
+def ugt (x y : Bitvec w) : Prop := x.toNat > y.toNat
+def uge (x y : Bitvec w) : Prop := x.toNat ≥ y.toNat
+
+instance {w : Nat} (x y : Bitvec w) : Decidable (ult x y) := by apply Nat.decLt
```
bollu, line 582: I am curious where the suggested change works: 
```suggestion
instance {w : Nat} (x y : Bitvec w) : Decidable (ult x y) := by decide
```

SSA/Projects/InstCombine/ForMathlib.lean
```diff
@@ -366,7 +366,7 @@ def srem? {w : Nat} (x y : Bitvec w) : Option $ Bitvec w :=
 /--
  If the condition is an i1 and it evaluates to 1, the instruction returns the first value argument; otherwise, it returns the second value argument.
 -/
-def select {w : Nat} (c : Bitvec 1) (x y : Bitvec w) : Bitvec w :=
+protected def select {w : Nat} (c : Bitvec 1) (x y : Bitvec w) : Bitvec w :=
```
bollu, line 369: why is `select` protected?

SSA/Projects/InstCombine/ForMathlib.lean
goens: that's kind of janky, why does that even work? 😅 anyway, don't you think it's better to keep the more explicit term, if it's about the same size to write?

SSA/Projects/InstCombine/ForMathlib.lean
goens: Oh, good point, maybe I should have explained that in a comment or so.

Basically, for some reason I do not understand, `open Bitvec` does not bring `select` into scope in `AliveStatements.lean`, so by making it protected, the extraction mecanism via my hacky `print_goal_as_error` extracts it with the right namespace 😅 maybe it does make sense to revert this or even investigate why the `open` statement won't work...

SSA/Projects/InstCombine/ForMathlib.lean
bollu: @goens It works because `decide`, IIUC perform typeclass search for a `Decidable` instance. I'd guess even `infer_instance` might do the trick here, if it's able to reduce `ult x y` into `x.toNat > y.toNat`

PR #80, opened 2023-06-29
WIP: Test ring tactic

goens:
Some testing on the ring tactic, it doesn't work entirely yet because it'll timeout trying to decide some concrete instances.


bollu:
@goens am I right in understanding that this is now super-ceded by the development upstream, and if so, it's safe to close the PR?

goens:
yeah

PR #82, opened 2023-08-02
the work on ICom and proving Peephole rewrites

ChrisHughes24:
I think this also bumps mathlib, as well as some changes to an `instCombine` tactic, which I don't think where made by me.


alexkeizer:
The `InstCombine` changes reflect the rename from `Bitvec.carry` to `Bool.carry` done in Mathlib, those lgtm

alexkeizer:
Do we want both `Context` and `ErasedContext`? We've shown that just the erased context is good enough, so we can probably get rid of the former (`Context`).

SSA/Experimental/ErasedContext.lean
```diff
@@ -0,0 +1,138 @@
...
+end Var
+  
+theorem toSnoc_injective {Γ : Ctxt} {t t' : Ty} : 
+    Function.Injective (@Ctxt.Var.toSnoc Γ t t') := by
+  let ofSnoc : (Γ.snoc t').Var t → Option (Γ.Var t) :=
+    fun v => Ctxt.Var.casesOn v some none
+  intro x y h
+  simpa using congr_arg ofSnoc h
+
+abbrev hom (Γ Γ' : Ctxt) := ⦃t : Ty⦄ → Γ.Var t → Γ'.Var t
```
alexkeizer, line 102:
```suggestion
abbrev Hom (Γ Γ' : Ctxt) := ⦃t : Ty⦄ → Γ.Var t → Γ'.Var t
```
Should `Hom` be capitalized? Since it's a type

SSA/Experimental/ErasedContext.lean
```diff
@@ -0,0 +1,138 @@
...
+
+def Var.snocMap {Γ Γ' : Ctxt} (f : hom Γ Γ') {t : Ty} :
```
alexkeizer, line 104:
```suggestion
def Hom.snocMap {Γ Γ' : Ctxt} (f : Hom Γ Γ') {t : Ty} : 
```

This def should probably live in the `Hom` namespace, for easy dot-notation

ChrisHughes24:
> Do we want both `Context` and `ErasedContext`? We've shown that just the erased context is good enough, so we can probably get rid of the former (`Context`).
> 
> Besides that, the PR looks good

I don't see any harm in keeping both just in case

ChrisHughes24:
I've just made Alex's suggested changes so feel free to merge

PR #83, opened 2023-08-03
Rewrite performance tweaks

alexkeizer:
Changes `addProgramInMiddle` to add the new bindings to the lets. This lets us go over the input program only once to adjust the variables, instead of iterating over the program for each new binding, making the process linear instead of quadratic.


SSA/Experimental/ErasedContext.lean
```diff
@@ -101,13 +101,30 @@ theorem toSnoc_injective {Γ : Ctxt} {t t' : Ty} :
...
+  fun _ v => v
+
+/--
+  `map.with v₁ v₂` adjusts a single variable of a Context map, so that in the resulting map
+   * `v₁` now maps to `v₂`
+   * all other variables `v` still map to `map v` as in the original map
+-/
+def Hom.with {Γ₁ Γ₂ : Ctxt} (f : Γ₁.Hom Γ₂) {t : Ty} (v₁ : Γ₁.Var t) (v₂ : Γ₂.Var t) : Γ₁.Hom Γ₂ :=
+  fun t' w =>
+    if h : ∃ h : t = t', h ▸ w = v₁ then
```
goens, line 114: why do you call these both `h`? it's a bit confusing, maybe you can call the two hypothesis differently for readibility?

SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -88,57 +88,54 @@ theorem ICom.denote_changeVars {Γ Γ' : Ctxt}
...
-        else s v')
-  | .ret e, inputProg => by
-    simp only [addProgramAtTop, ICom.denote_changeVars, ICom.denote]
+structure addProgramToLets.Result (Γ_in Γ_out : Ctxt) (ty : Ty) where
+  {Γ_out_new : Ctxt}
+  lets : Lets Γ_in Γ_out_new
+  diff : Ctxt.Diff Γ_out Γ_out_new
+  var : Γ_out_new.Var ty
+
+/--
```
goens, line 97: Maybe the documentation below should go here to explain the result? (I don't feel strongly about this at all, just a thought)

goens (commented):
It looks generally good to me

SSA/Experimental/ErasedContext.lean
alexkeizer: Good point, I just blindly copied this snippet, but it's indeed a bit confusing

PR #87, opened 2023-08-25
N-ary ops

alexkeizer:
Generalize Ops to take an arbitrary (but static / non-variadic) number of arguments

### DIFF
--- a/SSA/Experimental/ErasedContext.lean
+++ b/SSA/Experimental/ErasedContext.lean
@@ -191,25 +191,17 @@ theorem Valuation.snoc_toSnoc {Γ : Ctxt} {t t' : Ty} (s : Γ.Valuation) (x : t.
 /- ## VarSet -/
 
 /-- A `Ty`-indexed family of sets of variables in context `Γ` -/
-def VarSet (Γ : Ctxt) : Type := 
-  (t' : Ty) → Finset (Γ.Var t')
+abbrev VarSet (Γ : Ctxt) : Type := 
+  Finset (Σ t, Γ.Var t)
 
 namespace VarSet
 
 /-- A `VarSet` with exactly one variable `v` -/
 @[simp]
 def ofVar {Γ : Ctxt} (v : Γ.Var t) : VarSet Γ :=
-  fun t' => if ty_eq : t = t' then {ty_eq ▸ v} else {}
-
-@[simp]
-instance : EmptyCollection (VarSet Γ) := ⟨fun _ => ∅⟩
-
-@[simp]
-instance : Union (VarSet Γ) := ⟨fun S₁ S₂ t => S₁ t ∪ S₂ t⟩
+  {⟨_, v⟩} 
 
 end VarSet
-
-
 
 namespace Var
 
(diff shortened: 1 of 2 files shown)


SSA/Experimental/ErasedContext.lean
```diff
@@ -191,26 +191,18 @@ theorem Valuation.snoc_toSnoc {Γ : Ctxt} {t t' : Ty} (s : Γ.Valuation) (x : t.
...
 
 namespace VarSet
 
 /-- A `VarSet` with exactly one variable `v` -/
 @[simp]
 def ofVar {Γ : Ctxt} (v : Γ.Var t) : VarSet Γ :=
-  fun t' => if ty_eq : t = t' then {ty_eq ▸ v} else {}
-
-@[simp]
-instance : EmptyCollection (VarSet Γ) := ⟨fun _ => ∅⟩
```
goens, line 205: why do we want to remove these instances?

SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -9,83 +9,133 @@ import Mathlib.Tactic.Ring
...
+    ∀ {l : List Ty}, B → HVector A l → B
+  | [], b, .nil => b
+  | t::_, b, .cons a as => foldl f (f t b a) as
+
+@[reducible]
+def Op.denote : (op : Op) →
+    HVector Ty.toType op.sig → op.outTy.toType
   | .cst n, _ => n
-  | .add a b, Γv => Γv a + Γv b
+  | .add, .cons a (.cons b .nil) => a + b
```
goens, line 83: that's pretty cool that we can just pattern match on these like that! Do you think it makes sense to add syntatic sugar for the pattern matching, so we don't have to call these constructors explicitly and it just sort of "looks like a tuple/list"?

SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -720,38 +901,38 @@ theorem denote_splitProgramAt {pos : ℕ} {prog : ICom Γ₁ t}
 `target`. If it can match the variables, it inserts `rhs` into the program
 with the correct assignment of variables, and then replaces occurences
 of the variable at position `pos` in `target` with the output of `rhs`.  -/
```
goens, line 903: it seems like the docstring doesn't match the function anymore - could we update the docstring?

SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -773,29 +954,27 @@ structure PeepholeRewrite (Γ : List Ty) (t : Ty) where
...
     rcases v with ⟨i, hi⟩
-    simp only [Erased.out_mk] at hi  
+    simp only [Erased.out_mk] at hi
     rcases List.get?_eq_some.1 hi with ⟨h', rfl⟩
     simp at h'
     convert h ⟨i, h'⟩
   . intro h i
     apply h
 
-def rewritePeepholeAt (pr : PeepholeRewrite Γ t)
```
goens, line 794: maybe this could also get a short docstring?

goens (approved):
Left a few minor comments; I'm happy to merge this ofc.

SSA/Experimental/ErasedContext.lean
alexkeizer: Now that `VarSet` is just an abbrev for a `Finset _`, these instances are found directly as the `Finset` instances

PR #91, opened 2023-09-11
Parse llvm

tobiasgrosser:


SSA/Experimental/ErasedContext.lean
```diff
@@ -39,10 +39,10 @@ def ofList : List Ty → Ctxt Ty :=
   -- Erased.mk
   fun Γ => Γ 
 
+-- Why was this noncomutable? (removed it to make transformation computable)
 @[simp]
```
tobiasgrosser, line 43: Because @ChrisHughes24 suggested to use a non-computable List such that the data-structure can be eliminated when executing the code.

ChrisHughes24: At some point we seem to have got rid of any use of `Erased` in the `Erased` context file. So we're not actually using erased contexts any more

tobiasgrosser: @alexkeizer is working on getting it back.

alexkeizer: https://github.com/opencompl/ssa/tree/ctxt-enhancements

alexkeizer: What you probably want to do, is to have the various `mk*` types that need to compute with the context take a `List Ty` instead of the erased `Ctxt Ty`

alexkeizer (commented):
This PR is now ready for review.

The new EDSL works, and is able to translate concrete syntax with concrete bitwidths into the new `ICom` and co. representation.

There is still a slight issue when trying to make syntax generic over the bitwidth, but we can address that in the future in a subsequent PR if needed.

bollu:
I'd suggest nuking the files that have to do with the Lean compiler's LLVM representation, as they are currently untested/unused.

I would also like to see CI be green before I hit merge 😄 

The PR is a little too large to review sensibly, but overall LGTM!

alexkeizer:
I've had to disable the `mlirnatural` test, since it was written against the old EDSL, and relied on having arbitrary bit-width (which is broken for the new EDSL). But otherwise the CI is now happy.

PR #92, opened 2023-09-12
re-erase the `Ctxt`

alexkeizer:
At some point, we changed `Ctxt` to be just a `List`. We moved away from the reason behind this change,
so we can move back to erased contexts


alexkeizer:
Turns out I was wrong, the `failed to generate splitter for match auxiliary declaration` error is still very much there

alexkeizer:
I'm closing this since it feels unlikely we will go back to erased contexts.
If we find that carrying non-erased contexts becomes a bottle-neck, we should look into encoding `Com`/`Expr` as a subtype of an underlying, untyped datastructure, like we are exploring in https://github.com/opencompl/lean-mlir-common. That way, we only store a single context, at the top-level, rather than a context at every sub-expression.

PR #93, opened 2023-09-14
feat: add example op that uses region: runK

bollu:
In adding this example, I seem to have run into a Lean bug:

```lean
-- (bollu:) unknown free variable: _kernel_fresh.104
@[reducible]
instance : OpDenote ExOp ExTy where
  denote
    | .cst n, _, _ => n
    | .add, .cons (a : Nat) (.cons b .nil), _ => a + b
    | .beq, .cons (a : Nat) (.cons b .nil), _ => a == b
    | .runK (k : Nat), _, _ => (0 : Nat)
```

### DIFF
--- a/SSA/Experimental/IntrinsicAsymptotics.lean
+++ b/SSA/Experimental/IntrinsicAsymptotics.lean
@@ -1059,7 +1059,6 @@ inductive ExOp :  Type
   | add : ExOp
   | beq : ExOp
   | cst : ℕ → ExOp
-  | runK (k : ℕ) : ExOp
   deriving DecidableEq
 
 instance : OpSignature ExOp ExTy where
@@ -1067,46 +1066,12 @@ instance : OpSignature ExOp ExTy where
     | .add => .nat
     | .beq => .bool
     | .cst _ => .nat
-    | .runK _ => .nat
   sig
     | .add => [.nat, .nat]
     | .beq => [.nat, .nat]
     | .cst _ => []
-    | .runK _ => [.nat]
   regSig 
-   | .runK _ => [([.nat], .nat)]
    | _ => []
-
--- Disgusting, @chris, there's gotta be a sensible application
--- operation for our contexts?
-/-- Coerce a very typed function `f`
-    into the expected type (ℕ → ℕ) -/
-def coe_ctxt_nat2nat_to_fun
-  (f : Ctxt.Valuation [ExTy.nat] → ⟦ExTy.nat⟧) (x :  ℕ) : ℕ := sorry
-/-
-  f (fun t i => 
-     match H : t with 
-     | .nat => x
-     | _ => sorry)
- -/
- /-
-  by 
-    simp[Ctxt.Valuation, Var] at f;
-    simp[Goedel.toType] at f;
-    apply f;
-    intros ty var;
-    cases var;
-    case mk var_val var_property =>  {
-      have H : var_val = 0 := by {
-        cases var_val <;> simp at var_property ⊢;
-      };
-      subst H;
-      simp at var_property;
-      subst var_property;
-      simp;
-      exact x; -- apply x
-    }
--/
   
 /-- Compose `f` with itself `n` times -/
 def loopN (f : α → α) (n : ℕ) (x: α) : α := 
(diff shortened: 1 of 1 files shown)


SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -1358,3 +1358,142 @@ example : rewritePeepholeAt p4 5 ex3 = (
...
+    | .runK _ => .nat
+  sig
+    | .add => [.nat, .nat]
+    | .runK _ => [.nat]
+  regSig
+   | .runK _ => [([.nat], .nat)]
+   | _ => []
+
+/-- Compose `f` with itself `n` times.
+  @chris: What's the mathlib correct version of this?
```
ChrisHughes24, line 1391: `Nat.iterate`. There is notation for it. `f^[n]`

SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -1358,3 +1358,142 @@ example : rewritePeepholeAt p4 5 ex3 = (
...
+instance : OpSignature ExOp ExTy where
+  outTy
+    | .add => .nat
+    | .runK _ => .nat
+  sig
+    | .add => [.nat, .nat]
+    | .runK _ => [.nat]
+  regSig
+   | .runK _ => [([.nat], .nat)]
+   | _ => []
```
alexkeizer, line 1388:
```suggestion
    | .runK _ => [([.nat], .nat)]
    | _ => []
```
The spacing of `regSig`'s match arms is a bit off w.r.t. the other definitions

SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -1358,3 +1358,142 @@ example : rewritePeepholeAt p4 5 ex3 = (
...
+      try clear a;
+      try revert f;
+      try revert e;
+      try revert d;
+      try revert c;
+      try revert b;
+      try revert a;
+      clear ll;
+      )
+   )
```
alexkeizer, line 1449: I would prefer to see a single `simp_peephole` tactic that takes the context-specific simp-lemmas as argument.

So, things like `ICom.denote`, `IExpr.denote` can be part of the base `simp_peephole`, and then we can define the specialization to each context as:
```
macro "simp_reg_peephole" : tactic => `(tactic| simp_peephole [rgn, loopN, ...] )
```

bollu:
@ChrisHughes24 , @alexkeizer Another round of review please!

SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -1148,7 +1176,7 @@ def p1 : PeepholeRewrite ExOp [.nat, .nat] .nat:=
   { lhs := m, rhs := r, correct :=
     by
       rw [m, r]
-      simp_peephole
+      simp_peephole [add, cst]
```
alexkeizer, line 1179: I would prefer to see an alias for `simp_peephole [add, cst]` to reduce duplication, but would not consider this a blocker

alexkeizer (approved):
LGTM

SSA/Experimental/IntrinsicAsymptotics.lean
bollu: I prefer the current, explicit version.

tobiasgrosser: I have no idea what [add, cst] means. Why do we now need to add arguments and before simp_peephole was generic?

bollu: `simp_peephole` was never generic, we needed to unfold the (specific) definitions of `add` and `cst` from the example dialect. From now on, since we have multiple dialects, one will need to write `simp_peephole [op1, op2, ..., opn]` for the ops of a dialect

bollu: We can also create an alias as @alexkeizer suggests, one for each dialect. Fundamentally, there's going to be a version of `simp_peephole` for each dialect, as we need to tell `simp` to unfold ops.

tobiasgrosser: I see l
Thank you.

PR #94, opened 2023-09-21
Add Region MVars

ChrisHughes24:
Are people happy with this implementation of regions and region MVars before I work on it further?


SSA/Experimental/IntrinsicAsymptotics.lean
```diff
@@ -1450,17 +1513,31 @@ def ex1_lhs : ICom ExOp [.nat] .nat :=
...
+
+def p1' : PeepholeRewrite ExOp [.nat] .nat:=
+  { lhs := ex1'_lhs, rhs := ex1'_rhs, correct := by
+      rw [ex1'_lhs, ex1'_rhs]
       simp_peephole [add, rgn]
-      simp
+      intros a
+    -- | @chris: what's the correct lemma we need
+    -- to automate this proof?
+      simp[Ctxt.Valuation.snoc]
```
bollu, line 1533: What's the correct `simp` lemma instead of this? The proof state is:

```lean
Ctxt.Valuation.snoc ll✝ ((fun val => mv✝ [ExTy.nat] ExTy.nat 0 fun _ty _var => val)^[0] a)
    (Var.last [ExTy.nat] ExTy.nat) =
  a
```

bollu:
I'm going to close this, since I believe development on this line of proof strategy has halted for the time being.

PR #95, opened 2023-09-26
Generate alive theorems for new encoding

bollu:
This is a WIP patch. Unfortunately, currently, both approaches we have prototyped appear to fail. 

--- 
The meta code version dies with a `max_heartbeats` timeout, while the object code version based on `mkCom (...) |> toOption.get (rfl)` fails due to the `rfl` not holding.

### DIFF
--- a/SSA/Projects/InstCombine/AliveAutoGenerated.lean
+++ b/SSA/Projects/InstCombine/AliveAutoGenerated.lean
@@ -1,0 +1,54 @@
+import SSA.Projects.MLIRSyntax.EDSL
+import SSA.Projects.InstCombine.LLVM.Transform
+import SSA.Projects.InstCombine.LLVM.EDSL
+open MLIR AST
+open MLIR AST
+
+
+-- Name:AddSub:1043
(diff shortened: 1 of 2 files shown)


SSA/Projects/InstCombine/AliveAutoGenerated.lean
```diff
@@ -0,0 +1,54 @@
...
+
+/- It does not like the generic `w`. -/
+def com := 
+  mkCom (AddSub_1043_src) |>.toOption |>.get (by rfl)
+
+
+
+set_option maxHeartbeats 99999999999 in
+def AddSub_1043_tgt (w : Nat):=
+[mlir_icom| {
```
bollu, line 43: Note: The meta code is triggered by our use of `mlir_icom`.

tobiasgrosser:
What is the status on this one? Can we merge a version that works for static types only, maybe?

bollu:
Closed in favor of a new approach that @alexkeizer is prorotyping

PR #97, opened 2023-09-27
Port Tensor1D to new framework

bollu:
We see the dreaded error

> -- Error: unknown free variable: _kernel_fresh.459

in our instance definition

```lean
instance : OpDenote ExOp ExTy where
...
```

This suggests that we should spend the time working around
or fixing this kernel bug.

### DIFF
--- a/SSA/Core/ErasedContext.lean
+++ b/SSA/Core/ErasedContext.lean
@@ -1,5 +1,6 @@
 import Mathlib.Data.Erased
 import Mathlib.Data.Finset.Basic
+import SSA.Core.HVector
 
 /-- 
   Typeclass for a `baseType` which is a Gödel code of Lean types.
@@ -197,6 +198,11 @@ theorem Valuation.snoc_last {Γ : Ctxt Ty} {t : Ty} (s : Γ.Valuation) (x : toTy
 theorem Valuation.snoc_toSnoc {Γ : Ctxt Ty} {t t' : Ty} (s : Γ.Valuation) (x : toType t)
     (v : Γ.Var t') : (s.snoc x) v.toSnoc = s v := by
   simp [Ctxt.Valuation.snoc]
+
+/-- Build valuation from a list of values of types `types`. -/
+def Valuation.ofList {types : List Ty} : HVector toType types → Valuation (Ctxt.ofList types)
+  | .nil => (default : Ctxt.Valuation (∅ : Ctxt Ty))
+  | .cons x xs => (Valuation.ofList xs).snoc x
 
 end Valuation
 
--- a/SSA/Core/ErasedContext.lean
+++ b/SSA/Core/ErasedContext.lean
@@ -199,10 +199,10 @@ theorem Valuation.snoc_toSnoc {Γ : Ctxt Ty} {t t' : Ty} (s : Γ.Valuation) (x :
     (v : Γ.Var t') : (s.snoc x) v.toSnoc = s v := by
   simp [Ctxt.Valuation.snoc]
 
-/-- Build valuation from a list of values of types `types`. -/
-def Valuation.ofList {types : List Ty} : HVector toType types → Valuation (Ctxt.ofList types)
+/-- Build valuation from a vector of values of types `types`. -/
+def Valuation.ofHVector {types : List Ty} : HVector toType types → Valuation (Ctxt.ofList types)
   | .nil => (default : Ctxt.Valuation (∅ : Ctxt Ty))
-  | .cons x xs => (Valuation.ofList xs).snoc x
+  | .cons x xs => (Valuation.ofHVector xs).snoc x
 
 end Valuation
 
--- a/SSA/Projects/Tensor1D/Tensor1D.lean
+++ b/SSA/Projects/Tensor1D/Tensor1D.lean
@@ -315,17 +315,17 @@ This matches the MLIR model, which has a separate `index` type for indexing
 and `iXX/f32/f64` types for values held in tensors.
 -/
 inductive ExOp
-| add
-| const (v: Index)
-| sub
-| map1d
-| extract1d
+| /-- add two integers -/ add
+| /-- create a constant index -/ const (v: Index)
+| /-- subtract two integers -/ sub
+| /-- map a function onto a tensor -/ map1d
+| /-- extract a value at an index of a tensor -/ extract1d
 deriving DecidableEq
 
 inductive ExTy
-| int : ExTy -- values held in tensors
-| ix : ExTy -- indexing into tensors.
-| tensor1d  : ExTy -- tensor type.
+| /-- values held in tensors -/ int : ExTy 
+| /-- shapes and indexes of tensors -/ ix : ExTy
+| /-- tensor type -/ tensor1d  : ExTy
 deriving DecidableEq, Inhabited
 
 instance : Goedel ExTy where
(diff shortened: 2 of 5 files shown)


bollu:
Ping @tobiasgrosser .

tobiasgrosser:
@bollu, can this be rebased and merged?

SSA/Projects/Tensor1D/Tensor1D.lean
```diff
@@ -294,142 +294,74 @@ theorem Tensor1d.tile [Inhabited α] (t : Tensor1d α) (SIZE : 4 ∣ t.size) (f
...
-
-
-
 namespace ArithScfLinalg
 
-open Val
 
-inductive Op
+
+inductive ExOp
```
goens, line 301: why the rename?

SSA/Projects/Tensor1D/Tensor1D.lean
```diff
@@ -294,142 +294,74 @@ theorem Tensor1d.tile [Inhabited α] (t : Tensor1d α) (SIZE : 4 ∣ t.size) (f
...
-    let y : Int := y;
-    x + y
-  | .sub =>
-    let (x, y) := arg;
+
+instance : OpSignature ExOp ExTy where
+  outTy : ExOp → ExTy
+  | .add => .int
+  | .sub => .int
+  | .const _ => .nat
```
goens, line 325: why does `.const` go to `.nat` and not to `.int`?

SSA/Projects/Tensor1D/Tensor1D.lean
```diff
@@ -294,142 +294,74 @@ theorem Tensor1d.tile [Inhabited α] (t : Tensor1d α) (SIZE : 4 ∣ t.size) (f
...
+@[reducible]
+instance : OpDenote ExOp ExTy where
+  denote
+  | .const v, _, _ => v
+  | .add, (.cons x (.cons y nil)), _ =>
+      let x : Int := x;
+      let y : Int := y
+      x + y
+  | .sub, (.cons x (.cons y nil)), _ =>
     let x : Int := x;
```
goens, line 350: does this add implicit coercions, or force `x` and `y` to be `.int`s? why don't you use the `toTuple` function? I guess it's a style question but I'd find it more readable

goens (changes requested):
I find it strange having different `.int` and `.nat` base types and how you can only build constants of `.nat` type - how are you supposed to convert to int? Is it really necessary to have both (couldn't you just have one and either have `0 - n = 0` like in `Nat` or `extract t n m` return garbage for negative `n` or `m`?

Actionably: can I at least please have a docstring documenting the ops and what they do and the relationship between ints and nats here?

SSA/Projects/Tensor1D/Tensor1D.lean
bollu: It forces `x` and `y` to be `.int`. I played around with other ways of doing it, including `toTuple`, but I was unable to convince `Lean` that `Int` was defeq to `[[ExTy.int]]` without this kind of explicit let-shadowing.

SSA/Projects/Tensor1D/Tensor1D.lean
tobiasgrosser: Do you you use `.add` for index computation as well? I am surprised add/sub/const seem to belong together, so either are value computation of index computation. Maybe name them index_add, index_sub, index_const as well as .add, .sub, .const for the value domain?

bollu:
@goens addressed comments!

SSA/Projects/Tensor1D/Tensor1D.lean
bollu: To be consistent with the others in InstCombine

SSA/Projects/Tensor1D/Tensor1D.lean
bollu: @tobiasgrosser The IR is limited, I only use add/sub for values. Happy to extend the name.

SSA/Projects/Tensor1D/Tensor1D.lean
bollu: feel invited to give it a go!

SSA/Projects/Tensor1D/Tensor1D.lean
bollu: Done, remaned.

SSA/Projects/Tensor1D/Tensor1D.lean
goens: Not sure I understand, `InstCombine` doesn't call them `ExOp`, it calls them `Op` (and similarly for `Ty`)

SSA/Core/ErasedContext.lean
```diff
@@ -198,6 +199,11 @@ theorem Valuation.snoc_toSnoc {Γ : Ctxt Ty} {t t' : Ty} (s : Γ.Valuation) (x :
     (v : Γ.Var t') : (s.snoc x) v.toSnoc = s v := by
   simp [Ctxt.Valuation.snoc]
 
+/-- Build valuation from a list of values of types `types`. -/
+def Valuation.ofList {types : List Ty} : HVector toType types → Valuation (Ctxt.ofList types)
```
alexkeizer, line 203: I feel like this should be `Valuation.ofHVector`, but I guess arguments can be made for both

SSA/Projects/Tensor1D/Tensor1D.lean
alexkeizer: I'd argue for sticking with `Op` as well.
I think @bollu is referring to the example dialects in the `Framework.lean` (formerly `IntrinsicAsymptotics`) file, which do use  the`Ex-` prefix to make it explicit they are examples, but those should arguably be put in their own file anyway

SSA/Projects/Tensor2D/Tensor2D.lean
```diff
@@ -205,14 +168,14 @@ theorem map_fill_2d
...
-  (e := e) (Op := Op) (β := BaseType) [dsl_bb2|
+  (e := e) (Op := Op) (β := ) [dsl_bb2|
     return op:fill2d (op:constInt(i) (), 
       op:extract2d (
           ((op:constIx(sz₀) () , op:constIx(sz₁) ()),
            (op:constIx(ix₀) (), op:constIx(ix₁) ())),
            op:constTensor(t) ()))
   ] =
-  TSSA.eval (e := e) (Op := Op) (β := BaseType) [dsl_bb2|
+  TSSA.eval (e := e) (Op := Op) (β := ) [dsl_bb2|
```
alexkeizer, line 178:
```suggestion
  TSSA.eval (e := e) (Op := Op) [dsl_bb2|
```
I'm surprised this is legal syntax, but we probably want to remove the unspecified argument completely (also at line 171 above).

alexkeizer (commented):
Looks good, besides some minor nitpicks

SSA/Core/ErasedContext.lean
bollu: Renamed as requested

SSA/Projects/Tensor2D/Tensor2D.lean
alexkeizer: You forgot the one on line 171, so I took the liberty of removing it

alexkeizer (approved):
LGTM, @goens feel free to merge if you think your concerns have been addressed

PR #100, opened 2023-10-02
Enable projects after move intrinsic asymptotics

bollu:


SSA/MLIRNatural.lean
```diff
@@ -65,4 +64,4 @@ def mainCmd := `[Cli|
     ]
 
 def main (args : List String): IO UInt32 :=
-  mainCmd.validate args
```
goens, line 68: what's up with that?

SSA/Projects/InstCombine/Tactic.lean
```diff
@@ -1,4 +1,3 @@
--- import SSA.Core.WellTypedFramework
```
alexkeizer, line 1: Should this not import `SSA.Core.Framework` instead?

goens (approved):

SSA/Projects/InstCombine/Tactic.lean
alexkeizer: It gets transitively imported anyway, so the import has no effect (but might still be desirable for explicitness, I'll leave that for your judgement).

It got commented because for some reason importing both `TSSA` and the new framework caused an error, but now that reason is gone so it's fine to uncomment it.

Anyway, not a blocker, feel free to merge

PR #102, opened 2023-10-03
Cli entry point

bollu:
This adds an entry point to parse MLIR code and execute it via our denotational semantics.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
           lake -R exe cache get # download cache of mathlib docs.
           lake -R build AliveScaling
 
-      - name: Compile Executable 🧐
+      - name: Compile `mlirnatural` Executable 🧐
         run: |
           lake -R exe cache get # download cache of mathlib docs.
           lake -R build mlirnatural
@@ -47,7 +47,16 @@ jobs:
       - name: LLVM Exhaustive Enumeration
         run: |
           sudo apt install llvm-15 # for opt-15, used to simplify LLVM for reference semantics.
-          cd test/bruteforce-correctness && ./run.sh
+          (cd test/bruteforce-correctness && ./run.sh)
+
+      - name: Compile `opt` Executable 🧐
+        run: |
+          lake -R exe cache get # download cache of mathlib docs.
+          lake -R build opt
+
+      - name: LLVM opt round trip test
+        run: |
+          lake exec opt test/LLVMDialect/InstCombine/bb0.mlir
 
       - name: Check for changes in AliveStatements
         run: |
--- a/SSA/Core/MLIRSyntax.lean
+++ b/SSA/Core/MLIRSyntax.lean
@@ -3,6 +3,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 -/
 import SSA.Core.MLIRSyntax.AST
 import SSA.Core.MLIRSyntax.EDSL
+import SSA.Core.MLIRSyntax.Parser
 import SSA.Core.MLIRSyntax.GenericParser
 import SSA.Core.MLIRSyntax.Transform
 
--- a/SSA/Core/MLIRSyntax/GenericParser.lean
+++ b/SSA/Core/MLIRSyntax/GenericParser.lean
@@ -283,7 +283,6 @@ syntax ident: mlir_type
 syntax "_" : mlir_type
 
 
-set_option hygiene false in -- allow i to expand
 macro_rules
   | `([mlir_type| $x:ident ]) => do
         let xstr := x.getId.toString
@@ -447,18 +446,20 @@ syntax  "{" ("^" ident ("(" sepBy(mlir_bb_operand, ",") ")")? ":")? mlir_ops "}"
 syntax "[mlir_region|" mlir_region "]": term
 
 macro_rules
-| `([mlir_region| { ^ $name:ident ( $operands,* ) : $ops }  ]) => do
+| `(mlir_region| { ^ $name:ident ( $operands,* ) : $ops }) => do
    let initList <- `(@List.nil (MLIR.AST.SSAVal × MLIR.AST.MLIRType _))
    let argsList <- operands.getElems.foldrM (init := initList) fun x xs => `([mlir_bb_operand| $x] :: $xs)
    let opsList <- `([mlir_ops| $ops])
    `(Region.mk $(Lean.quote (name.getId.toString)) $argsList $opsList)
-| `([mlir_region| {  ^ $name:ident : $ops } ]) => do
+| `(mlir_region| {  ^ $name:ident : $ops } ) => do
    let opsList <- `([mlir_ops| $ops])
    `(Region.mk $(Lean.quote (name.getId.toString)) [] $opsList)
-| `([mlir_region| { $ops:mlir_ops } ]) => do
+| `(mlir_region| { $ops:mlir_ops }) => do
    let opsList <- `([mlir_ops| $ops])
    `(Region.mk "entry" [] $opsList)
 
+macro_rules
+| `([mlir_region| $q:mlir_region ]) => `(mlir_region| $q)
 
 macro_rules
 | `([mlir_region| $$($q) ]) => return q
--- a/SSA/Core/MLIRSyntax/Parser.lean
+++ b/SSA/Core/MLIRSyntax/Parser.lean
@@ -1,0 +1,135 @@
+import SSA.Core.MLIRSyntax.AST
+import SSA.Core.MLIRSyntax.EDSL
+import Lean
+/- Parse  a raw file into `MLIR.AST.*` constructs. -/
+open Lean
+
+variable {ParseOutput : Type} [ToString ParseOutput]
+
+abbrev ParseError := String
+abbrev ParseFun : Type := Lean.Environment → String → EIO ParseError ParseOutput
+
+-- We use unification with `isDefEq` to find the synthetic metavariables, and then we synthesize and instantiate them
+-- This lets us instantiate arguments like the `Nat` argument to `MLIR.AST.Region`.
+-- For more info, see: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Calling.20unification.20from.20meta.20to.20remove.20metavariables
+unsafe def elabIntoTermTactic {α : Type} (expectedType : Expr) (stx : Lean.Syntax) : Elab.Tactic.TacticM α := do
+  let expr ← Lean.Elab.Tactic.elabTerm stx none
+  let _ ← Meta.isDefEq (← Meta.inferType expr) expectedType
+  Elab.Term.synthesizeSyntheticMVarsNoPostponing
+  let expr ← instantiateMVars expr
+  Meta.evalExpr α expectedType expr
+
+namespace Lean.Elab.Tactic
+
+@[inline] private def TacticM.runCore (x : TacticM α) (ctx : Context) (s : State) : TermElabM (α × State) :=
+  x ctx |>.run s
+
+@[inline] private def TacticM.runCore' (x : TacticM α) (ctx : Context) (s : State) : TermElabM α :=
+  Prod.fst <$> x.runCore ctx s
+
+end Lean.Elab.Tactic
+
+unsafe def elabIntoTermElab {α : Type} (expectedType : Expr) (stx : Lean.Syntax) : Elab.Term.TermElabM α := do
+  elabIntoTermTactic (α := α) expectedType stx  |>.runCore' { elaborator := `ParserHack} default
+
+unsafe def elabIntoMeta {α : Type} (expectedType : Expr) (stx : Lean.Syntax) : MetaM α :=
+  elabIntoTermElab (α := α) expectedType stx |>.run'
+
+unsafe def elabIntoCore {α : Type} (expectedType : Expr) (stx : Lean.Syntax) : CoreM α :=
+  elabIntoMeta (α := α) expectedType stx |>.run'
+
+def printException : Except Exception α → IO String
+  | Except.ok _ => throw <| IO.userError "printException called on Except.ok"
+  | Except.error e => e.toMessageData.toString
+
+unsafe def elabIntoEIO {α : Type} (env : Lean.Environment) (typeName : Lean.Expr) (stx : Lean.Syntax) : EIO ParseError α :=
+  fun s =>
+    let resE : EIO Exception α := elabIntoCore (α := α) typeName stx |>.run' {fileName := "parserHack", fileMap := default} {env := env}
+    match resE s with
+    | .ok a s => .ok a s
+    | .error exception s =>
+      let errMsgIO : IO String := exception.toMessageData.toString
+      match errMsgIO s with
+      | .ok errMsg s => .error s!"failed elaboration {stx}. Error: {errMsg}" s
+      | .error e s => .error s!" Failed elaborating {stx}.\nUnable to pretty-print exception at 'elabIntoEIO':\n{e}." s
+
+def region0Expr := (Expr.app (Expr.const `MLIR.AST.Region []) (Expr.const `Nat.zero []))
+
+def op0Expr := (Expr.app (Expr.const `MLIR.AST.Op []) (Expr.const `Nat.zero []))
+/--
+  Parse `Lean.Syntax` into `MLIR.AST.Region`.
+
+  This runs the Lean frontend stack and adds it to the TCB.
+  It should be safe (if we trust the lean parser) since we ensure that the types match: see
+  https://leanprover-community.github.io/mathlib4_docs/Std/Util/TermUnsafe.html#Std.TermUnsafe.termUnsafe_
+ -/
+def elabRegion (env : Lean.Environment) (stx : Lean.Syntax) : EIO ParseError (MLIR.AST.Region 0) := do
+  let reg ← unsafe elabIntoEIO (α := MLIR.AST.Region 0) env region0Expr stx
+  return reg
+
+/--
+  Parse `Lean.Syntax` into `MLIR.AST.Op`.
+
+  This runs the Lean frontend stack and adds it to the TCB.
+  It should be safe (if we trust the lean parser) since we ensure that the types match: see
+  https://leanprover-community.github.io/mathlib4_docs/Std/Util/TermUnsafe.html#Std.TermUnsafe.termUnsafe_
+ -/
+def elabOp (env : Lean.Environment) (stx : Lean.Syntax) : EIO ParseError (MLIR.AST.Op 0) :=
+  unsafe elabIntoEIO (α := MLIR.AST.Op 0) env op0Expr stx
+
+def mkParseFun (syntaxcat : Name)
+    (ntparser : Syntax → EIO ParseError ParseOutput)
+    (s : String) (env : Environment) : EIO ParseError ParseOutput := do
+  match Parser.runParserCategory env syntaxcat s with
+    | .error msg => throw msg
+    | .ok syn => ntparser syn
+
+private def mkNonTerminalParser (syntaxcat : Name)
+    (ntparser : Syntax → EIO ParseError ParseOutput)
+    (env : Environment) (s : String) : EIO ParseError ParseOutput :=
+  let parseFun := mkParseFun syntaxcat ntparser
+  parseFun s env
+
+-- TODO: do we need this copy of the environment, or can I remove it from one of the two?
+def regionParser : @ParseFun (MLIR.AST.Region 0) :=
+  fun env : Lean.Environment => mkNonTerminalParser `mlir_region (elabRegion env) env
+
+private def parseFile (env: Lean.Environment)
+    (parser: @ParseFun ParseOutput)
+    (filepath: System.FilePath) : IO (Option ParseOutput) := do
+  let lines ← IO.FS.lines filepath
+  let fileStr := "\n".intercalate lines.toList
+  let parsed ← EIO.toIO' <| parser env fileStr
+  match parsed with
+  | .ok parseOutput => return parseOutput
+  | .error msg => IO.println s!"Error parsing {filepath}:\n{msg}"; return none
+
+private def isFile (p: System.FilePath) : IO Bool := do
+  return (← p.metadata).type == IO.FS.FileType.file
+
+def runParser (parser : @ParseFun ParseOutput) (fileName : String) : IO (Option ParseOutput) := do
+  /- We expect our parser to run with `lake exec`, which sets `LEAN_PATH` to be a colon separated
+    list of package paths.
+    We parse the package paths and set the search path accordingly.
+    For more, see `lake env help.` -/
+  let packagePaths : List String :=
+    match (← IO.getEnv "LEAN_PATH") with
+    | .none => []
+    | .some colonSeparatedPaths => colonSeparatedPaths.splitOn ":"
+  if packagePaths.isEmpty then
+    throw <| IO.userError "Expected `LEAN_PATH` environment variable to be set. Are you running via `lake exec opt`?"
+  initSearchPath (← Lean.findSysroot) packagePaths
+  let modules : Array Import := #[⟨`SSA.Core.MLIRSyntax.EDSL, false⟩]
+  let env ← importModules modules {}
+  let filePath := System.mkFilePath [fileName]
+  if !(← isFile filePath) then
+    throw <| IO.userError s!"File {fileName} does not exist"
+  parseFile env parser filePath
+
+def parseRegionFromFile (fileName : String)
+    (regionParseFun : MLIR.AST.Region 0 → Except ParseError α) : IO (Option α)  := do
+  let ast ← runParser regionParser fileName
+  match Option.map regionParseFun ast with
+  | .some (.ok res) => return res
+  | .some (.error msg) => throw <| IO.userError s!"Error parsing {fileName}:\n{msg}"; return none
+  | .none => return none
--- a/SSA/Projects/InstCombine/LLVM/Parser.lean
+++ b/SSA/Projects/InstCombine/LLVM/Parser.lean
@@ -1,0 +1,15 @@
+import SSA.Core.MLIRSyntax.Parser
+import SSA.Projects.InstCombine.LLVM.EDSL
+import Init.Data.Repr
+
+open MLIR AST InstCombine
+def regionTransform (region : Region 0) : Except ParseError
+  (Σ (Γ' : Context) (eff : EffectKind) (ty : Ty), Com LLVM Γ' eff ty) :=
+    let res := mkCom (d:= LLVM) region
+    match res with
+      | Except.error e => Except.error s!"Error:\n{reprStr e}"
+      | Except.ok res => Except.ok res
+
+def parseComFromFile (fileName : String) :
+ IO (Option (Σ (Γ' : Context) (eff : EffectKind) (ty : InstCombine.Ty), Com LLVM Γ' eff ty)) := do
+ parseRegionFromFile fileName regionTransform
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -12,6 +12,13 @@ lean_exe mlirnatural {
   root := `SSA.MLIRNatural
 }
 
+
+lean_exe opt {
+  root := `SSA.Projects.InstCombine.LLVM.Opt
+  -- `supportInterpreter` flag needed to link executable with the Lean frontend
+  supportInterpreter := true
+}
+
 lean_exe ssaLLVMEnumerator {
   root := `SSA.Projects.InstCombine.LLVM.Enumerator
 }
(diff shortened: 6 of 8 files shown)


bollu:
@goens I suggest we do this in a follow-up PR, but we should test that running our entrypoint actually applies the rewrites. I've historically used the [`llvm-lit`](https://llvm.org/docs/CommandGuide/lit.html) infrastructure for this type of test.

goens:
@bollu sure! maybe we want to open an issue to keep track of that later?

bollu:
@goens I pushed some changes, mainly to reflow `Parser.lean`. Let me know if the changes work for you!

bollu:
@goens I think we've now moved the `Transform` machinery into `SSA/Core/`,  so we should delete that from this PR, and include it from `SSA/Core`.

goens:
@bollu I was working on that :( I'm not sure now,  but I would appreciate if you would coordinate before doing that next time

bollu:
Aha, go ahead and revert @goens !

bollu:
I was merely procrastinating from admin work 😄

bollu:
And yes, I'll ping before I snipe some low hanging PR fruit next time (:

lake-manifest.json
```diff
@@ -4,7 +4,7 @@
  [{"url": "https://github.com/leanprover/std4",
    "type": "git",
    "subDir": null,
-   "rev": "32983874c1b897d78f20d620fe92fc8fd3f06c3a",
+   "rev": "e840c18f7334c751efbd4cfe531476e10c943cdb",
```
tobiasgrosser, line 7: I am curious why the std library changes in this PR?

goens: That was unintentional (my lake-manifest was broken from the merge). I can revert that change.

goens:
@bollu sorry, I did try to merge but you had moved all the whitespace around and that broke the semantic changes I was working on

lake-manifest.json
tobiasgrosser: That would be good. 😄

bollu:
@goens : quickly noting that I no longer remember the difference between the `opt` executable and the `mlirnatural` executable. Could you jog my memory for me please?

tobiasgrosser:
You probably want to merge main after the mathlib bump is in to reduce the diff shown here.

goens:
> @goens could you please edit the changes to the CI in [`.github/workflows/ci.yml`](https://github.com/opencompl/ssa/pull/102/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f) to test the `opt` that we build? what's the correct way to run the test?

what you added there is the right way! (it's just not working yet, that's a different thing)

tobiasgrosser:
> > You probably want to merge main after the mathlib bump is in to reduce the diff shown here.
> 
> yep! that was precisely the point of that bump :)

For some reason it looks as if you cherry-picked the relevant commits instead of doing an `git remote update; git merge origin/main`. Without the actual merge commit, github will not use the new main as base for its diff, no? Maybe you are still waiting for the merge queue to finish. In that case, just ignore this comment.

SSA/Core/MLIRSyntax/AST.lean
```diff
@@ -221,7 +221,6 @@ def Region.ops (region: Region φ) : List (Op φ) :=
   match region with
   | Region.mk _ _ ops => ops
 
-
```
tobiasgrosser, line 224: Unrelated change.

bollu:
It maybe that we need to substitute the metavars that have been assigned in the `Meta` monad context.

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,136 @@
...
+-- TODO: do we need this copy of the environment, or can I remove it from one of the two?
+def regionParser : @ParseFun (MLIR.AST.Region 0) :=
+  fun env : Lean.Environment => mkNonTerminalParser `mlir_region (elabRegion env) env
+
+private def parseFile (env: Lean.Environment)
+    (parser: @ParseFun ParseOutput)
+    (filepath: System.FilePath) : IO (Option ParseOutput) := do
+  let lines ← IO.FS.lines filepath
+  let fileStr := lines.foldl (λ s₁ s₂ => s₁ ++ "\n" ++ s₂) "" -- TODO: use intercalate.
+  -- IO.println s!"[DEBUG] fileStr=\n{fileStr}"
```
tobiasgrosser, line 98: Leftover debug statement?

lakefile.lean
```diff
@@ -12,6 +12,12 @@ lean_exe mlirnatural {
   root := `SSA.MLIRNatural
 }
 
+
+lean_exe opt {
+  root := `SSA.Projects.InstCombine.LLVM.Opt
+  supportInterpreter := true
```
tobiasgrosser, line 18: I am curious why we need the `supportInterpreter` here.

.github/workflows/ci.yml
```diff
@@ -39,14 +39,24 @@ jobs:
           lake -R exe cache get # download cache of mathlib docs.
           lake -R build AliveScaling
 
-      - name: Compile Executable 🧐
+      - name: Compile `mlirnatural`Executable 🧐
```
tobiasgrosser, line 42: whitespace? A

tobiasgrosser:
How lovely. Thank you, @goens for pushing this through!

SSA/Core/MLIRSyntax/Parser.lean
goens: yeah, I like to keep those around in case things break again :sweat_smile: don't know what the ideal way is to do that

lakefile.lean
goens: that's because we use the Lean frontend in our code (we call the Lean parser), so that it also links it.

tobiasgrosser: Maybe add a comment explaining this.

SSA/Core/MLIRSyntax/Parser.lean
tobiasgrosser: I guess I would either just leave it in the git history, add a comment why you want to leave this, or make this code that is disabled on a flag and set the flag to off (and add a comment there).

tobiasgrosser (approved):

SSA/Core/MLIRSyntax/Parser.lean
goens: OK, I removed it. Just please don't squash this PR when merging so that the git history is not lost.

tobiasgrosser: Do you not feel the history in this PR is sufficient? I feel this will make it significantly harder to actually see the real change in the history, will add commits-that-do-not-compile to the history, and also will randomly change our base std library back and forth. I am curious why the history in the PR is not sufficient for someone who wants to investigate this change?

goens: because if we squash this PR then we lose this history. It will be only on github and on some random stale branch that's not in the project's history. So if we're working on a file later and want to figure out what happened, why something was there or look at an older version of it, we cannot do 'git log path/to/file.lean`, we have to look it up in every single such PR branch, and that makes it completely unviable in my opinion. 

In my personal opinion `git log` and `git blame` etc are significantly more valuable than `git bisect`, and accurate, fine-grained helps more than a curated history where CI passes for every commit. I could see why for a very large project written in a chaotic language like C++ these priorities would be different, but for an academic project with a much lower PR/commit volume, in a strong typesystem, I think the value of the history is higher (I've never needed to bisect anything in this project, have you?)

tobiasgrosser: I feel that `git blame` is a lot more efficient as it points to the single PR that had a functional change to the repo rather than hitting some of these 42 commits and having to figure it out. In any case, this seems to need further discussion. Happy to have these eventually and not block this PR over our different perspectives here.

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,138 @@
...
+  return (← p.metadata).type == IO.FS.FileType.file
+
+def runParser (parser : @ParseFun ParseOutput) (fileName : String) : IO (Option ParseOutput) := do
+  let executablePath : System.FilePath ← IO.appPath
+  let dotLakePath : System.FilePath :=
+    match executablePath.parent with
+      | .none => ".lake"
+      | .some binPath => match binPath.parent with
+        |.none => ".lake"
+        |.some buildPath => match buildPath.parent with
```
bollu, line 116:
```suggestion
        | .none => ".lake"
        | .some buildPath => match buildPath.parent with
```

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,138 @@
...
+  | .ok parseOutput => return parseOutput
+  | .error msg => IO.println s!"Error parsing {filepath}:\n{msg}"; return none
+
+private def isFile (p: System.FilePath) : IO Bool := do
+  return (← p.metadata).type == IO.FS.FileType.file
+
+def runParser (parser : @ParseFun ParseOutput) (fileName : String) : IO (Option ParseOutput) := do
+  let executablePath : System.FilePath ← IO.appPath
+  let dotLakePath : System.FilePath :=
+    match executablePath.parent with
```
bollu, line 112: This feels very fragile to me; Can we generalize this, to walk up the parent tree till it finds a `.lake` folder, throwing an error if it doesn't find one, rather than assuming the depth?

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,138 @@
...
+
+-- TODO: do we need this copy of the environment, or can I remove it from one of the two?
+def regionParser : @ParseFun (MLIR.AST.Region 0) :=
+  fun env : Lean.Environment => mkNonTerminalParser `mlir_region (elabRegion env) env
+
+private def parseFile (env: Lean.Environment)
+    (parser: @ParseFun ParseOutput)
+    (filepath: System.FilePath) : IO (Option ParseOutput) := do
+  let lines ← IO.FS.lines filepath
+  let fileStr := lines.foldl (λ s₁ s₂ => s₁ ++ "\n" ++ s₂) "" -- TODO: use intercalate.
```
bollu, line 100:
```suggestion
  let fileStr := "\n".intercalate lines.toList
```

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,138 @@
...
+      | .some binPath => match binPath.parent with
+        |.none => ".lake"
+        |.some buildPath => match buildPath.parent with
+          | .none => ".lake"
+          | .some dotLakePath => dotLakePath
+  let libPath : System.FilePath := dotLakePath / "build" / "lib"
+  -- TODO: can we get these from the lakefile?
+  let deps := ["mathlib", "std", "aesop", "Qq", "importGraph", "partax", "proofwidgets", "Cli"]
+  let depsPackages : List System.FilePath := deps.map fun depName =>
+    dotLakePath / "packages" / depName / ".lake" / "build" / "lib"
```
bollu, line 123: This too feels fragile. Any chance we can as Mac if it's possible to use `Lake` as a library to achieve this?

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,138 @@
...
+  let reg ← unsafe elabIntoEIO (α := MLIR.AST.Op 0) env op0Expr stx
+  return reg
+
+def mkParseFun (syntaxcat : Name)
+    (ntparser : Syntax → EIO ParseError ParseOutput)
+    (s : String) (env : Environment) : EIO ParseError ParseOutput := do
+  let syn : Except String Syntax := Parser.runParserCategory env syntaxcat s
+  match syn with
+    | .error msg => throw msg
+    | .ok syn => ntparser syn
```
bollu, line 84:
```suggestion
  match syn with
    | .error msg => throw msg
    | .ok syn => ntparser syn
```
  match Parser.runParserCategory env syntaxcat s with
    | .error msg => throw msg
    | .ok syn => ntparser syn

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,138 @@
...
+/--
+  Parse `Lean.Syntax` into `MLIR.AST.Op`.
+
+  This runs the Lean frontend stack and adds it to the TCB.
+  It should be safe (if we trust the lean parser) since we ensure that the types match: see
+  https://leanprover-community.github.io/mathlib4_docs/Std/Util/TermUnsafe.html#Std.TermUnsafe.termUnsafe_
+ -/
+def elabOp (env : Lean.Environment) (stx : Lean.Syntax) : EIO ParseError (MLIR.AST.Op 0) := do
+  let reg ← unsafe elabIntoEIO (α := MLIR.AST.Op 0) env op0Expr stx
+  return reg
```
bollu, line 76:
```suggestion
def elabOp (env : Lean.Environment) (stx : Lean.Syntax) : EIO ParseError (MLIR.AST.Op 0) :=
  unsafe elabIntoEIO (α := MLIR.AST.Op 0) env op0Expr stx
```

.github/workflows/ci.yml
bollu: @tobiasgrosser for things like this, it's often easier for everyone if you press  <kbd>⌘</kbd> + <kbd> g </kbd> when writing the review, so you can write down a suggested change, as I've done here: 
```suggestion
      - name: Compile `mlirnatural` Executable 🧐
```

SSA/Core/MLIRSyntax/Parser.lean
goens: lol, I don' t know why I left that TODO instead of doing it :thinking:

SSA/Core/MLIRSyntax/Parser.lean
goens: @bollu agreed! However, I' d feel that suggestion is also too fragile, because it will also not work if we at some point have a ` make install`. For that reason I' m not happy with that change, I'd request that we either do it right right (i.e. by having a configurable prefix or an $SSA_HOME or something) or we leave it like it is until we have a proper solution.

bollu:
@goens wondering if you got the chance to follow up with mac.

goens:
@bollu no sorry, I have not, haven't had the time.

SSA/Core/MLIRSyntax/Parser.lean
bollu: @goens how about making the user give us the root repo directory, and we navigate relative to this?

bollu:
@goens no worries, was wondering.

SSA/Core/MLIRSyntax/Parser.lean
goens: with a default argument? yeah that sounds like a good compromise!

bollu: > with a default argument? yeah that sounds like a good compromise!

Yes, where the default argument, IMO, recursively walks the directory tree to find a `.lake`.

goens: I'm trying to avoid writing code that recusively walks down the directory tree to find a `.lake` because it feels like not so useful work (if the directory structure changes it breaks), I can find almost no use-case for it over what we just have + being able to provide that directory explicitly

goens:
@bollu thanks for the change, that does seem more sane and solves both issues :) so we've got green light to merge, right?

bollu:
Yes, with maybe a final quick skim from @tobiasgrosser

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,132 @@
+import SSA.Core.MLIRSyntax.AST
+import SSA.Core.MLIRSyntax.EDSL
+import Lean
+
```
bollu, line 4: @goens can we have a docstring that says what this file does?

SSA/Core/MLIRSyntax/GenericParser.lean
```diff
@@ -283,7 +283,7 @@ syntax ident: mlir_type
 syntax "_" : mlir_type
 
 
-set_option hygiene false in -- allow i to expand
+--set_option hygiene false in -- allow i to expand
```
bollu, line 286:
```suggestion
```
drop dead code.

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,132 @@
...
+    throw <| IO.userError s!"File {fileName} does not exist"
+  parseFile env parser filePath
+
+def parseRegionFromFile (fileName : String)
+    (regionParseFun : MLIR.AST.Region 0 → Except ParseError α) : IO (Option α)  := do
+  let ast ← runParser regionParser fileName
+  match Option.map regionParseFun ast with
+  | .some (.ok res) => return res
+  | .some (.error msg) => throw <| IO.userError s!"Error parsing {fileName}:\n{msg}"; return none
+  | .none => return none
```
bollu, line 132: My feeling is that the contents of this file as well can be moved into `SSA/Projects/InstCombine/LLVM/Opt.lean`, and can be judiciously refactored out when necessary.

SSA/Projects/InstCombine/LLVM/Parser.lean
```diff
@@ -0,0 +1,15 @@
...
+def regionTransform (region : Region 0) : Except ParseError
+  (Σ (Γ' : Context) (eff : EffectKind) (ty : Ty), Com LLVM Γ' eff ty) :=
+    let res := mkCom (d:= LLVM) region
+    match res with
+      | Except.error e => Except.error s!"Error:\n{reprStr e}"
+      | Except.ok res => Except.ok res
+
+def parseComFromFile (fileName : String) :
+ IO (Option (Σ (Γ' : Context) (eff : EffectKind) (ty : InstCombine.Ty), Com LLVM Γ' eff ty)) := do
+ parseRegionFromFile fileName regionTransform
```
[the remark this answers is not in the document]
goens: I have to push back here @bollu. This is conceptually a library, that's why I split it here. The fact that it is *currently* only used by `opt` is just because we haven't implemented anything else, but this could be reused by anything that would want to parse from file.

SSA/Core/MLIRSyntax/Parser.lean
goens: Hmm, I disagree. This is not about the verified `opt` executable, this is a library about parsing. I wouldn't mind it if this had been done all in the first place and could be refactored later, but refactoring so that we can later refactor again seems silly :)

.github/workflows/ci.yml
```diff
@@ -39,14 +39,24 @@ jobs:
           lake -R exe cache get # download cache of mathlib docs.
           lake -R build AliveScaling
 
-      - name: Compile Executable 🧐
+      - name: Compile `mlirnatural` Executable 🧐
```
tobiasgrosser, line 42: Needs a more descriptive name than MLIR-natural.

.github/workflows/ci.yml
```diff
@@ -39,14 +39,24 @@ jobs:
...
         run: |
           lake -R exe cache get # download cache of mathlib docs.
           lake -R build mlirnatural
+
```
tobiasgrosser, line 46: Independent cleanup (best to be pushed separately)

.github/workflows/ci.yml
```diff
@@ -39,14 +39,24 @@ jobs:
...
       - name: LLVM Exhaustive Enumeration
         run: |
           sudo apt install llvm-15 # for opt-15, used to simplify LLVM for reference semantics.
-          cd test/bruteforce-correctness && ./run.sh
+          cd test/bruteforce-correctness && ./run.sh && cd ../../
```
tobiasgrosser, line 50:
```suggestion
          (cd test/bruteforce-correctness && ./run.sh)
```

SSA.lean
```diff
@@ -4,8 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 
 -- Core
 -- ====
-import SSA.Core.Framework
-
+import SSA.Core
```
tobiasgrosser, line 7: This and the next two changes seem a general refactor that is worth committing separately, no?

SSA/Core/MLIRSyntax/GenericParser.lean
```diff
@@ -447,18 +447,20 @@ syntax  "{" ("^" ident ("(" sepBy(mlir_bb_operand, ",") ")")? ":")? mlir_ops "}"
...
+| `(mlir_region| {  ^ $name:ident : $ops } ) => do
    let opsList <- `([mlir_ops| $ops])
    `(Region.mk $(Lean.quote (name.getId.toString)) [] $opsList)
-| `([mlir_region| { $ops:mlir_ops } ]) => do
+| `(mlir_region| { $ops:mlir_ops }) => do
    let opsList <- `([mlir_ops| $ops])
    `(Region.mk "entry" [] $opsList)
 
+macro_rules
+| `([mlir_region| $q:mlir_region ]) => `(mlir_region| $q)
```
tobiasgrosser, line 463: Is this needed for this PR or can we push this separately (to demonstrate that this is indeed an independent change).

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,132 @@
...
+import Lean
+
+open Lean
+
+variable {ParseOutput : Type} [ToString ParseOutput]
+
+abbrev ParseError := String
+abbrev ParseFun : Type := Lean.Environment → String → EIO ParseError ParseOutput
+
+-- see https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Calling.20unification.20from.20meta.20to.20remove.20metavariables
```
tobiasgrosser, line 12: If we link to a zulip thread, can we also commit a 1-2 sentence summary of the thread to the repo, such that we know why it is linked.

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,132 @@
...
+    match resE s with
+    | .ok a s => .ok a s
+    | .error exception s =>
+      let errMsgIO : IO String := exception.toMessageData.toString
+      match errMsgIO s with
+      | .ok errMsg s => .error s!"failed elaboration {stx}. Error: {errMsg}" s
+      | .error e s => .error s!" Failed elaborating {stx}.\nUnable to pretty-print exception at 'elabIntoEIO':\n{e}." s
+
+def region0Expr := (Expr.app (Expr.const `MLIR.AST.Region []) (Expr.const `Nat.zero []))
+def op0Expr := (Expr.app (Expr.const `MLIR.AST.Op []) (Expr.const `Nat.zero []))
```
tobiasgrosser, line 55: Newline?

SSA/Core/MLIRSyntax/Parser.lean
```diff
@@ -0,0 +1,132 @@
...
+    | .error msg => throw msg
+    | .ok syn => ntparser syn
+
+private def mkNonTerminalParser (syntaxcat : Name)
+    (ntparser : Syntax → EIO ParseError ParseOutput)
+    (env : Environment) (s : String) : EIO ParseError ParseOutput :=
+  let parseFun := mkParseFun syntaxcat ntparser
+  parseFun s env
+
+-- TODO: do we need this copy of the environment, or can I remove it from one of the two?
```
tobiasgrosser, line 90: Should this be resolved, still?

.github/workflows/ci.yml
goens: That's an unrelated change (renaming `mlirnatural`), since the executable already has this name and has had it for a while. This is change is just disambiguating the two executables in CI

SSA/Core/MLIRSyntax/GenericParser.lean
goens: yes, it's needed for this PR

SSA/Core/MLIRSyntax/Parser.lean
goens: are you asking if this TODO is still necessary or if we can address it and then remove it after having addressed it? I don't know how to do the latter, but if you want to take a stab at it, that'd be very helpful!

SSA.lean
goens: I feel like this relates to the PR vs commit thing. Absolutely, it should be a different commit, but if PRs and commits are separate things then it's a non-issue.

SSA/Core/MLIRSyntax/Parser.lean
goens: I don't quite understand this comment. Do you want a newline between the two definitions, or a newline at the end of the file? or do you want something else?

More generally, it sounds like a sort of style question, but we don't have an agreed-upon, fix code style. I don't particularly care for this, happy to commit a suggestion (or you can push a stylistic change that you want to)

bollu, line 57:
```suggestion

def op0Expr := (Expr.app (Expr.const `MLIR.AST.Op []) (Expr.const `Nat.zero []))
```

add newline.

bollu: resolved, we agree on mathlib style that uses a newline.

SSA/Core/MLIRSyntax/Parser.lean
bollu: @goens I'm not sure what the comment means --- I'm trying to address the TODO, but it's unclear to me what I would need to change to test this. Can I have a sentence or two explaining what environment copy I can try and remove?

SSA/Core/MLIRSyntax/Parser.lean
bollu:
```suggestion
/- see https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Calling.20unification.20from.20meta.20to.20remove.20metavariables

The "metavariable" concept consists of a couple things:
1. the `Expr.mvar` node in an expression that contains a metavariable ID
2. the table of metavariable declarations in the monad, mapping IDs to `Option Expr` (the assignments)

"Assignment" is the process of adding entries to the table, and "instantiation" is the process of replacing `Expr.mvar`s with their assigned values in a given expression.

So we first call `isDefEq` to force unification of metavars, we say `Elab.Term.synthesizeSyntheticMVarsNoPostponing` to also force lean to infer all postponed inferences, and we finally replace the old `expr` with the new instantiated expr with `instantiateMVars`.  

-/
```

.github/workflows/ci.yml
bollu: @tobiasgrosser I'm going to resolve this, since coming up with good names for these tools is beyond the scope of this PR.

lake-manifest.json
tobiasgrosser: This seems to have been resolved.

.github/workflows/ci.yml
tobiasgrosser: I guess this is a strict improvement.

SSA/Core/MLIRSyntax/Parser.lean
tobiasgrosser: Just an actionable comment would be enough. Just a bit of information that would people like @bollu enable to pick this issue in the future (or now) would be helpful.

SSA.lean
tobiasgrosser: @bollu kindly factored this out.

.github/workflows/ci.yml
tobiasgrosser: @bollu kindly factored this out.

bollu:
@tobiasgrosser this PR LGTM, I've addressed all comments, save for the TODO, which I'm happy to fixup in another PR.

SSA.lean
```diff
@@ -6,7 +6,6 @@ Released under Apache 2.0 license as described in the file LICENSE.
 -- ====
 import SSA.Core
 
-
```
tobiasgrosser, line 9: An unrelated change. This can be dropped.

SSA/Core/MLIRSyntax.lean
```diff
@@ -3,6 +3,6 @@ Released under Apache 2.0 license as described in the file LICENSE.
 -/
 import SSA.Core.MLIRSyntax.AST
 import SSA.Core.MLIRSyntax.EDSL
+import SSA.Core.MLIRSyntax.Parser
 import SSA.Core.MLIRSyntax.GenericParser
 import SSA.Core.MLIRSyntax.Transform
-
```
tobiasgrosser, line 8: An unrelated change, this removal of a new line, is not needed here.

tobiasgrosser (approved):
Nice. I just had some minor comments. It might be useful to manually squash this commit to get a good commit message (otherwise it will just be the titles of all 50 commits).

Thank you @goens @bollu! This is amazing.

SSA.lean
bollu: @tobiasgrosser I urge you to use the suggestion button :D

PR #103, opened 2023-10-07
Pull in Holor development

bollu:
We have basic algebra instances now, so it would be nice to have this in-tree.

What we do with this would be good to figure out. I can extend the API to support almost all the operations listed in the file. The only nit that I can see is things like `-inf`, which we don't have since we don't have a model of floating point. 

There is the question of what will give us most bang for buck, in terms of time spent developing. I feel @tobiasgrosser has thoughts on this.

If we want to prove theorems, then I need someone's help t find interesting theorems about these objects.

### DIFF
--- a/SSA.lean
+++ b/SSA.lean
@@ -10,6 +10,7 @@ import SSA.Core.Framework
 import SSA.Projects.InstCombine.Alive
 import SSA.Projects.Tensor1D.Tensor1D
 import SSA.Projects.Tensor2D.Tensor2D
+import SSA.Projects.Holor.Holor
 
 
 -- EXPERIMENTAL
--- a/SSA/Projects/Holor/Holor.lean
+++ b/SSA/Projects/Holor/Holor.lean
@@ -1,0 +1,3 @@
+import SSA.Core.WellTypedFramework
+import SSA.Core.Util
+import Mathlib.Data.Holor
--- a/SSA/Projects/Holor/Holor.lean
+++ b/SSA/Projects/Holor/Holor.lean
@@ -2,6 +2,7 @@ import SSA.Core.WellTypedFramework
 import SSA.Core.Util
 import Mathlib.Data.Holor
 import Mathlib.Algebra.Algebra.Basic
+import Mathlib.Algebra.Algebra.Pi
 
 /- Holor is profunctorial, covariant in the values and contavariant in the indexes. -/
 
@@ -56,8 +57,7 @@ def Holor.fill (a : α) : Holor α ds := fun _ => a
 -- I want to declare the map (r : R) -> (fill (ralg R) : Holor A ds).
 -- This will be the map that turns (Holor A ds) into an R-algebra.
 -- I'm not sure how to define this.
-instance [CommSemiring R] [Semiring A] [Algebra R A] : Algebra R (Holor A ds) :=  sorry
-
+instance [CommSemiring R] [Semiring A] [Algebra R A] : Algebra R (Holor A ds) := Pi.algebra _ _
 
 /-
 inductive op
--- a/SSA/Projects/Holor/Holor.lean
+++ b/SSA/Projects/Holor/Holor.lean
@@ -3,9 +3,7 @@ import SSA.Core.Util
 import Mathlib.Data.Holor
 import Mathlib.Algebra.Algebra.Basic
 
-#check Holor
-
-/- Holor is profunctorial, covariant in the first argument and contavariant in the second argument. -/
+/- Holor is profunctorial, covariant in the values and contavariant in the indexes. -/
 
 /-- Mapping is covariant in the value -/
 def Holor.map (f : α → β) (h : Holor α ds) : Holor β ds :=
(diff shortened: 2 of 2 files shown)


tobiasgrosser:
Does this model a particular MLIR dialect? Until know I was under the impression we have arith/Alive, scf (for regions), and poly (for FHE)? Where does Holor fit in here?

goens:
I don't think there's an scf dialect yet (but we should do that), but we have the tensor1/2D versions that do cover regions already (and this generalizes to ND, if I understand correctly)

tobiasgrosser:
I am still not sure what the relationship between holor and MLIR is? Can we name a concrete dialect?

goens:
I don't think there's an scf dialect yet (but we should do that), but we have the tensor1/2D versions that do cover regions already (and this generalizes to ND, if I understand correctly)

goens:
huh, what's wrong with [tensor](https://mlir.llvm.org/docs/Dialects/TensorOps/)? Or am I missing something here?

tobiasgrosser:
I guess everything is right with this PR, especially with respect that holor seems to work. The next question for me is now how this relates to MLIR on the dialect level, as I feel this would be useful to spin our paper story.

The tensor dialect has the following operations:

```
[tensor.bitcast (tensor::BitcastOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorbitcast-tensorbitcastop)
[tensor.cast (tensor::CastOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorcast-tensorcastop)
[tensor.collapse_shape (tensor::CollapseShapeOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorcollapse_shape-tensorcollapseshapeop)
[tensor.dim (tensor::DimOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensordim-tensordimop)
[tensor.empty (tensor::EmptyOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorempty-tensoremptyop)
[tensor.expand_shape (tensor::ExpandShapeOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorexpand_shape-tensorexpandshapeop)
[tensor.extract (tensor::ExtractOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorextract-tensorextractop)
[tensor.extract_slice (tensor::ExtractSliceOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorextract_slice-tensorextractsliceop)
[tensor.from_elements (tensor::FromElementsOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorfrom_elements-tensorfromelementsop)
[tensor.gather (tensor::GatherOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorgather-tensorgatherop)
[tensor.generate (tensor::GenerateOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorgenerate-tensorgenerateop)
[tensor.insert (tensor::InsertOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorinsert-tensorinsertop)
[tensor.insert_slice (tensor::InsertSliceOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorinsert_slice-tensorinsertsliceop)
[tensor.pack (tensor::PackOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorpack-tensorpackop)
[tensor.pad (tensor::PadOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorpad-tensorpadop)
[tensor.parallel_insert_slice (tensor::ParallelInsertSliceOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorparallel_insert_slice-tensorparallelinsertsliceop)
[tensor.rank (tensor::RankOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorrank-tensorrankop)
[tensor.reshape (tensor::ReshapeOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorreshape-tensorreshapeop)
[tensor.scatter (tensor::ScatterOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorscatter-tensorscatterop)
[tensor.splat (tensor::SplatOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorsplat-tensorsplatop)
[tensor.unpack (tensor::UnPackOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorunpack-tensorunpackop)
[tensor.yield (tensor::YieldOp)](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensoryield-tensoryieldop)
```

and @bollu's code suggests the following ops:

```
/-
inductive op
| none
| dup --Identity function : Tensor k → Tensor k
| add --Addition : Tensor k → Tensor k → Tensor k
| sub --Subtraction : Tensor k → Tensor k → Tensor k
| mul --Pointwise Multiplication : Tensor k → Tensor k → Tensor k
| div --Pointwise Division : Tensor k → Tensor k → Tensor k
| sqr --Pointwise Squaring : Tensor k → Tensor k
| sqrt --Pointwise Square Root : Tensor k → Tensor k
| sum -- Add 'em up : Tensor k → Tensor (identity of tensor product)
| mean -- Average 'em out : Tensor k → Tensor (identity of tensor product)
| repeat_ -- Complicated
| abs -- Pointwise
| sgn -- Pointwise
| neg -- Pointwise
| step -- Pointwise fun x => if x < 0 then 0 else 1
| relu -- Pointwise fun x => if x < 0 then 0 else x
| gelu -- Pointwise fun x => x * 0.5 * (1 + erf(x/sqrt(2)))
| silu -- Pointwise fun x => x * 1/ (1+exp(-x))
| norm -- not the sqrt of sum of squares
| rms_norm -- Don't know yet : Tensor k -> Tensor k rms(a) = sqrt(1/n * sum_i (a_i^2)), rms_norm i = a_i
| mul_mat -- Matrix multiplication, undefined on stuff that isn't a matrix
| scale -- Tensor 0 → Tensor k → Tensor k
| cpy -- identity function
| reshape -- Tensor → Shape → Tensor --Changes the data to fit new layout of same (linear algebra sense) dimension
| view1D -- Tensor → (Num_elements : ℕ) → (offset : ℕ) → Tensor --Output at i = input at i+offset output
| view2D -- Number of elements in 2 axes and offsets in 2 axes
| permute4D -- take a permutation of 4 numbers and Return a 4d tesnro with everything permuted
| transpose -- Matrix transpose
| get_rows -- (x : Matrix) (I : vector of Nats)  out(i,j) = x(I(i),j)
| diag_mask_inf -- Take a matrix and set upper triangle to -infinity
| soft_max -- Take a vector and exponentiate pointwise and normalize by L1-norm
| rope -- Screwed up rotatory positional embedding
| conv_1d_1s -- Polynomial multiplication of vectors
| conv_1d_2s -- Convolve 1d 2d
| flash_attn -- 3 input matrices Q K V, compute softmax (QKᵀ / sqrt(number of columns of q)) V
| flash_ff -- TODO
| count
-/
```

Yes, does not mention any dialect yet. 

The set of operations does not seem to overlap 100%, hence I would like to confirm the dialect on the MLIR side we want to model.

Also, it seems we now have 4 Projects for the evaluation. It might help me to understand which they are and what they should demonstrate.

bollu:
@tobiasgrosser The set of operations were taken from the [`ggml`](https://github.com/ggerganov/ggml) library, since it's used to run transformer models. It does not relate to any MLIR in-tree dialect in particular. 

The point, originally, was to show a use-case where we model something mathematically complex. Since we now have FHE, maybe we drop this one, and focus more on a different use-case that shows many, simple examples of regions. 

We should take a call.

tobiasgrosser:
Is there an MLIR dialect that covers (a subset) of these operations?

bollu:
I would imagine an out of tree dialect like tensorflow or IREE would.

tobiasgrosser:
What about linalg, hlo, ...? Maybe ask Kunwar? I think need a pre-existing dialect to be able to make the case that these operations are actively used in the MLIR ecosystem, no?

bollu:
It depends on what case we want to make, right? If we are writing a general framework to express compiler IRs, it's okay to have an example that's not from an already existing MLIR dialect, and rather from a different compiler IR.

bollu:
I don't think `linalg` has support for any of the specialized operations like `relu` or `flash_attn`. HLO will contain these, as HLO is tensorflow's IR. [`relu` is indeed available in their out of tree implementation](https://github.com/tensorflow/mlir-hlo/blob/e53cb6966f10711c47927bfc82d36c5b3b564a6d/lhlo_gpu/IR/lhlo_gpu_ops_enums.td#L28). 

We need to careful about supporting HLO. Their IR contains information about the compute graph as well IIRC, and has non-trivial semantics.

bollu:
@goens @alexkeizer : My suggestion is that we pull this in, since it adds math stuff, and then in the paper, we *discuss* holor, but we do not say that we implement it.

PR #107, opened 2023-10-18
chore: nuke unused MLIRNatural runner

bollu:
Quick chore cleanup to make sure we don't have dead code.


goens:
This is not dead code. On the contrary should actually use this more to test that our semantics actually match LLVM's! As we discussed, maybe we should just rename this to `interpreter` or something

PR #109, opened 2023-10-18
Dead Code Elimination

bollu:
An implementation of dead code elimination for our framework.

This implements a naive O(n^2) forward pass algorithm for dead
code elimination.

I wrote it for two purposes:
1. To get used to, and understand precisely, our framework for
   manipulating programs, and
2. To have the three (in my view) cornerstones of SSA:
   peephole rewrites on the def/use chain, *dead code elimination*,
   and common subexpression elimination.

The implementation carries a proof of semantics preservation.
The unfortunate part: I'm not sure how to run it on a program
from within Lean (without using the Lean compiler), since:
- DCE uses `partial def` which cannot be unfolded in proof mode.
- `decide` cannot be used as we do not have `DecidableEq` on `ICom`.
- `#reduce` gets stuck on well founded induction.


SSA/Projects/DCE/DCE.lean
```diff
@@ -0,0 +1,492 @@
...
+
+attribute [local simp] Ctxt.snoc
+
+def ex1_pre_dce : ICom ExOp ∅ .nat :=
+  ICom.lete (cst 1) <|
+  ICom.lete (cst 2) <|
+  ICom.ret ⟨0, by simp [Ctxt.snoc]⟩
+
+/-- TODO: how do we evaluate 'ex1_post_dce' within Lean? :D -/
+def ex1_post_dce : ICom ExOp ∅ .nat := (dce' ex1_pre_dce).val
```
bollu, line 484: Can someone tell me how I actually evaluate this definition?

alexkeizer: Or do you mean, how do we actually reason about the result of `dce'`? That is a bit more difficult

SSA/Projects/DCE/DCE.lean
```diff
@@ -0,0 +1,492 @@
+import SSA.Core.Framework
+
+/-- Delete a variable from a list. -/
+def Ctxt.delete (Γ : Ctxt Ty) (v : Γ.Var α) : Ctxt Ty :=
+  Γ.removeNth v.val
+
+/-- Witness that Γ' is Γ without v -/
+def Deleted {α : Ty} (Γ: Ctxt Ty) (v : Γ.Var α) (Γ' : Ctxt Ty) : Prop :=
```
ChrisHughes24, line 8: This strikes me as potentially evil in the same way that equalities of types are evil. It may be better to think about replacing this definition with the data of a universal property. It may still be necessary though. But in my head a `Ctxt` is morally just `Ty -> Type`, but I know we're not quite doing that.

SSA/Projects/DCE/DCE.lean
```diff
@@ -0,0 +1,492 @@
...
+    induction xs
+    case nil => simp at hn
+    case cons x xs' IH =>
+      simp[removeNth_succ]
+      apply IH'
+      simp at hn
+      exact hn
+
+/-- removing at any index `≥ xs.length` does not change the list. -/
+theorem List.removeNth_gt_len (hn : xs.length ≤ n) : List.removeNth xs n = xs := by
```
ChrisHughes24, line 35:
```suggestion
theorem List.removeNth_of_length_le (hn : xs.length ≤ n) : List.removeNth xs n = xs := by
```

SSA/Projects/DCE/DCE.lean
```diff
@@ -0,0 +1,492 @@
...
+
+/-- build a `Deleted` for a `(Γ.snoc α) → Γ`-/
+def Deleted.deleteSnoc (Γ : Ctxt Ty) (α : Ty) : Deleted (Γ.snoc α) (Ctxt.Var.last Γ α) Γ := rfl
+
+theorem List.removeNth_zero : List.removeNth (List.cons x xs) 0 = xs := rfl
+
+theorem List.removeNth_succ : List.removeNth (List.cons x xs) (.succ n) = x :: List.removeNth xs n := rfl
+
+/- removing from `xs ++ [x]` at index `(length xs)` equals `xs`. -/
+theorem List.removeNth_eq_len_snoc (hn : n = xs.length) : List.removeNth (xs ++ [x]) n = xs := by
```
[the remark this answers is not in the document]
ChrisHughes24: Also, why not `List.removeNth (xs ++ [x]) xs.length` I have a feeling it's because there's some `motive is not type correct issue` somewhere. But that's why you should try to use morphisms of contexts and not equalities.

SSA/Projects/DCE/DCE.lean
```diff
@@ -0,0 +1,492 @@
...
+    induction xs
+    case nil => simp[removeNth]
+    case cons x xs' IH =>
+      simp[removeNth_succ]
+      apply IH'
+      simp at hn
+      linarith
+
+/- removing at index `n` does not change indices `k < n` -/
+theorem List.get_removeNth_lt_n (hk: k < n) : List.get? (List.removeNth xs n) k = List.get? xs k := by
```
ChrisHughes24, line 51:
```suggestion
theorem List.get?_removeNth_of_lt (hk: k < n) : List.get? (List.removeNth xs n) k = List.get? xs k := by
```

SSA/Projects/DCE/DCE.lean
```diff
@@ -0,0 +1,492 @@
...
+      case zero => simp
+      case succ k' =>
+        simp[List.get?]
+        apply IHxs
+        linarith
+        simp at N_LEN; linarith
+
+
+/-- Removing index `n` shifts entires of `k ≥ n` by 1. -/
+theorem List.get_removeNth_geq_n {xs : List α} {n : Nat} {k : Nat} (hk: n ≤ k) :
```
ChrisHughes24, line 73:
```suggestion
theorem List.get?_removeNth_of_le {xs : List α} {n : Nat} {k : Nat} (hk: n ≤ k) :
```

ChrisHughes24 (commented):
I'll also have a closer look in VSCode. Just some initial comments

SSA/Projects/DCE/DCE.lean
ChrisHughes24: You can use `decreasing_by sorry` at the end instead of `partial def`, and then you will be able to unfold things.

bollu: That doesn't work, see that there's a commented out block of code doing exactly this marked as `Lean bug`.

SSA/Core/Framework.lean
```diff
@@ -1568,3 +1568,35 @@ def p2 : PeepholeRewrite ExOp [.nat] .nat:=
   }
 
 end RegionExamples
+
+section Unfoldings
+
+/-- Equation lemma to unfold `denote`, which does not unfold correctly due to the presence
```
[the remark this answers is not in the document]
bollu: It is the exact pattern match, but Lean doesn't seem to generate/use the right auto-generated equation lemma to unfold it, which is why I had to do this.

SSA/Projects/DCE/DCE.lean
bollu: Hm, what ought the universal property be?

ChrisHughes24: Basically it's `Option`. So the universal property is

The map `Γ -> Γ'`. This is the unit of the adjunction.
The map `extend` that extends a morphism `Γ -> Δ` and a `w : Δ.var α` to a map `Γ' -> Δ`.
The fact that `(extend f w).comp unit = f`, and `(extend f w) v = w`
The fact that `extend unit v = id`, or alternative the fact that two maps `Γ' -> Δ` are equal if they're equal after composition with `unit` and application at `v`

But anything simpler that's equivalent to this could work. Like saying there's an injective map and all but one `Var` is in the image of it.

alexkeizer: Hmm, that seems a little overkill. Would it be enough to have the property be that `Γ'` is isomorphic to `Γ.delete v`? We already have a notion of context morphism, so isomorphism should not be too much extra abstraction

alexkeizer, line 8: `Deleted` should be in a namespace, say `Ctxt.Deleted`

SSA/Core/ErasedContext.lean
```diff
@@ -206,9 +214,38 @@ def Valuation.snoc {Γ : Ctxt Ty} {t : Ty} (s : Γ.Valuation) (x : toType t) :
   intro t' v
   revert s x
   refine Ctxt.Var.casesOn v ?_ ?_
-  . intro _ _ _ v s _; exact s v
+  . intro _ _ _ v s _;  exact s v
```
alexkeizer, line 217:
```suggestion
  . intro _ _ _ v s _; exact s v
```
spurious whitespace

SSA/Projects/DCE/DCE.lean
ChrisHughes24: Yeah, that's equivalent I guess. I'm not saying it has to be done that way by the way. But you probably want to know most of the things I just said about delete about it anyway.

SSA/Projects/DCE/DCE.lean
```diff
@@ -0,0 +1,492 @@
...
+        linarith
+
+/-- Lean bug: linarith fails to find a contradiction -/
+theorem linarith_failure (x y : Nat) (H1 : ¬ (x = y)) (H2 : ¬ (x < y)) : x > y :=
+  by
+    try linarith
+    sorry
+
+/-- Given  `Γ' := Γ /delv`, transport a variable from `Γ'` to `Γ`. -/
+def Deleted.pullback_var (DEL : Deleted Γ delv Γ') (v : Γ'.Var β) : Γ.Var β :=
```
alexkeizer, line 100: Why have this go through `Deleted`, instead of direclty writing `(Γ.delete v).Var β -> Γ.Var β`?

bollu: Because `Deleted` acts as a hook for keeping these theorems.

SSA/Core/ErasedContext.lean
```diff
@@ -162,6 +162,10 @@ abbrev Hom (Γ Γ' : Ctxt Ty) := ⦃t : Ty⦄ → Γ.Var t → Γ'.Var t
 abbrev Hom.id {Γ : Ctxt Ty} : Γ.Hom Γ :=
   fun _ v => v
 
+/-- `f.composeRange g := g(f(x))` -/
+def Hom.composeRange {Γ Γ' Γ'' : Ctxt Ty} (self : Hom Γ Γ') (rangeMap : Hom Γ' Γ'') : Hom Γ Γ'' :=
```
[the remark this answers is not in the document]
bollu: Renamed to `Hom.comp`

SSA/Core/ErasedContext.lean
```diff
@@ -198,6 +202,10 @@ variable [Goedel Ty] -- for a valuation, we need to evaluate the Lean `Type` cor
 def Valuation (Γ : Ctxt Ty) : Type :=
   ⦃t : Ty⦄ → Γ.Var t → (toType t)
 
+/-- A valuation for a context. Provide a way to evaluate every variable in a context. -/
+def Valuation.eval {Γ : Ctxt Ty} (VAL : Valuation Γ) ⦃t : Ty⦄ (v : Γ.Var t) : toType t :=
```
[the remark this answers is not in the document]
bollu: I'd like to keep it then, it sometimes makes the evaluation more explicit.

SSA/Projects/DCE/DCE.lean
bollu: I propose we refactor this in the next commit. I agree that this ought to be moved.

SSA/Core/ErasedContext.lean
```diff
@@ -224,6 +261,14 @@ def Valuation.ofHVector {types : List Ty} : HVector toType types → Valuation (
   | .nil => (default : Ctxt.Valuation (∅ : Ctxt Ty))
   | .cons x xs => (Valuation.ofHVector xs).snoc x
 
+/-- transport/pullback a valuation along a context homomorphism. -/
+def Valuation.hom {Γi Γo : Ctxt Ty} (Γiv: Γi.Valuation) (hom : Ctxt.Hom Γo Γi) : Γo.Valuation :=
+  fun _to vo =>  Γiv (hom vo)
+  
+/-- abbreviation for Valuation.hom -/
+abbrev Valuation.pulback {Γi Γo : Ctxt Ty} (Γiv: Γi.Valuation) (hom : Ctxt.Hom Γo Γi) : Γo.Valuation :=
```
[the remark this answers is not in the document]
bollu: Renamed,

SSA/Core/ErasedContext.lean
```diff
@@ -224,6 +261,14 @@ def Valuation.ofHVector {types : List Ty} : HVector toType types → Valuation (
   | .nil => (default : Ctxt.Valuation (∅ : Ctxt Ty))
   | .cons x xs => (Valuation.ofHVector xs).snoc x
 
+/-- transport/pullback a valuation along a context homomorphism. -/
+def Valuation.hom {Γi Γo : Ctxt Ty} (Γiv: Γi.Valuation) (hom : Ctxt.Hom Γo Γi) : Γo.Valuation :=
```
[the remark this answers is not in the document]
bollu: Renamed.

SSA/Projects/DCE/DCE.lean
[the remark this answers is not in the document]
bollu: The style with `n` is easier to rewrite with, as it creates another proof obligation to show that `<something> = xs.length` instead of failing on the rewrite. I prefer to have terms that are "linear" (ie, no repeated patterns) and proof obligations to show that stuff that needs to be equal is equal.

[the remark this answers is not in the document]
bollu: I changed the signature as requested.

PR #110, opened 2023-10-18
Move Alive statements to new Std.BitVec

tobiasgrosser:
Std4 gained a new BitVec library which we want to use as a foundation for our proof development. This PR transitions the AliveStatement.lean file to the new BitVec library to (a) see if it is expressive enough and (b) see if the decision procedures already work.


bollu:
LGTM, thanks for the port! Merging this now as CI succeeded.

PR #121, opened 2023-10-25
feat: add examples from paper as a project

bollu:
This adds a new subproject, containing the simple examples from the paper to ensure that what we write in the paper is type correct.


goens:
LGTM

There's some "noise" in the `simp_peephole` tactic (I personally think splitting that into a separate PR is overkill).

PR #123, opened 2023-10-26
Eliminate dot-notation for `Var` and `Valuation` types

alexkeizer:
Using dot-notation might be confusing in the paper, especially since we write `Expr \G t` / `Com \G t`, so this PR proposes to change `\G.Var t` to `Var \G t` to be more consistent


goens:
LGTM

PR #125, opened 2023-10-26
Remove `snoc'`

alexkeizer:
The core issue was `Valuation.snoc_last` not applying
because one context was spelled \emptyset, while the other was spelled `[]`
By being more consistent, and introducing a simp-lemma that rewrites the former to the latter
we remove the need for `snoc'`.


bollu:
LGTM, thanks for the cleanup!

PR #127, opened 2023-10-31
Common Subexpression Elimination

bollu:
This patch implements naive common subexpression elimination, which walks the `Com` data structure top-down, keeping a canonical mapping from canonical expressions to their representative variable.

This is not the best we can do: The best pass would simultaneously eliminate the old, dead expression. 
This is the same trade-off the rewriter chooses to perform, where we only make the change, but don't eliminate the old dead value.

The algorithm has only been proven correct, but has not been tested on any examples 😉 I need to add tests to the file before it is reviewable.


goens:
why are there so many repeated commits from `main` with different hashes? It sounds like a broken rebase?

PR #130, opened 2023-11-07
Llvm sem

goens:
Check our denotation with the LLVM semantics (and document those). 

The only thing I'm not 100% sure is what happens with constants when they exceed the bitwidth (will check that with examples later)


SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -0,0 +1,283 @@
...
+
+/--
+Shift left instruction.
+The value produced is op1 * 2op2 mod 2n, where n is the width of the result.
+If op2 is (statically or dynamically) equal to or larger than the number of
+bits in op1, this instruction returns a poison value.
+-/
+def shl? {m n k} (op1 : BitVec n) (op2 : BitVec m) : Option (BitVec k) :=
+  let bits := op2.toNat -- should this be toInt?
+  if bits >= n then .none
```
bollu, line 135: shouldn't this be `bits >= 2^n` ?

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -0,0 +1,283 @@
...
+the shift.
+
+If op2 is (statically or dynamically) equal to or larger than the number of bits in op1,
+this instruction returns a poison value.
+
+Corresponds to `Std.BitVec.ushiftRight` in the `some` case.
+-/
+def lshr? {m n k} (op1 : BitVec n) (op2 : BitVec m) : Option (BitVec k) :=
+  let bits := op2.toNat -- should this be toInt?
+  if bits >= n then .none
```
bollu, line 150: same question, isn't it `bits >= 2^n`

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -0,0 +1,283 @@
...
+The most significant bits of the result will be filled with the sign bit of op1.
+
+If op2 is (statically or dynamically) equal to or larger than the number of bits in op1,
+this instruction returns a poison value.
+
+Corresponds to `Std.BitVec.sshiftRight` in the `some` case.
+-/
+def ashr? {m n k} (op1 : BitVec n) (op2 : BitVec m) : Option (BitVec k) :=
+  let bits := op2.toNat -- should this be toInt?
+  if bits >= n then .none
```
bollu, line 165: `bits >= 2^n`?

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -0,0 +1,283 @@
...
+Note that signed integer division and unsigned integer division are distinct operations; for unsigned integer division, use ‘udiv’.
+Division by zero is undefined behavior.
+Overflow also leads to undefined behavior; this is a rare case, but can occur, for example, by doing a 32-bit division of -2147483648 by -1.
+-/
+def sdiv? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
+  if y.toInt = 0
+  then none
+  else
+    let div := (x.toInt / y.toInt)
+    if div < Int.ofNat (2^w)
```
bollu, line 82: @tobiasgrosser is the only case where division can overflow is when performing, e.g.
> 32-bit division of -2147483648 by -1.

? If so, I would like for us to simplify the check to check that specific combination of values (`x = -2^n && y = -1`), since that will make proofs much easier.

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -0,0 +1,283 @@
...
+of the division and the remainder.)
+-/
+def srem? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
+  if y.toInt = 0
+  then none
+  else
+    let div := (x.toInt / y.toInt)
+    if div < Int.ofNat (2^w)
+      then some $ BitVec.ofInt w (x.toInt.rem y.toInt)
+      else none
```
bollu, line 123: This should be reviewed carefully, but I lack the mental acuity today to do so. I'm postponing this to tomorrow, I'm afraid :)

bollu (commented):
Thanks a lot for doing the hard work of checking these andres! I have a single flavour of question about the `bits`, and one that I have left unreviewed because I don't trust myself enough tonight.

SSA/Projects/InstCombine/LLVM/Semantics.lean
goens: I believe `n` is right here, `2^n` is the maximal number that is representable by `n` bits, but since this is a shift, it corresponds to a multiplication by `2^op`, so we do want the logarithm of that multiplication, bits, to be less than the number of bits in `op1`. The power was lost through the formatting in the copy-paste, I fixed the docstring.

SSA/Projects/InstCombine/LLVM/Semantics.lean
goens: same as above

SSA/Projects/InstCombine/LLVM/Semantics.lean
goens: I'm actually not comfortable with changing the *specification* to be easier to prove things. If at all we should add a theorem saying this equivalent version is equivalent, don't you think? That way we can use that characterization in any proofs, maybe even tag it with `@[simp]`, but not change the definition which should be the more general specification

SSA/Projects/InstCombine/LLVM/Semantics.lean
bollu: I believe the explanation now, thanks!

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -0,0 +1,283 @@
...
+Taking the remainder of a division by zero is undefined behavior.
+For vectors, if any element of the divisor is zero, the operation has undefined behavior.
+Overflow also leads to undefined behavior; this is a rare case, but can occur, for example,
+by taking the remainder of a 32-bit division of -2147483648 by -1.
+(The remainder doesn’t actually overflow, but this rule lets srem be implemented using instructions that return both the result
+of the division and the remainder.)
+-/
+def srem? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
+  if y.toInt = 0
+  then none
```
bollu, line 143:
```suggestion
  then none -- Taking the remainder of a division by zero is undefined behavior.
```

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -0,0 +1,283 @@
...
+  else
+    let div := (x.toInt / y.toInt)
+    if div < Int.ofNat (2^w)
+      then some $ BitVec.ofInt w (x.toInt.rem y.toInt)
```
bollu, line 147:
```suggestion
      then some $ BitVec.ofInt w (x.toInt.rem y.toInt) Overflow also leads to undefined behavior; this is a rare case, but can occur
```

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -0,0 +1,283 @@
...
+by taking the remainder of a 32-bit division of -2147483648 by -1.
+(The remainder doesn’t actually overflow, but this rule lets srem be implemented using instructions that return both the result
+of the division and the remainder.)
+-/
+def srem? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
+  if y.toInt = 0
+  then none
+  else
+    let div := (x.toInt / y.toInt)
+    if div < Int.ofNat (2^w)
```
bollu, line 146: Do we not need to check overflow on the negative side as well?
```suggestion
    if div < Int.ofNat (2^w) && div >= - (Int.ofNat (2^w))
```

SSA/Projects/InstCombine/LLVM/Semantics.lean
bollu: I agree with this argument,

bollu, line 81: Do we not need to check the other side of overflow as well?

```suggestion
    if div < Int.ofNat (2^w) && div >= - (Int.ofNat (2^w))
```

bollu (commented):
I believe your answers to my previous questions, and left one more type of question: I wonder if we should also check for negative overflow, with `div >= - Int.ofNat (2^w)`

SSA/Projects/InstCombine/LLVM/Semantics.lean
goens: @bollu good point! but then it should be `< intMin` or `> intMax`, right? (in particular, that's `2^(w-1)`, no?)

SSA/Projects/InstCombine/LLVM/Semantics.lean
goens: same as above

SSA/Projects/InstCombine/LLVM/Semantics.lean
bollu: You're right, in 2s complement, something of width `w` goes from `[-2^{w-1}, 2^{w-1)`

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -67,6 +67,16 @@ def udiv? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
     | 0 => none
     | _ => some <| BitVec.ofInt w (x.toNat / y.toNat)
 
+def intMin (w : Nat) : BitVec w :=
+  BitVec.ofInt w <| 1 - ↑(2^(w-1))
+
+def intMax (w : Nat) : BitVec w :=
+  BitVec.ofInt w ↑(2^(w-1))
```
bollu, line 74:
```suggestion
  BitVec.ofInt w ↑(2^(w-1) - 1)
```

Unless I’m missing something?

bollu (approved):
I wonder if there is some kind of unit test we can write for these, since it seems quite error prone to get these right.

It all looks consistent to me now, so feel invited to merge!

goens:
yeah, ideally we should use the mlirnatural entry point and run a bunch of tests to compare with LLVM/MLIR. I'll add an issue for that.

goens:
> I wonder if there is some kind of unit test we can write for these, since it seems quite error prone to get these right.
> 
> It all looks consistent to me now, so feel invited to merge!

yeah, ideally we should use the mlirnatural entry point and run a bunch of tests to compare with LLVM/MLIR. I'll add an issue for that.

PR #131, opened 2023-11-08
Add alive example from paper

goens:
This example is not in the AutoGenerated, but it is what we use to motivate undefined behavior. Adding it here so we can add this code to the paper too.


goens:
oh nvm I just noticed I miscopied the RHS, will fix that

bollu:
@goens hit merge when you're happy with it, I eyeballed the change and it looks sane to me.

PR #133, opened 2023-11-13
Scf.if, Scf.for operations.

bollu:


SSA/Projects/Scf/Scf.lean
```diff
@@ -0,0 +1,175 @@
...
+
+/-- 'if' condition of a true variable evaluates to the true region body. -/
+theorem if_true {t : Ty} (cond : Var Γ .bool) (hcond : Γv cond = true) (v : Var Γ t) (then_ else_ : Com Op [t] t) :
+  Expr.denote (ScfRegion.if_ (t := t) cond v then_ else_) Γv =
+  Expr.denote (ScfRegion.run (t := t) v then_) Γv := by
+    simp[Expr.denote, if_, run]
+    simp_peephole[hcond] at Γv
+    simp[ite_true]
+
+/-- 'if' condition of a true variable evaluates to the true region body. -/
```
tobiasgrosser, line 130:
```suggestion
/-- 'if' condition of a false variable evaluates to the else region body. -/
```

SSA/Projects/Scf/Scf.lean
```diff
@@ -0,0 +1,175 @@
...
+    (regArgs := HVector.cons body <| HVector.nil)
+
+def for_ {Γ : Ctxt _} {t : Ty} (niter : Var Γ .nat) (v : Var Γ t) (body : Com Op [.nat, t] t) : Expr Op Γ t :=
+  Expr.mk
+    (op := .for t)
+    (ty_eq := rfl)
+    (args := .cons niter <| .cons v .nil)
+    (regArgs := HVector.cons body <| HVector.nil)
+
+/-- 'if' condition of a true variable evaluates to the true region body. -/
```
tobiasgrosser, line 122:
```suggestion
/-- 'if' condition of a true variable evaluates to the then region body. -/
```

SSA/Projects/Scf/Scf.lean
```diff
@@ -0,0 +1,175 @@
...
+/-- running `f(x) = x + x` 0 times is the identity. -/
+def lhs : Com Op [.int] .int :=
+  Com.lete (iterate (k := 0) ⟨0, by simp[Ctxt.snoc]⟩ (
+      Com.lete (add ⟨0, by simp[Ctxt.snoc]⟩ ⟨0, by simp[Ctxt.snoc]⟩) -- fun x => (x + x)
+      <| Com.ret ⟨0, by simp[Ctxt.snoc]⟩
+  )) <|
+  Com.ret ⟨0, by simp[Ctxt.snoc]⟩
+
+def rhs : Com Op [.int] .int :=
+  Com.ret ⟨0, by simp[Ctxt.snoc]⟩
```
tobiasgrosser, line 149: Is there a reason this is not defined in MLIR syntax?

SSA/Projects/Scf/Scf.lean
```diff
@@ -0,0 +1,175 @@
...
+              (Com.ret { val := 0, property := lhs.proof_2 })))
+          (Com.ret { val := 0, property := lhs.proof_2 }))
+        Γv =
+      Com.denote (Com.ret { val := 0, property := rhs.proof_1 }) Γv
+      -/
+      simp_peephole [add, iterate] at Γv
+      /-  ∀ (a : BitVec 32), (fun v => v + v)^[0] a = a -/
+      simp [Function.iterate_zero]
+      done
+  }
```
tobiasgrosser, line 173: Can we define a similar peephole rewrite for the "if true" and "if false" rewrites?

tobiasgrosser: We could also hardcode the region, and later replace it with a mvar.

SSA/Projects/Scf/Scf.lean
tobiasgrosser: #128 is green. Is sth missing?

SSA/Projects/Scf/Scf.lean
```diff
@@ -0,0 +1,456 @@
...
+
+/-- evaluating a function that does not access the index (const_index_fn) -/
+theorem loop_conter_decorator_const_index_fn_eval
+  (δ : Int) (i : Int) (vstart : α) (f : Int → α → α) (f' : α → α) (hf : f = fun i a => f' a) :
+  (loop_counter_decorator δ f) (i, vstart) = (i + δ, f' vstart) := by
+  simp[loop_counter_decorator, hf]
+
+
+/-- iterating a function that does not access the index (const_index_fn) -/
+theorem loop_conter_decoraor_const_index_fn_iterate
```
tobiasgrosser, line 72: decorator?

SSA/Projects/Scf/Scf.lean
```diff
@@ -0,0 +1,456 @@
...
+    rw[loop_counter_decorator_constant_iterate]
+
+
+set_option pp.proofs false
+set_option pp.proofs.withType false
+
+/-# Repeatedly adding a constant in a loop is replaced with a multiplication.
+
+We keep the increment outside the loop so that we don't need to deal with creating and deleting tuples for the "for" region body,
+since our regions are isolatedFromAbove.
```
tobiasgrosser, line 240: How can we keep the increment outside, if the region is isolated from above?

SSA/Projects/Scf/Scf.lean
```diff
@@ -0,0 +1,456 @@
...
+    /- we still have things like
+    Ctxt.Valuation.snoc Γv (b * ↑c + a) { val := 2, property := rhs.proof_1 }
+    in the proof state, because it's not sure how to simplify stuff with `snoc`? -/
+    simp[Goedel.toType] at a b c d
+    have h1 : Ctxt.Valuation.snoc (t := .int) Γv (b * (c : ℤ) + a) { val := 2, property := rhs.proof_1 } = b := by
+      conv => rhs; rw[← B]
+    simp[h1]
+    have h2 : Ctxt.Valuation.snoc Γv (t := .int) (b * ↑c + a) { val := 3, property := rhs.proof_2 } = c := by
+      conv => rhs; rw[← C]
+    sorry
```
tobiasgrosser, line 362: What about this sorry?

SSA/Projects/Scf/Scf.lean
```diff
@@ -0,0 +1,456 @@
...
+  /- niters1 + niters2 = -/ Com.lete (cst_nat <| niters1 + niters2) <|
+  /- start1 = -/ Com.lete (cst start1) <|
+  /- c1 = -/ Com.lete (cst 1) <|
+  -- start step niter v
+  Com.lete (for_ (t := t) ⟨1, by simp[Ctxt.snoc]⟩ ⟨0, by simp[Ctxt.snoc]⟩ ⟨2, by simp[Ctxt.snoc]⟩ ⟨3, by simp[Ctxt.snoc]⟩ rgn) <|
+  Com.ret ⟨0, by simp[Ctxt.snoc]⟩
+#check rhs
+
+theorem correct :
+  Com.denote (lhs rgn niters1 ninters2 start1) Γv = Com.denote (rhs rgn niters1 ninters2 start1) Γv := by sorry
```
tobiasgrosser, line 410: And this one?

SSA/Projects/Scf/Scf.lean
tobiasgrosser: What about this comment. At the very least, it would be nice to have an end-to-end rewrite that shows the MLIR syntax if the lhs/rhs things dont work.

SSA/Projects/Scf/Scf.lean
tobiasgrosser: What about this comment?

SSA/Projects/Scf/Scf.lean
bollu: > How can we keep the increment outside, if the region is isolated from above?

The increment variable isn't in the IR, it's a "lean variable". I'll change the proof to bring it into the IR in the second pass.

SSA/Projects/Scf/Scf.lean
bollu: That's a different proofs, which I will work on next. Note that the proof of the loop transformation in the intro does not use `sorry` (from the `#print axioms correct`, which **does not list** `sorryAx`)

bollu:
@tobiasgrosser , I've added TODOs for missing stuff. 

because our regions are `isolatedFromAbove`, for the rewrite that converts

```
%res2 = scf.for(0, %n, %res) {
  ^entry(%res): 
    %out = arith.add %res, %k -- %k is outside the loop body!
    scf.yield %out
}
```

into
 
```
%nk = arith.mul %n, %k
%out = arith.add %res, %nk
```

we will need some tupling and untupling functions, because right now, from an `scf.for`, we can only pass a single value into the region, but we need to use `%k` (which is outside the region) inside the region.

so, our proved theorem will say that the LHS:

```
%tuple = tuple.mk(%k, %res)
%res2 = scf.for(0, %n, %tuple) {
  ^entry(%tuple): 
     %k = tuple.first %tuple
     %res = tuple.snd %res
    %out_res = arith.add %res, %k -- %k is outside the loop body!
    %out = tuple.mk (%k, %out_res)
    scf.yield %out
}
```

is equivalent to the same RHS as before:

```
%nk = arith.mul %n, %k
%out = arith.add %res, %nk
```

tobiasgrosser:
It's OK to merge with the TODOs I believe.

PR #140, opened 2023-11-15
feat: always test AliveAutoGenerated

bollu:
Add a #exit into the file, but always test it via SSA.Alive. This ensures that we don't accidentally break alive code generation.


tobiasgrosser (approved):
This is fine with me.

PR #141, opened 2023-11-15
Scf for fusion

bollu:
Adjacent for loops can be fused, assuming that the second loop ends where the first one starts.

### DIFF
--- a/SSA/Core/ErasedContext.lean
+++ b/SSA/Core/ErasedContext.lean
@@ -250,6 +250,16 @@ theorem Valuation.snoc_toSnoc {Γ : Ctxt Ty} {t t' : Ty} (s : Γ.Valuation) (x :
     (v : Γ.Var t') : (s.snoc x) v.toSnoc = s v := by
   simp [Ctxt.Valuation.snoc]
 
+/-!
+# Helper to simplify context manipulation with toSnoc and variable access.
+-/
+/--  (ctx.snoc v₁) ⟨n+1, _⟩ = ctx ⟨n, _⟩ -/
+@[simp]
+theorem Valuation.snoc_eval {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
+    (hvar : Ctxt.get? (Ctxt.snoc Γ ty) (n+1) = some var_val) :
+    (V.snoc v) ⟨n+1, hvar⟩ = V ⟨n, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ :=
+  rfl
+
 /-- Make a a valuation for a singleton value -/
 def Valuation.singleton {t : Ty} (v : toType t) : Ctxt.Valuation [t] :=
   Ctxt.Valuation.nil.snoc v
--- a/SSA/Projects/Scf/Scf.lean
+++ b/SSA/Projects/Scf/Scf.lean
@@ -324,61 +324,14 @@ section ValuationVariableAccess
 
 /-- (ctx.snoc v₁) ⟨1, _⟩ = ctx ⟨0, _⟩ -/
 @[simp]
-theorem Ctxt.Valuation.snoc_eval_one {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
-    (hvar : Ctxt.get? (Ctxt.snoc Γ ty) 1 = some var_val) :
-    (V.snoc v) ⟨1, hvar⟩ = V ⟨0, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ :=
+theorem Ctxt.Valuation.snoc_eval {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
+    (hvar : Ctxt.get? (Ctxt.snoc Γ ty) (n+1) = some var_val) :
+    (V.snoc v) ⟨n+1, hvar⟩ = V ⟨n, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ :=
   rfl
 
-/-- (ctx.snoc v₁) ⟨2, _⟩ = ctx ⟨1, _⟩ -/
-@[simp]
-theorem Ctxt.Valuation.snoc_eval_two {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
-    (hvar : Ctxt.get? (Ctxt.snoc Γ ty) 2 = some var_val) :
-    (V.snoc v) ⟨2, hvar⟩ = V ⟨1, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ :=
-  rfl
 
-/-- (ctx.snoc v₁) ⟨3, _⟩ = ctx ⟨2, _⟩ -/
-@[simp]
-theorem Ctxt.Valuation.snoc_eval_three {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
-    (hvar : Ctxt.get? (Ctxt.snoc Γ ty) 3 = some var_val) :
-    (V.snoc v) ⟨3, hvar⟩ = V ⟨2, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ :=
-  rfl
-
-/-- (ctx.snoc v₁) ⟨4, _⟩ = ctx ⟨3, _⟩ -/
-@[simp]
-theorem Ctxt.Valuation.snoc_eval_four {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
-    (hvar : Ctxt.get? (Ctxt.snoc Γ ty) 4 = some var_val) :
-    (V.snoc v) ⟨4, hvar⟩ = V ⟨3, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ :=
-  rfl
-
-/-- (ctx.snoc v₁) ⟨5, _⟩ = ctx ⟨4, _⟩ -/
-@[simp]
-theorem Ctxt.Valuation.snoc_eval_five {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
-    (hvar : Ctxt.get? (Ctxt.snoc Γ ty) 5 = some var_val) :
-    (V.snoc v) ⟨5, hvar⟩ = V ⟨4, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ :=
-  rfl
-
-/-- (ctx.snoc v₁) ⟨6, _⟩ = ctx ⟨5, _⟩ -/
-@[simp]
-theorem Ctxt.Valuation.snoc_eval_six {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
-    (hvar : Ctxt.get? (Ctxt.snoc Γ ty) 6 = some var_val) :
-    (V.snoc v) ⟨6, hvar⟩ = V ⟨5, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ :=
-  rfl
-
-/-- ((ctx.snoc v₁).snoc v₂) ⟨2, _⟩ = ctx ⟨0, _⟩ -/
-@[simp]
-theorem Ctxt.Valuation.snoc_snoc_eval_two {ty₁ ty₂: Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v₁ : ⟦ty₁⟧) (v₂ : ⟦ty₂⟧)
-    (hvar : Ctxt.get? ((Ctxt.snoc Γ ty₁).snoc ty₂) 2 = some var_val) :
-    ((V.snoc v₁).snoc v₂) ⟨2, hvar⟩ = V ⟨0, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ :=
-  rfl
-
-/-- (((ctx.snoc v₁).snoc v₂).snoc v₃) ⟨3, _⟩ = ctx ⟨0, _⟩ -/
-@[simp]
-theorem Ctxt.Valuation.snoc_snoc_snoc_eval_three {ty₁ ty₂ ty₃: Ty} (Γ : Ctxt Ty) (V : Γ.Valuation)
-    (v₁ : ⟦ty₁⟧) (v₂ : ⟦ty₂⟧) (v₃ : ⟦ty₃⟧)
-    (hvar : Ctxt.get? (((Ctxt.snoc Γ ty₁).snoc ty₂).snoc ty₃) 3 = some var_val) :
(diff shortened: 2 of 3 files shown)


SSA/Projects/Scf/Scf.lean
```diff
@@ -56,6 +56,18 @@ instance : OpSignature Op Ty where
...
 
+/-- iterated value of the fst of the tuple of loop_counter_decorator (ie, the loop counter) -/
+@[simp]
+theorem loop_counter_decorator_iterate_fst_val (δ: Int) (f : Int → α → α) (i₀ : Int) (v₀ : α) (k : ℕ) :
+  ((loop_counter_decorator δ f)^[k] (i₀, v₀)).1 = i₀ + k * δ := by
+    induction k generalizing i₀ v₀
+    case zero => simp
+    case succ i hi =>
+      simp
+      rw[hi]
```
tobiasgrosser, line 67: Whitespace before `[`. Let's use Mathlib style.

tobiasgrosser (approved):

tobiasgrosser:
Let's see if @alexkeizer has further comments.

SSA/Projects/Scf/Scf.lean
bollu: Fixed by latest commit.

bollu:
Ping @alexkeizer for comments!

SSA/Projects/Scf/Scf.lean
```diff
@@ -223,16 +291,68 @@ abbrev RegionRet [Goedel Ty] (t : Ty) {Γ : Ctxt Ty} (v : Var Γ t) : Com Op Γ
...
+  (hvar : Ctxt.get? (Ctxt.snoc Γ ty) 2 = some var_val) :
+  (V.snoc v) ⟨2, hvar⟩ = V ⟨1, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ := rfl
+
+/-- (ctx.snoc v₁) ⟨3, _⟩ = ctx ⟨2, _⟩ -/
+@[simp]
+theorem Ctxt.Valuation.snoc_eval_three {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
+  (hvar : Ctxt.get? (Ctxt.snoc Γ ty) 3 = some var_val) :
+  (V.snoc v) ⟨3, hvar⟩ = V ⟨2, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ := rfl
+
+/-- (ctx.snoc v₁) ⟨4, _⟩ = ctx ⟨3, _⟩ -/
```
alexkeizer, line 325: It feels weird that these statements are hardcoded for speficic constants, can we make this a general theorem about `(V.snoc v) ⟨n+1, hvar⟩ = V ⟨n, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩`?

SSA/Projects/Scf/Scf.lean
```diff
@@ -223,16 +291,68 @@ abbrev RegionRet [Goedel Ty] (t : Ty) {Γ : Ctxt Ty} (v : Var Γ t) : Com Op Γ
...
+@[simp]
+theorem Ctxt.Valuation.snoc_eval_six {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
+  (hvar : Ctxt.get? (Ctxt.snoc Γ ty) 6 = some var_val) :
+  (V.snoc v) ⟨6, hvar⟩ = V ⟨5, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ := rfl
+
+/-- ((ctx.snoc v₁).snoc v₂) ⟨2, _⟩ = ctx ⟨0, _⟩ -/
+@[simp]
+theorem Ctxt.Valuation.snoc_snoc_eval_two {ty₁ ty₂: Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v₁ : ⟦ty₁⟧) (v₂ : ⟦ty₂⟧)
+  (hvar : Ctxt.get? ((Ctxt.snoc Γ ty₁).snoc ty₂) 2 = some var_val) :
+  ((V.snoc v₁).snoc v₂) ⟨2, hvar⟩ = V ⟨0, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ := rfl
```
alexkeizer, line 347: Is this not already provable by simp by the preceding simp-lemmas?

SSA/Projects/Scf/Scf.lean
```diff
@@ -401,25 +508,46 @@ def lhs : Com Op [/- v0 -/ t] t :=
...
+      (start1, Γv (Var.last [] t)) = (loop_counter_decorator 1 fun i v =>
+          Com.denote rgn (Ctxt.Valuation.snoc (Ctxt.Valuation.snoc default v) i))^[niters2 + niters1]
+      (start1, Γv (Var.last [] t)) := by
+        congr
+    simp [H]
+    simp [Function.iterate_add_apply]
+    congr
+    rw [loop_counter_decorator_iterate_fst_val]
+    linarith
+    done
```
alexkeizer, line 550: This proof feels like it could use some golfing/simplifying

alexkeizer (changes requested):
Mostly nitpicks about naming and organization

SSA/Projects/Scf/Scf.lean
```diff
@@ -56,27 +56,95 @@ instance : OpSignature Op Ty where
...
+/-- Evaluating a loop index invariant function is the same as evaluating it at 0 -/
+theorem LoopIndexInvariant_eval (f : Int → t → t) (hf : LoopIndexInvariant f) (i : Int) (v : t) :
+  f i v = f 0 v := by unfold LoopIndexInvariant at hf; rw [hf]
+
+/-- Loop invariant functions can be simulated by a simpler function -/
+def LoopInvariantVersion (f : Int → t → t) : t → t := fun v => f 0 v
+
+/-- If there exists a function `g : t → t` which agrees with `f`, then `f` is loop index invariant. -/
+theorem LoopIndexInvariant_eq_other_function_LoopIndexInvariant
+    (f : Int → t → t) (g : t → t) (hf : ∀ (i : Int) (v : t), f i v = g v) :
```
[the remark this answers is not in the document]
bollu: Because the intuition for "loop invariant" is that any two iterations are equal. We prove the equivalence with this other definition later.

SSA/Projects/Scf/Scf.lean
```diff
@@ -223,16 +291,68 @@ abbrev RegionRet [Goedel Ty] (t : Ty) {Γ : Ctxt Ty} (v : Var Γ t) : Com Op Γ
...
 set_option pp.proofs.withType false
 
+/-# Helpers for simplifying context manipulation with toSnoc and variable access -/
+section ValuationVariableAccess
+
+/-- (ctx.snoc v₁) ⟨1, _⟩ = ctx ⟨0, _⟩ -/
+@[simp]
+theorem Ctxt.Valuation.snoc_eval_one {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
+  (hvar : Ctxt.get? (Ctxt.snoc Γ ty) 1 = some var_val) :
+  (V.snoc v) ⟨1, hvar⟩ = V ⟨0, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ := rfl
```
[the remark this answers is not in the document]
bollu: Resolved, these are all consolidated into a single, properly formatted `simp` lemma.

bollu:
@alexkeizer Refactored the implementation to be correctly namespaced, killed a bunch of dead / redundant code. The open issue that I can see is the golfing of the large `simp` set --- this ought to be a `simp_peephole`, but I'm not actually able to get it to `simp` correctly. Do you have some cycles to try?

SSA/Projects/Scf/Scf.lean
bollu: @alexkeizer Proof has been correctly golfed now!

bollu: Golfed!

bollu:
@alexkeizer I've managed to cut down the proof size by adding to the `simp` set of `simp_peephole`, and working through the ramifications :) This is the cleanest I was able to get the largest proof in the file: that of `for` loop fusion: https://github.com/opencompl/ssa/pull/141/files#diff-38348534503242b1ccef430c9c2aadc0bd40ada759c54a383516d206465856cdR471

Do you feel this is good to merge?

SSA/Projects/Scf/Scf.lean
```diff
@@ -51,52 +57,138 @@ instance : OpSignature Op Ty where
...
+-/
+abbrev LoopBody (t : Type) : Type := Int → t → t
+
+namespace LoopBody
+
 /-- Convert a function `f` which is a single loop iteration into a function
   that iterates and updates the loop counter. -/
-def loop_counter_decorator (δ: Int) (f : Int → α → α) : Int × α → Int × α :=
+def CounterDecorator (δ: Int) (f : LoopBody α) : Int × α → Int × α :=
   fun (i, v) => (i + δ, f i v)
```
alexkeizer, line 80: This is should be `lowerCamelCase`, since it's a function returning `Int \times \a : Type _`

SSA/Core/ErasedContext.lean
```diff
@@ -250,6 +250,16 @@ theorem Valuation.snoc_toSnoc {Γ : Ctxt Ty} {t t' : Ty} (s : Γ.Valuation) (x :
...
 
+/-!
+# Helper to simplify context manipulation with toSnoc and variable access.
+-/
+/--  (ctx.snoc v₁) ⟨n+1, _⟩ = ctx ⟨n, _⟩ -/
+@[simp]
+theorem Valuation.snoc_eval {ty : Ty} (Γ : Ctxt Ty) (V : Γ.Valuation) (v : ⟦ty⟧)
+    (hvar : Ctxt.get? (Ctxt.snoc Γ ty) (n+1) = some var_val) :
+    (V.snoc v) ⟨n+1, hvar⟩ = V ⟨n, by simp [Ctxt.get?,Ctxt.snoc] at hvar; exact hvar⟩ :=
+  rfl
```
alexkeizer, line 261: I think this should be called `Valuation.snoc_succ` or something similar.

Somewhat related: we should pick a normal form between `V (Var.last ..)`/`V (Var.toSnoc ..)` and `V ⟨_, _⟩`. I would prefer the former, making this lemma unnecessary (with `snoc_toSnoc` already existing)

SSA/Projects/Scf/Scf.lean
```diff
@@ -51,52 +57,138 @@ instance : OpSignature Op Ty where
...
+theorem eval (f : LoopBody t) (hf : LoopBody.IndexInvariant f) (i : Int) (v : t) :
+  f i v = f 0 v := by unfold LoopBody.IndexInvariant at hf; rw [hf]
+
+/-- Loop invariant functions can be simulated by a simpler function -/
+def atZero (f : LoopBody t) : t → t := fun v => f 0 v
+
+/-- If there exists a function `g : t → t` which agrees with `f`, then `f` is loop index invariant. -/
+theorem eq_invariant_fn
+    (f : LoopBody t) (g : t → t) (hf : ∀ (i : Int) (v : t), f i v = g v) :
+    LoopBody.IndexInvariant f /\ atZero f = g:= by
```
alexkeizer, line 97: Unicode `\land` instead of `/\`

SSA/Projects/Scf/Scf.lean
```diff
@@ -51,52 +57,138 @@ instance : OpSignature Op Ty where
...
+        rw [hf 0 v]
+
+end LoopBody
+
+namespace LoopBody.IndexInvariant
 
+/-- Evaluating a loop index invariant function is the same as evaluating it at Invariant -/
+@[simp]
+theorem eval' {f : LoopBody t} (hf : LoopBody.IndexInvariant f) (i : Int) (v : t) :
+  f i v = f.atZero v := by unfold LoopBody.IndexInvariant at hf; rw [LoopBody.atZero, hf]
```
alexkeizer, line 114:
```suggestion
theorem eval' {f : LoopBody t} (hf : LoopBody.IndexInvariant f) (i : Int) (v : t) :
    f i v = f.atZero v := by 
  unfold LoopBody.IndexInvariant at hf; rw [LoopBody.atZero, hf]
```
A bit easier to read with the extra newline IMO

SSA/Projects/Scf/Scf.lean
```diff
@@ -51,52 +57,138 @@ instance : OpSignature Op Ty where
...
+theorem iterate {f : LoopBody t} (hf : LoopBody.IndexInvariant f) (δ : Int)  (i₀ : Int) (v₀ : t) (k : ℕ) :
+  (f.CounterDecorator δ)^[k] (i₀, v₀) = (i₀ + δ * k, (f.atZero)^[k] v₀) := by
+    induction k generalizing i₀ v₀
+    case zero => simp
+    case succ i hi =>
+      simp
+      rw [hi]
+      simp [LoopBody.CounterDecorator]
+      simp [eval' hf]
+      linarith
```
alexkeizer, line 127: We indent the type with 4 spaces, and the body with 2 (at least, this is the Mathlib style). 

You should check this across the whole file, there's more defs that break this convention.

SSA/Projects/Scf/Scf.lean
```diff
@@ -51,52 +57,138 @@ instance : OpSignature Op Ty where
...
+      case right =>
+        simp [atZero]
+        funext v
+        rw [hf 0 v]
+
+end LoopBody
+
+namespace LoopBody.IndexInvariant
 
+/-- Evaluating a loop index invariant function is the same as evaluating it at Invariant -/
```
alexkeizer, line 111:
```suggestion
/-- Evaluating a loop index invariant function is the same as evaluating it at `atZero` -/
```
We renamed `Invariant`

alexkeizer (commented):
I've mostly got some nitpicks around style and naming conventions. 
Since I'll be unavailable next week, and none of it is a big blocker, feel free to ignore any comments you disagree with.

SSA/Projects/Scf/Scf.lean
alexkeizer: Ah, sorry, I missed your original message. In general, I think we ought to have a closer look at `simp_peephole`. The current implementation was nice as a proof of concept, but it's way too brittle.

SSA/Projects/Scf/Scf.lean
bollu: Done.

SSA/Core/ErasedContext.lean
bollu: Renamed to `snoc_succ`. I am not sure how to setup the `simp` normal form correctly, I tried adding more lemmas, but was unable to get it to normalize as I wanted to. Is there a tool to debug simp normal forms?

SSA/Projects/Scf/Scf.lean
bollu: Done.

SSA/Projects/Scf/Scf.lean
bollu: Fixed across the file.

SSA/Core/ErasedContext.lean
alexkeizer: None that I know of. Usually, when a `simp` lemma I expect to apply isn't working, I'll try using it explicitly with `rw` and try to debug from there

alexkeizer: We can defer this to an issue, I'm merging for now

PR #142, opened 2023-11-15
Scf for reverse

bollu:
Stack-patch, stacked on top of #141 

I don't trust that this goes through, because while I have the appropriate `@[simp]` lemmas about functions that are loop invariant, I find it utterly dubious that Lean managed to put these together into a proof.

I'm going to hold back and come back to this with fresh eyes tomorrow.


bollu:
This already exists, and was merged with #141 . Closing this.

PR #144, opened 2023-11-16
[Does not work right now due to Lean bug] feat: expr pattern match to ignore ty_eq field

bollu:


tobiasgrosser:
What was the intention of this patch? Why do we try to ignore `ty_eq`?

bollu:
@tobiasgrosser It makes some of our code look cleaner, since then we don't need to match for `ty_eq` to be `rfl`. I'm closing this PR since I don't see this being resolved anytime soon.

PR #146, opened 2024-01-08
feat: add derived contexts

alexkeizer:
Factored out from #128.

A derived context is a context bundled with a context difference w.r.t. a static base context.
This makes it easy to define coercions from variables of the base context into the derived context.


bollu:
LGTM!

PR #147, opened 2024-01-08
feat: Separate lib for Alive examples

alexkeizer:
Compile alive examples in a separate library to limit the impact on build times


bollu:
I am afraid of burning too many CI credits. @tobiasgrosser , how many credits do we have?

tobiasgrosser:
We don't seem to oush super often, no? Let's try this until it becomes a problem. Then we just setup a custom builder at Cambridge

alexkeizer:
Note that I haven't removed the early exit from the AliveAutogenerated file, so this PR doesn't actually change the CI runtime that much

bollu:
@alexkeizer Then I am confused, does it actually take a long time to compile? (I see that it takes 30s, but this is a one-time cost, no?)

alexkeizer:
Well eventually we should probably have the whole file checked by CI at least semi-regularly, so this is a step in that direction. I didn't remove the early exit yet out of concern for the CI usage, but if that is not expected to be a problem let's remove it

bollu:
@alexkeizer OK, that rationale makes sense! I am pulling it in.

PR #158, opened 2024-01-11
Edsl cleanup refactor

alexkeizer:
A refactor to clean up some of the EDSL code that got merged in #128.

By adding a bit more structure to the `MOp` type, we can greatly reduce the LOC we need in EDSL, by consolidating a bunch of carbon-copy match arms into just one.
Also, I've removed `mkOpExpr` since it did not seem to be used anywhere.


bollu:
Lgtm, i really like the match patterns :)

PR #159, opened 2024-01-14
feat: recursion principles for contexts and valuations

alexkeizer:
Attempt at fixing some of the alive examples that are currently not working.
Mostly dumping this here in case anyone wants to have a look next week.

The most usable parts so far are the extra recursion principles for `Ctxt` and `Ctxt.Valuation`: I hope to replace the current `generalize ...` trick with a bit of meta code that analyzes the type of a `Ctxt.Valuation`, recursively applies the case distinction until the remaining context is empty, and then reduces the `(Ctxt.Valuation.snoc _ v) (Var.last ..)` to just `v` --- we have a simp-lemma that should do this rewrite, but it somehow fails to unify with the lhs correctly, this might be a job for `simproc`.

### DIFF
--- a/SSA/Core/ErasedContext.lean
+++ b/SSA/Core/ErasedContext.lean
@@ -60,6 +60,15 @@ instance : LawfulFunctor Ctxt where
   id_map    := by simp only [List.map_eq_map, List.map_id, forall_const]
   map_const := by simp only [Functor.mapConst, Functor.map, Function.const, forall_const]
 
+/-- Recursion principle for contexts in terms of `Ctxt.nil` and `Ctxt.snoc` -/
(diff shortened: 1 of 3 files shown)


bollu:
@alexkeizer Is this superceded by #166 ?

alexkeizer:
Somewhat, in the sekse that that tactic seems to fix the specific issue we were seeing.

Still, this PR aims to replace the generalize trick/hack with a more intelligent tactic that normalizes the valuation in terms of `toSnoc`, which I think is still something worth exploring.

alexkeizer:
I've minimized this up to be just the new recursion principles, @bollu if you could have a look at this and hit merge that would be appreciated!

PR #166, opened 2024-01-23
feat: `change_mlir_context` tactic

alexkeizer:
Add a new tactic that fully reduces the context in the type of a valuation. This replaces the brittle `simp at \Gv` that we were previously using to simplify the type of valuations with a more robust alternative.

I haven't regenerated the Alive examples yet, but a quick manual test suggests this PR should fix most of them, and hopefully simplify some of the new examples @bollu is working on.


bollu:
I'm going to merge this to try and see if the new tactic works better.

PR #160, opened 2024-01-15
wip: Formalize muldivrem, select, andorxor [stale]

bollu:
Closes #152, #153, #154

PR #167, opened 2024-01-25
wip: Formalize muldivrem, select, andorxor 

bollu:
Closes https://github.com/opencompl/ssa/issues/152, https://github.com/opencompl/ssa/issues/153, https://github.com/opencompl/ssa/issues/154, supercedes #160

### DIFF
--- a/SSA/Projects/InstCombine/ForLean.lean
+++ b/SSA/Projects/InstCombine/ForLean.lean
@@ -572,6 +572,210 @@ theorem intMin_not_gt_zero : ¬ (intMin w >ₛ (0#w)):= by
 theorem zero_sub_eq_neg {w : Nat} { A : BitVec w}: BitVec.ofInt w 0 - A = -A:= by
   simp [BitVec.ofInt_zero_eq]
 
+-- Any bitvec of width 0 is equal to the zero bitvector
+theorem width_zero_eq_zero (x : BitVec 0) : x = BitVec.ofNat 0 0 :=
+  Subsingleton.allEq ..
+
+@[simp]
+theorem toInt_width_zero (x : BitVec 0) : BitVec.toInt x = 0 := by
+  rw [BitVec.width_zero_eq_zero x]
+  simp
+
+@[simp]
+theorem toInt_ofInt_width_one_one : BitVec.toInt (BitVec.ofInt 1 1) = -1 := rfl
+
+@[simp]
+theorem toInt_ofInt_zero : BitVec.toInt (BitVec.ofInt w 0) = 0 := by
+  simp [BitVec.msb, BitVec.getMsb, BitVec.getLsb, BitVec.ofInt, BitVec.toInt]
+
+@[simp]
+theorem toInt_ofInt_1_width_zero :
+    BitVec.toInt (BitVec.ofInt (n := 0) 1) = 0 := rfl
+
+@[simp]
+theorem toInt_ofInt_1_width_one :
+    BitVec.toInt (BitVec.ofInt (n := 1) 1) = -1 := rfl
+
+-- if w = 0. then value is 0
+-- if w = 1, then value is -1.
+
+@[simp]
+theorem toNat_ofInt_one_width_zero : BitVec.toNat (BitVec.ofInt (n := 0) 1) = 0 := rfl
+
+@[simp]
+theorem toNat_ofInt_one_width_one : BitVec.toNat (BitVec.ofInt (n := 1) 1) = 1 := rfl
+
+@[simp]
+theorem ofNat_toNat_zero :
+BitVec.toNat (BitVec.ofInt w 0) = 0 := by
+  simp only [BitVec.toNat, BitVec.ofInt, BitVec.toFin, BitVec.ofNat, OfNat.ofNat]
+  norm_cast
+
+/-- The usual theorem is stated with nat as the index. -/
+@[simp] lemma getLsb_xor' (x y : BitVec w) (i : Nat) :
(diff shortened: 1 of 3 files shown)


SSA/Projects/InstCombine/AliveHandwrittenExamples.lean
```diff
@@ -79,4 +257,439 @@ theorem alive_DivRemOfSelect (w : Nat) :
...
+  rcases c3 with none | c3 <;>
+  rcases x with none | x <;> simp[Option.bind, Bind.bind]
+  . rcases (LLVM.lshr? (x ^^^ c1) c2) with none | irrelevant <;> simp
+  . -- TODO: develop automation to automatically do the case split. See also: prime motivation for simproc.
+    have hc2: (c2.toNat < w) ∨ ¬ (c2.toNat < w) := by tauto
+    rcases hc2 with lt | gt
+    . simp only [ge_iff_le, LLVM.lshr?_eq_some lt, BitVec.ushiftRight_eq,
+      BitVec.Refinement.some_some]
+      -- (x ^^^ c1) >>> BitVec.toNat c2 ^^^ c3 = c3 >>> BitVec.toNat c2 ^^^ (c1 >>> BitVec.toNat c2 ^^^ c3)
+      sorry -- I need bitwise reasoning here!
```
bollu, line 581: @alexkeizer I was wondering how I finish this off with bitwise reasoning! I can't seem to recall the tactic incantation to do this. IIRC, we had tried the bitwise version once upon a time, but I am not sure how to use it.

bollu:
I proved the bitshifting theorem, modulo lemmas. Speaking to @alexkeizer about this, work on proving the lemma would be redundant, since we are fighting with definitions that have been improved upstream. However, we cannot directly use upstream since we want both `ext` as well as `ring`. We do not have `ring` working yet with the new `Std` definitions.

@alexkeizer is checking if we can get the `ring` branch working with the latest Std. then we can just use the newest definitions for everything and be happy.

tobiasgrosser:
I updated this to the latest main. Not all proofs go through, but I think its useful to keep this up-to-date.

bollu:
We decided to clean this up after #225 , since #225 ensures that our meta code elaboration is stable.

tobiasgrosser:
This builds for me and is sorry-free. However, the CI seems to disagree. 😢 

The code is still a mess, but at least this goes through again. I really need to learn how to work with Int vs Nat conversions.

tobiasgrosser:
Read-to-review is probably a stretch. This PR is very ugly and a good base to pull things out for upstream.

tobiasgrosser:
Updated to `main` to remove the com wrapper diff. Now this PR is just about our lean proofs.

SSA/Projects/InstCombine/AliveHandwrittenLargeExamples.lean
```diff
@@ -1236,24 +1236,12 @@ def alive_simplifyMulDivRem290 (w : Nat) :
...
-  by_cases h : w ≤ BitVec.toNat B
-  simp only [h]
-  simp
-  simp only [h]
-  simp
-  simp at h
-  simp [trsa h]
+  intros A B
+  rcases A with none | A  <;> simp [Option.bind, Bind.bind] <;>
+  rcases B with none | B  <;> simp [Option.bind, Bind.bind] <;>
```
bollu, line 1241: The pattern 

```lean
  rcases A with none | A  <;> simp [Option.bind, Bind.bind] <;>
```

is AFAIU a little dangerous, since it also runs full `simp` on the `.some` case, without actually closing the goal. I would suggest

```lean
  rcases A with none | A  <;> try (simp [Option.bind, Bind.bind]; done) <;>
```

This will perform the following:
- In the case of `none`, since `simp [Option.bind, Bind.bind]` closes the goal, `done` succeeds, and thus the `try(...)` keeps the goal closed.
- In the case of `some _`. `simp [Option.bind, Bind.bind]` **does not** close the goal, `done` fails, and thus `try` undoes the simplification from an unguarded `simp`, reverting to the original proof state.

tobiasgrosser:
The actual LLVM-proofs are meanwhile all moved over. We leave this PR open for a moment to see if we want to extract some of the BitVec theorems.

SSA/Projects/InstCombine/ForLean.lean
```diff
@@ -572,6 +572,246 @@ theorem intMin_not_gt_zero : ¬ (intMin w >ₛ (0#w)):= by
...
+
+lemma toInt_pos_iff_small (w : Nat) (x : BitVec w) : x.toNat < (2 : Nat) ^ (w - 1) ↔ BitVec.toInt x ≥ 0 := by
+  constructor
+  apply BitVec.toInt_pos_of_small
+  apply BitVec.small_of_toInt_pos
+
+lemma toInt_pos_iff_small' (w : Nat) (x : BitVec (w + 1)) : x.toNat < (2 : Nat) ^ w ↔ BitVec.toInt x ≥ 0 := by
+  apply BitVec.toInt_pos_iff_small
+
+lemma toInt_zero_iff (w : Nat) (x : BitVec w) : BitVec.toInt x = 0 ↔ x = 0 := by
```
bollu, line 773: This should follow from something like `toInt_inj` and `toInt_zero`.

tobiasgrosser: It follow from `toInt_eq`.

bollu:
Thanks a lot for pushing this @tobiasgrosser !

PR #168, opened 2024-01-26
refactor: make variable introduction in `simp_peephole` more robust

alexkeizer:
This replaces the `generalize` trick with a more principled approach in terms of the recursor on valuations.

Credits to @bollu for the cute `repeat (conv ...)` trick ;) !


tobiasgrosser:
@alexkeizer and @bollu, this branch should have a significantly smaller diff than before as I already merged various of these changes into `main`. Maybe we can rebase this to `main` and then discuss together how we can make this codebase fundamentally more robust.

tobiasgrosser:
This is now worked on in #224.

PR #172, opened 2024-01-29
feat: fix bitwidth to a single width across args and return value

bollu:
This makes the LLVM semantics agree precisely with our definitions.


alexkeizer (approved):
`some` is exported in the root namespace, so no dot-notation necessary!
Otherwise, LGTM, feel free to merge

PR #175, opened 2024-02-05
Check LLVM semantics end-to-end

goens:
This checks the semantics including the entire parsing and elaboration pipeline


goens:
CI seems broken by some mathlib stuff that's independent of this (will wait until https://github.com/leanprover-community/mathlib4/pull/10455 is merged and then bump mathlib in a separate PR)

There's a `sorry` in there that I don't know how to fix, it's related to [this thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/unification.20under.20pattern.20matches.20inside.20.60structure.60s/near/420985872).

SSA/Core/Util.lean
```diff
@@ -109,3 +109,45 @@ instance [inst : Cli.ParseableType τ] {n : ℕ} : Cli.ParseableType (Vector τ
       return h ▸ Vector.ofArray arr
     else
       none
+
+
+def Std.BitVec.width : Std.BitVec w → Nat := fun _ => w
```
alexkeizer, line 114:
```suggestion
@[simp] abbrev Std.BitVec.width : Std.BitVec w → Nat := fun _ => w
```
I expect we don't really care about having the function in goals, so having it automatically `simp` away is nice

SSA/Core/Util.lean
```diff
@@ -109,3 +109,45 @@ instance [inst : Cli.ParseableType τ] {n : ℕ} : Cli.ParseableType (Vector τ
...
+
+-- From Haskell:
+def List.toTiers : List α  → List (List α) := (List.groupBy (fun _ _ => false))
```
alexkeizer, line 117: I don't know the haskell equivalent, could you add a doc-string that explains what this does?

SSA/Core/Util.lean
```diff
@@ -109,3 +109,45 @@ instance [inst : Cli.ParseableType τ] {n : ℕ} : Cli.ParseableType (Vector τ
...
+
+-- fugly; sorry, Lean doesn't have that neat list comprehension
+/-- In Haskell this was defined inline as `**` -/
+def prodWith : (α → β → γ) → List α → List β → List γ
+  | f, as, bs => Id.run do
+      let mut res := []
+      for a in as do
+        for b in bs do
+         res := res ++ [ f a b]
+      res
```
alexkeizer, line 134: Isn't this just `List.zip as bs |>.map f`?

SSA/Core/Util.lean
```diff
@@ -109,3 +109,45 @@ instance [inst : Cli.ParseableType τ] {n : ℕ} : Cli.ParseableType (Vector τ
...
+
+def productWith : (α -> β -> γ) → List (List α) → List (List β) → List (List γ)
+  | _, _, []  =>  []
+  | _, [], _  =>  []
+  | f, (xs::xss), yss =>
+    appendTiers (List.map (prodWith f xs ·) yss) (productWith f xss yss)
```
alexkeizer, line 140: Isn't this the same as `List.prodwith (List.prodWith) xss yss`

SSA/Projects/InstCombine/LLVM/CLITests.lean
```diff
@@ -187,34 +187,45 @@ def CliTest.eval (test : CliTest) (values : Vector ℤ test.context.length) (hMv
...
   | ty::tys, ⟨val::vals,hlen⟩ =>
-    let valsVec : Vector ℤ tys.length := ⟨vals,by aesop⟩
+    let valsVec : Vector (Option ℤ) tys.length := ⟨vals,by aesop⟩
     let valuation' := mkValuation tys valsVec
-    let newTy : toType ty := ↑(val)
-    Ctxt.Valuation.snoc valuation' newTy
+    match ty with
+      | .bitvec (.concrete w) =>
+         let newTy : toType (.bitvec (.concrete w)) := Option.map (Std.BitVec.ofInt w) val
+         Ctxt.Valuation.snoc valuation' newTy
```
alexkeizer, line 199: We should be consistent about using `Int` or `ℤ`. At least within one function, but it might make sense to agree on a preference for the whole project. I don't have strong opinions about one or the other.

SSA/Projects/InstCombine/LLVM/CLITests.lean
```diff
@@ -187,34 +187,45 @@ def CliTest.eval (test : CliTest) (values : Vector ℤ test.context.length) (hMv
...
   return test.code.denote valuation
 
+def ConcreteCliTest.eval? (test : ConcreteCliTest) (values : Array (Option ℤ)) :
+ IO (Except String ⟦test.ty⟧) := do
+   if h : values.size = test.context.length then
+     let valuesVec : Vector (Option ℤ) test.context.length := h ▸ (Vector.ofArray values)
+     return Except.ok <| (← test.eval valuesVec)
+   else
+     return Except.error s!"Invalid input length: {values} has length {values.size}, required {test.context.length}"
+
```
alexkeizer, line 214: The indentation is off: this function uses only one space per indent, rather than two

SSA/Projects/InstCombine/LLVM/Enumerator.lean
```diff
@@ -115,40 +127,84 @@ def icmpRows (pred : LLVM.IntPredicate) : Array Row := Id.run do
...
+    | _ => "Error"
+
+/-- Produce rows for a `ConcreteCliTest` -/
+def concreteCliTestRows (test : ConcreteCliTest) : IO <| Array Row := do
+  let mut rows := #[]
+  let argTys : List test.code.getTy := test.context.reverse -- reverse because context is stack
+  let mut args : Array (Array (Option ℤ)) := #[]
+  for ty in argTys do
+    match ty with
+      | .bitvec (.concrete w) => args := args.push <| (BitVec.inputsForWidth w).map (Option.map BitVec.toInt)
```
alexkeizer, line 144: Could we write this as `(List.map ...).toArray`?
That seems a bit cleaner than going full-imperative, although I wonder if there is a performance difference

SSA/Core/Util.lean
bollu: @alexkeizer no it is not, note that `zip` will have `n` elements, while this has `n^2` elements.

SSA/Core/Util.lean
bollu:
```suggestion
-- Takes a list of values xs and transform it into tiers on which each tier is occupied by a single element from xs.
def List.toTiers : List α  → List (List α) := (List.groupBy (fun _ _ => false))
example : [1, 2, 3].toTiers = [[1], [2], [3]] := rfl
```

SSA/Core/Util.lean
alexkeizer: Ahh, now I see: it's taking the cartesian product, rather than zipping the two lists, right

bollu:
```suggestion
def prodWith (f : α → β → γ) (as : List α) (bs : List β) : List γ :=
  as.bind fun a => bs.map fun b => f a b

example : prodWith (String.append) ["a", "b"] ["c", "d"] =
  ["ac", "ad", "bc", "bd"] := by decide
```

PR #187, opened 2024-02-28
Remove unintentional commits to main

tobiasgrosser:
I am not sure why these commits ended up in the main, but they seem to break the build and also should not be committed to an auto-generated file.


tobiasgrosser:
Push this to main, to ensure that my other PRs can be compile tests.

PR #189, opened 2024-02-28
Update to latest mathlib

tobiasgrosser:
- This updates to lean 4.6.0
- BitVec is still in std4 (to be in Lean4 from 4.7.0)
- Import the 'CommRing' instance into a file ForMathlib such that we can track Mathlib master

We have currently 47 out of 107 statements proven correct.


tobiasgrosser:
This is ready to review. However, I have the following open questions:

> Why is BitVec.negOne_eq_allOnes not a standard simp lemma? Or rather, why do we rewrite -1 to allOnes. That seems to be an unnecessary complication, no?

SSA/Projects/FullyHomomorphicEncryption/Basic.lean
```diff
@@ -34,19 +34,6 @@ variable (q t : Nat) [ hqgt1 : Fact (q > 1)] (n : Nat)
 -- see: https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Way.20to.20recover.20computability.3F/near/322382109
 -- and :https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Groebner.20bases
 
--- Question: Can we make something like d := 2^n work as a macro?
```
goens, line 37: Why is this comment removed? This is still an interesting/valid question

SSA/Projects/InstCombine/AliveStatements.lean
```diff
@@ -9,8 +10,11 @@ open Std.BitVec
 open LLVM
 
 @[simp] lemma getLsb_negOne' (w : ℕ) (i : Fin w) :
-    getLsb (-1#w) ↑i := by 
-  simp only [getLsb_val_eq_getLsb', getLsb'_neg_ofNat_one]
+    getLsb (-1#w) i := by
```
goens, line 13: I'm confused from seeing changes in this file. This is an auto-generated file, I thought we were not going to manually change this file anymore.

SSA/Projects/InstCombine/ForMathlib.lean
```diff
@@ -0,0 +1,30 @@
...
+  case succ w =>
+    simp only [Int.cast, IntCast.intCast, BitVec.ofInt]
+    unfold Int.castDef
+    cases' z with z z
+    · rfl
+    · simp only [cast_add, cast_one, neg_add_rev]
+      rw [← add_ofFin, ofFin_neg, ofFin_ofNat, ofNat_eq_ofNat, ofFin_neg, ofFin_natCast,
+        natCast_eq, negOne_eq_allOnes, ← sub_toAdd, allOnes_sub_eq_not]
+
+theorem toFin_intCast (z : ℤ) : toFin (z : BitVec w) = z := by
```
goens, line 22: would something like `@[norm_cast]` make sense here?

goens (changes requested):

SSA/Projects/FullyHomomorphicEncryption/Basic.lean
tobiasgrosser: Right. I will add it back.

SSA/Projects/InstCombine/AliveStatements.lean
tobiasgrosser: @goens, what is the process to get this file updated properly? Happy to follow your suggestions here.

SSA/Projects/InstCombine/ForMathlib.lean
tobiasgrosser: I have no idea. How would that work? Why would that be preferred?

SSA/Projects/InstCombine/AliveStatements.lean
goens: that's a good question, I don't know if we've refined this properly, we talked about it before the rebuttal, maybe @bollu remembers what the final process was? Ideally, these changes would go into the `alive_auto` tactic or the `extract_alive_statements.sh` script.

SSA/Projects/InstCombine/ForMathlib.lean
goens: this is a very minor comment/question, mostly interesting later for upstreaming. `norm_cast` is a tactic that handles all of these coercions (see https://lean-forward.github.io/norm_cast/norm_cast.pdf) and the tag is like the `@[simp]` tag to tell the tactic which lemmas to use for these coercions.

tobiasgrosser: To be honest, I just copied this from alex Mathlib pr without thinking too much about it. I guess we can refine this in the upstreaming process.

SSA/Projects/InstCombine/AliveStatements.lean
tobiasgrosser: Right. Maybe document the processes here among the two of you. I can then follow this process to regenerate this file.

SSA/Projects/FullyHomomorphicEncryption/Basic.lean
tobiasgrosser: https://github.com/opencompl/ssa/pull/192

PR #199, opened 2024-03-22
fix: replace generalize! by vanilla generalize by unfolding `Var.toSnoc` and `Var.last` first

alexkeizer:


tobiasgrosser:
Nice. Thank you, @alexkeizer.

PR #201, opened 2024-03-26
chore: add context simplification lemmas to simp_[alive_]peephole

tobiasgrosser:


bollu:
LGTM, I pair programmed this with @tobiasgrosser. @goens , the goal was to get the simp_peephole and simp_alive_peephole changes from the FHE branch into master, so we can make the diff smaller (and take away some of the merging complexity)

bollu:
GH merge queues are insane, they seem to simply merge without running appropriate checks!

bollu:
Let's hope that this succeeds.

alexkeizer:
A retro-active review:

I'm a bit confused, these simp lemmas (`Ctxt.destruct_cons` and `_nil`) were intended as a replacement of the `generalize`-trick, and rely on a goal where the valuation is reverted before `simp`-ing. This change seems to cherry pick just the simp lemmas, but not the reverting of the valuation, so I don't think these new lemma are actually used anywhere.

PR #204, opened 2024-03-26
Cleanup proofs

tobiasgrosser:


SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -11,38 +11,38 @@ namespace LLVM
 The ‘and’ instruction returns the bitwise logical and of its two operands.
 -/
 def and? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
- .some <| x &&& y
+ some <| x &&& y
```
goens, line 14: I'm curious about these changes, what's the reason for dropping the `.`s? is this aesthetical or functional?

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -192,10 +192,10 @@ def ashr? {n} (op1 : BitVec n) (op2 : BitVec n) : Option (BitVec n) :=
  If the condition is an i1 and it evaluates to 1, the instruction returns the first value argument; otherwise, it returns the second value argument.
 -/
 def select? {w : Nat} (c? : Option (BitVec 1)) (x? y? : Option (BitVec w)) : Option (BitVec w) :=
-  match c? with
```
goens, line 195: a bit wary about changing the semantics, this should be equivalent, right? if they are, what's the goal of using the harder to read bind instead of the more explicit pattern matching?

goens (commented):
the changes seem fine, but I don't quite understand the point, it seems to be making things (equally or) less readable and has no semantic content

SSA/Projects/InstCombine/LLVM/Semantics.lean
tobiasgrosser: I am not sure. Maybe @alexkeizer knows why this change was implemented?

SSA/Projects/InstCombine/LLVM/Semantics.lean
tobiasgrosser: This seems like a clear regression, indeed. I dropped this for now.

tobiasgrosser:
To give more context. This PR adds syntactic changes to `main` that were initially proposed in  https://github.com/opencompl/ssa/pull/168 and subsequently merged into https://github.com/opencompl/ssa/pull/196.

This PR demonstrates that these changes are purely syntactic or do at least not break the current CI. These changes also seem to be improvements towards the syntax of the framework and so should stand by themselves (even though the value may be rather small).

After these changes are merged, the diff on the other PRs vs main should be reduced by quite a bit, hopefully only leaving semantically relevant changes.

SSA/Projects/InstCombine/LLVM/Semantics.lean
tobiasgrosser: To me it seems to just be a readability improvement.

bollu: Yes, because Option is globally open, we choose to drop the `.`

tobiasgrosser: So I can merge this?

bollu:
LGTM

SSA/Projects/InstCombine/LLVM/Semantics.lean
alexkeizer: It was indeed purely aesthetical

PR #207, opened 2024-03-27
fix: avoid surprising symmetry inverse on refinement lemma

tobiasgrosser:


SSA/Projects/InstCombine/ForStd.lean
```diff
@@ -44,7 +44,7 @@ theorem none_right {x? : Option α} :
     rintro ⟨⟩
 
 theorem some_left {x : α} {y? : Option α} :
-    some x ⊑ y? ↔ y? = some x := by
+    some x ⊑ y? ↔ some x = y? := by
```
bollu, line 47: I am less sure about this change, we conventionally use `variable = expr`.

bollu: I would wait for @alexkeizer .

tobiasgrosser: I find it surprising that this change flips the direction of the statement. This will require updates of our auto-generated proofs when applies as a simp lemma. We can leave it as it is and drop this PR. I just then know that when we `activate` this, we need to be careful.

alexkeizer: The flip might be surprising, but it can be desirable. Say we have `some x \sqsubseteq y?` as an assumption, then simplifying that to `y? = some x` gives us a good hypothesis to rewrite along (eliminating the `y?` variable), while `some x = y?` would be a bad rewrite, losing some information.

tobiasgrosser: OK, got it. Thank you. I will close this then.

PR #209, opened 2024-04-06
chore: introduce simp set for llvm

tobiasgrosser:
This simp set will make our simp_alive_peephole tactic simpler.


bollu:
Lgtm, feel invited to pull.

PR #210, opened 2024-04-06
feat: hide monadic behavior of llvm semantics behind a function boundary.

tobiasgrosser:
We introduce with `LLVM.<op>` a set of formal definitions of the per-operation semantics for the LLVM dialect that must be unfolded with `simp_llvm_option` before the per-operation semantics are exposed. This allows us to introduce a tactic `simp_alive_ssa`, which eliminates only the overhead of SSA and leaves us with explicit statements about the not-yet-unfolded LLVM operation semantics.

### DIFF
--- a/SSA/Core/HVector.lean
+++ b/SSA/Core/HVector.lean
@@ -49,6 +49,24 @@ def get {as} : HVector A as → (i : Fin as.length) → A (as.get i)
   | .cons x  _, ⟨0,   _⟩  => x
   | .cons _ xs, ⟨i+1, h⟩  => get xs ⟨i, Nat.succ_lt_succ_iff.mp h⟩
 
+/-- To be used as an auto-tactic to prove that the index of getN is within bounds -/
+macro "hvector_get_elem_tactic" : tactic =>
+  `(tactic|
+      (
+      simp [List.length]
+      )
+   )
+
+/-- `x.getN i` indexes into a `HVector` with a `Nat`.
+
+Note that we cannot define a proper instance of `GetElem` for `HVector`,
+since `GetElem` requires us to specify a type `elem` such that `xs[i] : elem`,
+but `elem` is *not* allowed to depend on the concrete index `i`.
+Thus, `GetElem` does not properly support heterogeneous container types like `HVector`
+-/
+abbrev getN (x : HVector A as) (i : Nat) (hi : i < as.length := by hvector_get_elem_tactic) :
+    A (as.get ⟨i, hi⟩) :=
+  x.get ⟨i, hi⟩
 
 def ToTupleType (A : α → Type*) : List α → Type _
   | [] => PUnit
--- a/SSA/Projects/InstCombine/Base.lean
+++ b/SSA/Projects/InstCombine/Base.lean
@@ -247,30 +247,28 @@ instance : OpSignature (MOp φ) (MTy φ) where
   signature op := ⟨op.sig, [], op.outTy⟩
 
 @[simp]
-def Op.denote (o : Op) (arg : HVector TyDenote.toType (OpSignature.sig o)) :
+def Op.denote (o : Op) (op : HVector TyDenote.toType (OpSignature.sig o)) :
     (TyDenote.toType <| OpSignature.outTy o) :=
   match o with
   | Op.const _ val => const? val
-  | Op.and _ => pairBind and? arg.toPair
-  | Op.or _ => pairBind or? arg.toPair
-  | Op.xor _ => pairBind xor? arg.toPair
-  | Op.shl _ => pairBind shl? arg.toPair
-  | Op.lshr _ => pairBind lshr? arg.toPair
-  | Op.ashr _ => pairBind ashr? arg.toPair
-  | Op.sub _ => pairBind sub?  arg.toPair
-  | Op.add _ => pairBind add? arg.toPair
-  | Op.mul _ => pairBind mul? arg.toPair
-  | Op.sdiv _ => pairBind sdiv? arg.toPair
-  | Op.udiv _ => pairBind udiv? arg.toPair
-  | Op.urem _ => pairBind urem? arg.toPair
-  | Op.srem _ => pairBind srem? arg.toPair
-  | Op.not _ => Option.map (~~~.) arg.toSingle
-  | Op.copy _ => arg.toSingle
-  | Op.neg _ => Option.map (-.) arg.toSingle
-  | Op.select _ =>
-    let (ocond, otrue, ofalse) := arg.toTriple
-    select? ocond otrue ofalse
-  | Op.icmp c _ => pairBind (icmp? c) arg.toPair
+  | Op.copy _      =>             (op.getN 0)
+  | Op.not _       => LLVM.not    (op.getN 0)
+  | Op.neg _       => LLVM.neg    (op.getN 0)
+  | Op.and _       => LLVM.and    (op.getN 0) (op.getN 1)
+  | Op.or _        => LLVM.or     (op.getN 0) (op.getN 1)
+  | Op.xor _       => LLVM.xor    (op.getN 0) (op.getN 1)
+  | Op.shl _       => LLVM.shl    (op.getN 0) (op.getN 1)
+  | Op.lshr _      => LLVM.lshr   (op.getN 0) (op.getN 1)
+  | Op.ashr _      => LLVM.ashr   (op.getN 0) (op.getN 1)
+  | Op.sub _       => LLVM.sub    (op.getN 0) (op.getN 1)
+  | Op.add _       => LLVM.add    (op.getN 0) (op.getN 1)
+  | Op.mul _       => LLVM.mul    (op.getN 0) (op.getN 1)
+  | Op.sdiv _      => LLVM.sdiv   (op.getN 0) (op.getN 1)
+  | Op.udiv _      => LLVM.udiv   (op.getN 0) (op.getN 1)
+  | Op.urem _      => LLVM.urem   (op.getN 0) (op.getN 1)
+  | Op.srem _      => LLVM.srem   (op.getN 0) (op.getN 1)
+  | Op.icmp c _    => LLVM.icmp c (op.getN 0) (op.getN 1)
+  | Op.select _    => LLVM.select (op.getN 0) (op.getN 1) (op.getN 2)
 
 instance : OpDenote Op Ty := ⟨
   fun o args _ => Op.denote o args
--- a/SSA/Projects/InstCombine/LLVM/Semantics.lean
+++ b/SSA/Projects/InstCombine/LLVM/Semantics.lean
@@ -8,20 +8,36 @@ import SSA.Projects.InstCombine.LLVM.SimpSet
 
 namespace LLVM
 
+
+abbrev IntW w := Option <| BitVec w
+
 /--
 The ‘and’ instruction returns the bitwise logical and of its two operands.
 -/
 @[simp_llvm]
-def and? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
+def and? {w : Nat} (x y : BitVec w) : IntW w :=
   some <| x &&& y
+
+@[simp_llvm_option]
+def and {w : Nat} (x y : IntW w) : IntW w := do
+  let x' ← x
+  let y' ← y
+  and? x' y'
 
 /--
 The ‘or’ instruction returns the bitwise logical inclusive or of its two
 operands.
 -/
 @[simp_llvm]
-def or? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
+def or? {w : Nat} (x y : BitVec w) : IntW w :=
   some <| x ||| y
+
+
+@[simp_llvm_option]
+def or {w : Nat} (x y : IntW w) : IntW w := do
+  let x' ← x
+  let y' ← y
+  or? x' y'
 
 /--
 The ‘xor’ instruction returns the bitwise logical exclusive or of its two
@@ -29,8 +45,14 @@ operands.  The xor is used to implement the “one’s complement” operation, 
 is the “~” operator in C.
 -/
 @[simp_llvm]
-def xor? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
+def xor? {w : Nat} (x y : BitVec w) : IntW w :=
   some <| x ^^^ y
+
+@[simp_llvm_option]
+def xor {w : Nat} (x y : IntW w) : IntW w := do
+  let x' ← x
+  let y' ← y
+  xor? x' y'
 
 /--
 The value produced is the integer sum of the two operands.
@@ -38,8 +60,14 @@ If the sum has unsigned overflow, the result returned is the mathematical result
 Because LLVM integers use a two’s complement representation, this instruction is appropriate for both signed and unsigned integers.
 -/
 @[simp_llvm]
-def add? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
+def add? {w : Nat} (x y : BitVec w) : IntW w :=
   some <| x + y
+
+@[simp_llvm_option]
+def add {w : Nat} (x y : IntW w) : IntW w := do
+  let x' ← x
+  let y' ← y
+  add? x' y'
 
 /--
 The value produced is the integer difference of the two operands.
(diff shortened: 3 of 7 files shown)


SSA/Projects/InstCombine/Base.lean
```diff
@@ -246,19 +246,22 @@ def MOp.outTy : MOp φ → MTy φ
 instance : OpSignature (MOp φ) (MTy φ) where
   signature op := ⟨op.sig, [], op.outTy⟩
 
+
```
goens, line 249: stray whitespace?

SSA/Projects/InstCombine/Base.lean
```diff
@@ -246,19 +246,22 @@ def MOp.outTy : MOp φ → MTy φ
...
 @[simp]
 def Op.denote (o : Op) (arg : HVector TyDenote.toType (OpSignature.sig o)) :
     (TyDenote.toType <| OpSignature.outTy o) :=
   match o with
   | Op.const _ val => const? val
   | Op.and _ => pairBind and? arg.toPair
   | Op.or _ => pairBind or? arg.toPair
-  | Op.xor _ => pairBind xor? arg.toPair
+  | Op.xor _ => let p := arg.toPair
```
goens, line 257: why this change? isn't it better to keep the consistency with the other functions?

SSA/Projects/InstCombine/Base.lean
```diff
@@ -276,4 +279,6 @@ instance : OpDenote Op Ty := ⟨
   fun o args _ => Op.denote o args
 ⟩
 
+#check OpDenote.denote
```
goens, line 282: stray check?

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -4,23 +4,37 @@ import Mathlib.Tactic.Cases
 import Mathlib.Tactic.SplitIfs
 import Mathlib.Tactic.Tauto
 import SSA.Projects.InstCombine.LLVM.SimpSet
+import SSA.Core.Util
 
 
 namespace LLVM
 
+
+def IntN w := Option (BitVec w)
```
goens, line 13: nit: shouldn't it be Bitvec N (or IntW)?

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -4,23 +4,37 @@ import Mathlib.Tactic.Cases
...
+
+def and??? {w: Nat} (x y : IntN w) : IntN w :=
```
goens, line 15: why not add this to the IntN namespace and call it just `and`? (i.e. `IntN.and`)

SSA/Projects/InstCombine/Base.lean
```diff
@@ -62,7 +62,7 @@ def BitVec.width {n : Nat} (_ : BitVec n) : Nat := n
 
 instance : TyDenote Ty where
 toType := fun
-  | .bitvec w => Option <| BitVec w
+  | .bitvec w => Option (BitVec w)
```
goens, line 65: I guess this is a preference thing, but usually avoiding parenthesis scales better for readability (I think that's the Haskell style at least)

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -4,23 +4,37 @@ import Mathlib.Tactic.Cases
...
+
 /--
 The ‘and’ instruction returns the bitwise logical and of its two operands.
 -/
 @[simp_llvm]
-def and? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
+def and? {w : Nat} (x y : BitVec w) : IntN w :=
   some <| x &&& y
 
+def and?? {w : Nat} (x y : IntN w) : IntN w := do
```
goens, line 27: isn't it better to use `pairBind`? this definition style will end up with a lot of verbose repeated code

goens (commented):
I personally like the explicit `Option ` and the use of BitVec, a type defined in core now. It is more transparent about the semantics and it's easier to read familiar types when they're not too nested. This is my personal preference though. I think it's a perfectly valid change, I just prefer the way it is now.

goens:
would be curious what @bollu and @alexkeizer think

SSA/Projects/InstCombine/Base.lean
tobiasgrosser: They are now all consistent.

SSA/Projects/InstCombine/Base.lean
tobiasgrosser: Removed.

SSA/Projects/InstCombine/LLVM/Semantics.lean
tobiasgrosser: This was just test code. It is now removed. The actual new methods are called `LLVM.and`.

SSA/Projects/InstCombine/Base.lean
tobiasgrosser: I dropped this change as it was unrelated.

SSA/Projects/InstCombine/LLVM/Semantics.lean
tobiasgrosser: I updated the tests to use a uniform style. Let me know what you think.

SSA/Projects/InstCombine/LLVM/Semantics.lean
tobiasgrosser: Renamed to IntW.

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -8,48 +8,82 @@ import SSA.Projects.InstCombine.LLVM.SimpSet
...
 @[simp_llvm]
-def and? {w : Nat} (x y : BitVec w) : Option <| BitVec w :=
+def and? {w : Nat} (x y : BitVec w) : IntW w :=
   some <| x &&& y
 
+@[simp_llvm_option]
+def and {w : Nat} (x y : IntW w) : IntW w := do
+  let x' ← x
+  let y' ← y
+  and? x' y'
```
alexkeizer, line 25: Do we ever use `and?` outside of the definition of `and`?
I would prefer just inlining `and?` (idem for `or?`, `xor?`, etc)

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -198,18 +274,24 @@ this instruction returns a poison value.
...
  If the condition is an i1 and it evaluates to 1, the instruction returns the first value argument; otherwise, it returns the second value argument.
 
  If the condition is an i1 and it evaluates to 1, the instruction returns the first value argument; otherwise, it returns the second value argument.
 -/
 @[simp_llvm]
-def select? {w : Nat} (c? : Option (BitVec 1)) (x? y? : Option (BitVec w)) : Option (BitVec w) :=
+def select? {w : Nat} (c? : IntW 1) (x? y? : IntW w ) : IntW w :=
   match c? with
   | none => none
   | some true => x?
```
alexkeizer, line 297: Following the convention that `and` (without `?`) operates on and returns `IntW`, this should become `select`, rather than `select?`

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -317,7 +405,25 @@ the value `(2^w + (i mod 2^w)) mod 2^w`.
 TODO: double-check that truncating works the same as MLIR (signedness, overflow, etc)
 -/
 @[simp_llvm]
-def const? (i : Int): Option (BitVec w) :=
+def const? (i : Int): IntW w :=
   some <| BitVec.ofInt w i
 
+@[simp_llvm]
+def not? {w : Nat} (x : BitVec w) : IntW w := do
+  some <| (~~~.) x
```
alexkeizer, line 413:
```suggestion
  some ~~~x
```

SSA/Core/HVector.lean
```diff
@@ -67,6 +67,9 @@ abbrev toSingle : HVector A [a₁] → A a₁ := toTuple
 abbrev toPair   : HVector A [a₁, a₂] → A a₁ × A a₂ := toTuple
 abbrev toTriple : HVector A [a₁, a₂, a₃] → A a₁ × A a₂ × A a₃ := toTuple
 
+abbrev fst : HVector A [a₁, a₂] → A a₁ := fun x => (toTuple x).fst
+abbrev snd : HVector A [a₁, a₂] → A a₂ := fun x => (toTuple x).snd
+
```
alexkeizer, line 72: If we have a `GetElem` instance then accessing individual elements becomes more consistent, regardless of the size of the vector (at the cost of maybe a little less type-safety, since we would no longer assert that the vector is exactly of size 2 when passing arguments to, e.g., `LLVM.and`, but I would say the consistency is worth it). Having `HVector.fst` hard-coded for a 2-element vector is a bit surprising.

alexkeizer:
> I personally like the explicit `Option ` and the use of BitVec, a type defined in core now. It is more transparent about the semantics and it's easier to read familiar types when they're not too nested. This is my personal preference though. I think it's a perfectly valid change, I just prefer the way it is now.

At first I was inclined to agree having the explicit `Option` is nice, but with UB as a side-effect possibly adding another layer of `Option`, it might actually be better to have a `IntW` be its own inductive (rather than an alias of `Option`), so that we can have:
`.some (.some value)`
`.some .poison`
`.UB`

Instead of the last two being the way less informative `some none` and `none`.

SSA/Core/HVector.lean
tobiasgrosser: I can write this as:

```lean
@[simp]
def Op.denote (o : Op) (arg : HVector TyDenote.toType (OpSignature.sig o)) :
    (TyDenote.toType <| OpSignature.outTy o) :=
  match o with
  | Op.const _ val => const? val
  | Op.and _ => LLVM.and
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.or _ => LLVM.or
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.xor _ => LLVM.xor
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.shl _ => LLVM.shl
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.lshr _ => LLVM.lshr
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.ashr _ => LLVM.ashr
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.sub _ => LLVM.sub
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.add _ => LLVM.add
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.mul _ => LLVM.mul
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.sdiv _ => LLVM.sdiv
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.udiv _ => LLVM.udiv
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.urem _ => LLVM.urem
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.srem _ => LLVM.srem
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
  | Op.not _ => LLVM.not
                        (arg.get ⟨0, by simp [List.length]⟩)
  | Op.copy _ =>
                        (arg.get ⟨0, by simp [List.length]⟩)
  | Op.neg _ => LLVM.neg
                        (arg.get ⟨0, by simp [List.length]⟩)
  | Op.select _ =>
    select?
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
                        (arg.get ⟨2, by simp [List.length]⟩)
  | Op.icmp c _ => LLVM.icmp c
                        (arg.get ⟨0, by simp [List.length]⟩)
                        (arg.get ⟨1, by simp [List.length]⟩)
```

, but do not know how to shorten this.

I looked into `GetElem`, but https://lean-lang.org/functional_programming_in_lean/type-classes/indexing.html#overloading-indexing indicates the following:

> For the sake of flexiblity, GetElem has four parameters:
> 
> The type of the collection
> The type of the index
> The type of elements that are extracted from the collection
> A function that determines what counts as evidence that the index is in bounds

I am afraid I am not sure how to supply `The type of elements`

SSA/Projects/InstCombine/LLVM/Semantics.lean
tobiasgrosser: There are uses of sdiv/udiv/srem/urem autside of the non-questionmark function. We also have two simp-sets, one for eliminating the Option interface and one for eliminating our definitions and falling to BitVec. I feel for the current development this might be useful.

SSA/Core/HVector.lean
alexkeizer: Ah, sorry, I was too quick to suggest `GetElem`; that class does not really work for us because the return type of `get` depends on the particular index being accessed, while `GetElem` mandates a single `type of elements`, as you noted.

I guess I'm a bit concerned about special casing for pairs. The simplest generalization I see is to have `fst` and `snd` work so long as there are at least 1/2 elements, like:
```lean
abbrev fst : HVector A (a :: as) → A a                := head
abbrev snd : HVector A (a₁ :: a₂ :: as) → A a₂        := head ∘ tail
abbrev thd : HVector A (a₁ :: a₂ :: a₃ :: as) → A a₃  := head ∘ tail ∘ tail
```
This is still special casing in some sense, since we don't have a way to access element 4+, but we could in theory keep adding projections as needed (I imagine most operations wont have that many).

Alternatively, we can wrap `get` into our own method that indexes with a `Nat` and uses an auto-tactic for the proof obligations, as in
```lean
/-- `x.getN i` indexes into a `HVector` with a `Nat`,
using `get_elem_tactic` as an auto-tactic to prove that the index is within bounds -/
abbrev getN (x : HVector A as) (i : Nat) (hi : i < as.length := by get_elem_tactic) :
    A (as.get ⟨i, hi⟩) :=
  x.get ⟨i, hi⟩
```

This way, you can write, for example:
```lean
| Op.and _ => LLVM.and (op.getN 0) (op.getN 1)
```
The proof obligation is generally trivial, so `get_elem_tactic` should be able to work it out (and if not, we can extend it and/or use a different auto-tactic).

tobiasgrosser: Perfect. I introduced getN and cleaned up the code. Thank you for the suggestion. I did not know about auto-tactics.

tobiasgrosser:
This has now been updated to the latest `main` branch. Topics that are open:

> Should we make `IntW` an Inductive already or leave it as an `abbrv` and introduce the inductive at a later step

I prefer to leave Option as an inductive for now.
  
> Should we remove the level of indirection that `<op>?` introduces

This is not fully trivial. There are a couple of theorems that use `<op>?`, for example:

```lean
def srem? {w : Nat} (x y : BitVec w) : IntW w :=
  (sdiv? x y).map (fun div => x - div * y)
```

I feel unfolding them in a separate PR would keep this PR smaller.

I also feel that maybe the `<op>?` would allow us to reason about an option-free segment of the semantics without unfolding yet to the BitVec level. However, I am not sure here.

Overall, I would be happy to merge this PR as is. I think it is a step forward and would allow us to regenerate AliveStatements with a more consistent signature.

Feedback is much appreciated.

alexkeizer:
Sounds sensible!

`sdiv?` makes sense to have, since it actually has a codepath that returns none. I guess it might make sense to keep add? around for when we implement nsw/nuw flags.
It's mostly and?, which unconconsitionally returns some that I'd like to see inlined.

Regardless, we can punt that discussion until another PR. I'm happy to have this merged as-is

PR #212, opened 2024-04-06
Do not simplify ssa in alive peephole

tobiasgrosser:
chore: do not simplify SSA in alive_peephole

There is no need to simplify the SSA logic in alive_peephole, if we add the necessary `HVector.toSingle` into the core tactic. As HVector is used in denote tactics, it seems sensible to simplify it away.


bollu:
I am not certain that this will scale, I anticipate that simp may run out of heartbeats if we unfold too much in a single `simp` call. 

Since @tobiasgrosser is sheperding these sets of changes, I am confident that he will refactor if this turns out to be a problem in the future, so LGTM for now.

PR #213, opened 2024-04-07
chore: remove unused simp options

tobiasgrosser:
If we do not need special simp options, we should probably stick to the default options as much as possible or document why we use different options. These two options look as if they potentially increase the computational cost of simp. As our current tests pass without them, we remove them for now. We can always add them back if we find a test case that would require them. We keep `decide := false` for now, as this seems to be a performance optimization.


tobiasgrosser:
Is `decide` set to true in the default simp options? I somehow remembered it was not.

bollu:
LGTM as tests pass.

PR #214, opened 2024-04-07
feat: add scaling test case for LLVM alive

tobiasgrosser:
In previous discussions it was reported that the performance of our alive_peephole tactics has been fragile. To keep tracking it, we add a set of testcases that reaches the limits of maxRecDepth and maxHeartbeats.


lakefile.lean
```diff
@@ -21,7 +21,7 @@ lean_exe ssaLLVMEnumerator {
 
 /-- Separate lib since it takes quite a while to compile -/
 lean_lib AliveExamples {
-  roots := #[`SSA.Projects.InstCombine.AliveAutoGenerated]
+  roots := #[`SSA.Projects.InstCombine.AliveAutoGenerated].append #[`SSA.Projects.InstCombine.ScalingTest]
```
alexkeizer, line 24: Why do you have the `append`, intead of:
```suggestion
  roots := #[`SSA.Projects.InstCombine.AliveAutoGenerated, `SSA.Projects.InstCombine.ScalingTest]
```

I think I'd like the `ScalingTest` to be its own compilation target as well, since I'd rather not have to run it everytime I'd like to test improvements to the Alive automation

alexkeizer (commented):
When talking about fragility, I thought we were mostly talking about proofs breaking because of changes to the definitions and/or simp-set. @bollu have you run into issues with time-outs?

lakefile.lean
tobiasgrosser: The second did not work for me for some reason. I guess I missed the tick before the 2nd SSA. I introduced now a new Target AliveScaling, which should avoid this issue alltogether.

tobiasgrosser:
> When talking about fragility, I thought we were mostly talking about proofs breaking because of changes to the definitions and/or simp-set. @bollu have you run into issues with time-outs?

In https://github.com/opencompl/ssa/pull/212, @bollu said:

> I am not certain that this will scale, I anticipate that simp may run out of heartbeats if we unfold too much in a single simp call.

This PR is there to address this concern. Also, during my recent development I have also seen some increases in the heartbeats that we take. Hence, I wanted to get this covered before merging: https://github.com/opencompl/ssa/pull/210.

alexkeizer:
LGTM!

PR #215, opened 2024-04-08
chore: simplify 'simp' in AliveHandwrittenExample

tobiasgrosser:


SSA/Projects/InstCombine/AliveHandwrittenExamples.lean
```diff
@@ -52,20 +52,13 @@ theorem alive_DivRemOfSelect (w : Nat) :
...
-    InstCombine.Op.denote, HVector.toPair, HVector.toTriple, pairMapM, BitVec.Refinement,
-    bind, Option.bind, pure, DerivedCtxt.ofCtxt, DerivedCtxt.snoc,
-    Ctxt.snoc, ConcreteOrMVar.instantiate, Vector.get, HVector.toSingle,
-    LLVM.and?, LLVM.or?, LLVM.xor?, LLVM.add?, LLVM.sub?,
-    LLVM.mul?, LLVM.udiv?, LLVM.sdiv?, LLVM.urem?, LLVM.srem?,
-    LLVM.sshr, LLVM.lshr?, LLVM.ashr?, LLVM.shl?, LLVM.select?,
-    LLVM.const?, LLVM.icmp?, LLVM.udiv?,
-    HVector.toTuple, List.nthLe, BitVec.bitvec_minus_one,
-    Ctxt.Var.zero_eq_last, Ctxt.Var.succ_eq_toSnoc
+  simp (config := {decide := false }) only [OpDenote.denote,
```
alexkeizer, line 55:
```suggestion
  simp (config := {decide := false}) only [OpDenote.denote,
```

alexkeizer: `decide := false` is also part of the default simp config (this was changed a few versions back), so we could get rid of that option as well. Benefit of keeping it, though, is that we'd be resilient against the default being changed again at some point. Both are valid options, I'll leave it to you to decide

alexkeizer (approved):
LGTM

PR #217, opened 2024-04-09
chore: delete `Scf`

alexkeizer:
The `Scf` example was superseded by `ScfFunctor`


tobiasgrosser:
I do not know about this redundancy, but @bollu already confirmed this earlier. I think we can merge this.

PR #222, opened 2024-04-11
chore: update to lean 4.7.0

tobiasgrosser:


SSA/Core/Framework.lean
```diff
@@ -3,6 +3,8 @@
 import SSA.Core.ErasedContext
 import SSA.Core.HVector
 import Mathlib.Data.List.AList
+import Mathlib.Data.Matrix.Basic
```
tobiasgrosser, line 6: This include is a hack. I actually want to include `import Mathlib.Data.FinSet.Piecewise` here and keep the `import Mathlib.Data.Matrix.Basic` in `SSA/Projects/Tensor2D/Tensor2D.lean`. Unfortunately this fails with the following error in Tensor2D.lean:
import Mathlib.Data.Finset.Piecewise failed, environment already contains 'Finset.piecewise_idem_left' from Mathlib.Data.FinSet.Piecewise

alexkeizer: The setup you describe works just fine on my machine. I suspect your build cache was somehow corrupted: try `lake clean && lake exe cache get! && lake build`.
I've added a commit that minimizes the import in Framework, let's see what CI makes of it

tobiasgrosser:
This seems to go through. Thank you @alexkeizer for fixing this. I merge this now.

PR #223, opened 2024-04-12
chore: fix the try revert chain in simp_alive_base_bash

tobiasgrosser:
This was broken for two reasons:

  - The variables were called x0, x1, ... and not v0, v1 ...
  - By writing all try statements on a single line, the `try`s were associated in a way that a failure in the first revert meant later reverts were not executed

Unfortunately, the reverted variables conflict with the apply statements in `AliveAutoGenerated.lean` so we remove them once again with an `intros`. Given that this has now been investigated, I think it still makes sense to commit this code. It will probably require a further update as soon as we regenerate the AliveAutoGenerated.lean file.


alexkeizer:
It's bit jank to revert the variables and immediately re-introduce them: I'd say having the broken reverts is a good reminder that we should go through and fix the AliveAutoGenerated file at some point. That said, fixing that file is maybe not the highest priority at the moment, so I'm OK with this being merged right now.

PR #227, opened 2024-04-16
feat: do not unfold binds in alive_* tactics

tobiasgrosser:
We previously unfolded binds too eagerly, which led to unnecessary complex proof states such as:

```lean
⊢ ∀ (e e_1 : Option (BitVec w)),
  (match
      match e_1, fun x' =>
        match e, fun y' => some (x' &&& y') with
        | none, x => none
        | some a, b => b a with
      | none, x => none
      | some a, b => b a,
      fun x' =>
      match e, fun y' => some (x' &&& y') with
      | none, x => none
      | some a, b => b a with
    | none, x => none
    | some a, b => b a) ⊑
    match e_1, fun x' =>
      match e, fun y' => some (x' &&& y') with
      | none, x => none
      | some a, b => b a with
    | none, x => none
    | some a, b => b a
```

After this patch this is turned into:

```lean
⊢ ∀ (e e_1 : Option (BitVec w)),
  (do
      let x ← e_1
      let x_1 ← e
      let y' ← e
      pure (x &&& x_1 &&& y')) ⊑
    do
    let x' ← e_1
    let y' ← e
    pure (x' &&& y')
```

This is a lot more readable.

The reduced unfolding should lead to simpler expressions, even though only minor improvements in performance (54s -> 52s) are visible on our AliveScaling examples.

### DIFF
--- a/SSA/Projects/InstCombine/AliveHandwrittenExamples.lean
+++ b/SSA/Projects/InstCombine/AliveHandwrittenExamples.lean
@@ -67,7 +67,5 @@ theorem alive_DivRemOfSelect (w : Nat) :
      . subst h
        simp
      . subst h; simp
-       cases' x with vx <;>
-       cases' y with vy <;> simp [LLVM.udiv?]
 
 end AliveHandwritten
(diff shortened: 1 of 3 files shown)


bollu:
LGTM, seems to be a mechnical change from some -> pure to better enable rewrites to kick in.

PR #228, opened 2024-04-16
refactor: generalize `simp_alive_case_bash` to an arbitrary number of variables

alexkeizer:
This complements the recent changes in the core `simp_alive_peephole` tactic to make it not limited to just 5 variables.

Also, I took the opportuniuty to introduce a new simp-set, and squeeze the simp used in `simp_alive_case_bash` to just this simp-set (were previously it was just using `simp` with the global simp-set).


SSA/Projects/InstCombine/Tactic.lean
```diff
@@ -15,46 +15,24 @@ open Ctxt
...
+    try (
+      intro (v : Option (_))    -- Introduce the variable,
+      rcases v with _|x         -- Do the case distinction
+      <;> simp only [BitVec.Refinement.refl, BitVec.Refinement.some_some, bind_assoc]
+      --  ^^^^^^^^^^^^^^^^^^^^^^^^ Simplify, in the hopes that the `none` case is trivially closed
+      <;> simp_alive_case_bash  -- Recurse, to case bash the next variable (if it exists)
+      <;> revert x              -- Finally, revert the variable we got in the `some` case, so that
+                                --   we are left with a universally quantified goal of the form:
+                                --   `∀ (x₁ : BitVec _) ... (xₙ : BitVec _), ...`
+    ))
```
bollu, line 35:
```suggestion
macro_rules
  | `(tactic| simp_alive_case_bash) => `(tactic|
    repeat match goal with
    | `(fun ($v : Option (_)) => _) => 
       rcases $v with _|x
       <;> simp only [BitVec.Refinement.refl, BitVec.Refinement.some_some, bind_assoc]
       <;> revert x     
  )
```

bollu: (Note: Writing this suggestion to show how match goal would help here)

bollu:
CI fails:

```
./././SSA/Projects/InstCombine/AliveAutoGenerated.lean:698:2: error: tactic 'apply' failed, failed to unify
  (?X ||| ?C1) &&& ?C2 = (?X ||| ?C1 &&& ?C2) &&& ?C2
with
  (do
      let x ← e✝
      let x_1 ← e✝¹
      let y' ← e✝²
      pure ((x ||| x_1) &&& y')) ⊑
    do
    let x ← e✝
    let x_1 ← e✝¹
    let x_2 ← e✝²
    let y' ← e✝²
    pure ((x ||| x_1 &&& x_2) &&& y')
```

probably because we no longer return `some`, but we return `pure`, so we need some more lemmas to cleanup the `none`case.

alexkeizer:
> CI fails:
> 
> ```
> ./././SSA/Projects/InstCombine/AliveAutoGenerated.lean:698:2: error: tactic 'apply' failed, failed to unify
>   (?X ||| ?C1) &&& ?C2 = (?X ||| ?C1 &&& ?C2) &&& ?C2
> with
>   (do
>       let x ← e✝
>       let x_1 ← e✝¹
>       let y' ← e✝²
>       pure ((x ||| x_1) &&& y')) ⊑
>     do
>     let x ← e✝
>     let x_1 ← e✝¹
>     let x_2 ← e✝²
>     let y' ← e✝²
>     pure ((x ||| x_1 &&& x_2) &&& y')
> ```
> 
> probably because we no longer return `some`, but we return `pure`, so we need some more lemmas to cleanup the `none`case.

Ah, I did fix this locally, but it seems that commit got lost in a rebase I did to clean up the history... oops

SSA/Projects/InstCombine/Tactic.lean
alexkeizer: I guess the main benefit of match_goal here is that it doesn't swallow errors like `try` does. We could achieve the same by wrapping the tactic body in:
```first
    | fail_if_success (intro (v : Option (_)))
    -- If there is no variable to introduce, the tactic succeeds, but does nothing
    -- This is similar to `try`, except `first ...` does not swallow any errors that occur in
    -- the later tactics
    | intro (v : Option _); ...
`

tobiasgrosser (approved):
This looks lovely. I wonder what @bollu thinks.

bollu:
LGTM, thanks for polishing this!

PR #229, opened 2024-04-16
chore: drop unnecessary open

tobiasgrosser:
This is factored out from https://github.com/opencompl/ssa/pull/191 to minimize the diff to main.


tobiasgrosser:
Merged as trivial.

PR #230, opened 2024-04-16
feat: add Op.denote as a OpDenote instance for FHE

bollu:
This peels code from the mega PR https://github.com/opencompl/ssa/pull/196 into the first bite sized chunk, where we introduce the `OpDenote` instance.


alexkeizer:
Seems like trivially correct diff to me, feel free to merge

PR #231, opened 2024-04-16
feat: Implement [fhe_com |...] to write FHE examples with sugar.

bollu:
This adds a new `[fhe_com |...]`, mimicking the `[alive_com|...]`, that provides a prettier syntax to write FHE examples in.

---
Code peeled from pre-paper submission PR: https://github.com/opencompl/ssa/pull/196.
Code stacked on: https://github.com/opencompl/ssa/pull/230.

### DIFF
--- a/SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
+++ b/SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
@@ -91,15 +91,6 @@ def mlir2fhe (reg : MLIR.AST.Region 0) :
 end MkFuns -- we don't want q and i here anymore
 
 open Qq MLIR AST Lean Elab Term Meta in
-/--
-NOTE: Wanting Stuck Terms During Strong Normalization of Com
-
-See that we unfold with `TransparencyMode.all`. We do this so we can expose Com.denote
-fully. This is a crazy hack, and interacts extremely poorly with complex definitions found in FHE.
-
-One neater solution is a `match goal` like tactic to match on the proof state
-and run the correct equation.
--/
 elab "[fhe_com" qi:term "," ni:term "," hq:term " | " reg:mlir_region "]" : term => do
   let ast_stx ← `([mlir_region| $reg])
   let ast ← elabTermEnsuringTypeQ ast_stx q(Region 0)
--- a/SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
+++ b/SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
@@ -1,0 +1,127 @@
+/-
+Syntax definitions for FHE, providing a custom [fhe_com|...] with syntax sugar.
+
+Authors: Andrés Goens<andres@goens.org>
+-/
+import SSA.Projects.InstCombine.LLVM.Transform
+import SSA.Projects.MLIRSyntax.AST
+import SSA.Projects.MLIRSyntax.EDSL
+import SSA.Projects.FullyHomomorphicEncryption.Basic
+
+open MLIR AST Ctxt
+open Polynomial -- for R[X] notation
+
+section MkFuns
+variable {q : Nat} {n : Nat} [Fact (q > 1)]
+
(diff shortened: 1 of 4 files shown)


SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
```diff
@@ -0,0 +1,127 @@
...
+def mlir2fhe (reg : MLIR.AST.Region 0) :
+    MLIR.AST.ExceptM (Op q n) (Σ (Γ : Ctxt (Ty q n)) (ty : (Ty q n)), Com (Op q n) Γ ty) := MLIR.AST.mkCom reg
+
+end MkFuns -- we don't want q and i here anymore
+
+open Qq MLIR AST Lean Elab Term Meta in
+/--
+NOTE: Wanting Stuck Terms During Strong Normalization of Com
+
+See that we unfold with `TransparencyMode.all`. We do this so we can expose Com.denote
```
tobiasgrosser, line 97: Is this comment still up-to-date. It seems we unfold with TransparencyMode = default, no?

SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
```diff
@@ -0,0 +1,127 @@
...
+fully. This is a crazy hack, and interacts extremely poorly with complex definitions found in FHE.
+
+One neater solution is a `match goal` like tactic to match on the proof state
+and run the correct equation.
```
alexkeizer, line 101: In this case, `simp-proc`s should suffice, since we don't need to match on the context. Or really, just proper `Expr` matching, since we don't have to arbitrarily search for sub-expressions.

tobiasgrosser:
I rebased this on main, as #230 has already been merged.

alexkeizer (commented):
I'm not a big fan of just copy-pasting the same blob of elaborator meta-code for each dialect, and then tweaking it. This is a pretty big tech-debt risk.

For Alive we had to have separate meta-code, to deal with the meta vars, but FHE does not have meta variables, right? Have we attempted a common elaborator core for FHE and, say, the simple PaperExamples dialect?

Then again, I don't want to block the entire FHE effort on this, so feel free to merge if it turns out having this shared meta-code is difficult.

SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
bollu: You're right. I opt to drop the comment altogether, and revive it when I bring in the new operation in a future match.

bollu:
@alexkeizer would you like to take a stab at nicer meta-code?

The purpose of breaking the PR into reviewable pieces is precisely to get this kind of feedback / iteration.

goens:
> I'm not a big fan of just copy-pasting the same blob of elaborator meta-code for each dialect, and then tweaking it. This is a pretty big tech-debt risk.
> 
> For Alive we had to have separate meta-code, to deal with the meta vars, but FHE does not have meta variables, right? Have we attempted a common elaborator core for FHE and, say, the simple PaperExamples dialect?
> 
> Then again, I don't want to block the entire FHE effort on this, so feel free to merge if it turns out having this shared meta-code is difficult.

I quite agree with this sentiment. I think we should try and factor this into nice typeclasses + meta code. To the second point (and general) I would suggest when issues like these come up to add an issue with the merge (rather than blocking)

bollu:
@alexkeizer 

> but FHE does not have meta variables

IIRC, we have some kind of "meta variable", of the form of the FHE scheme parameters (`q`, `n`, and `Fact (n > 1)`).

bollu:
@alexkeizer let me know if you want to take a stab at cleaning up the meta code, or feel invited to add an issue to track that we should cleanup meta code and merge this as-is.

alexkeizer:
I've opened #232, and hit merge for this one.
This will break the purity-annotation branch again, could you merge main and fix that one after this goes through?

PR #191, opened 2024-02-28
feat: add purity annotation

bollu:
Add a custom field into each `Op` that declares if its signature is pure or not. If it is pure, it returns `a`, and if impure, it returns `IO a`. The default signature object is pure.

PR #233, opened 2024-04-17
chore: update mathlib

tobiasgrosser:


tobiasgrosser:
Merged into https://github.com/opencompl/ssa/pull/191 successfully. Hence, I will push this to main and then update #191 as suggested by @bollu.

Issue #232, opened 2024-04-16
Cleanup *_com elaborator meta-code


alexkeizer:
Currently, we share no code between `alive_com`, `simple_com`, and `fhe_com` (as being developed in #231).
`alive_com` was originally split off because of the meta variable instantiation that is alive-specific, but we should still be able to factor out the common parts into a generic meta-function, at least.

alexkeizer:
@bollu Could you document in which way `fhe_com` differs from `simple_com`? I believe there was trickery involved because the reduction caused problems?

PR #235, opened 2024-04-17
refactor: move `MLIRSyntax` project into `Core`

alexkeizer:
We're already using the DSL in multiple other projects, so it belongs in Core

This is a first step towards #232


alexkeizer:
That's totally fine with me. Then I might change `MLIRSyntax.Parser` to `MLIRSyntax.GenericParser` to emphasize this is the part that parses generic syntax into AST, not the part that parses into a specific dialect, does that sound sensible @bollu?

PR #237, opened 2024-04-17
feat: add example from paper for FHE.

bollu:
To make this example work, we need a definition of `R.ofZ` that is computable, and which is not agressively unfolded. This is ensured by our `axiom` trick.

Code peeled from pre-paper submission PR: https://github.com/opencompl/ssa/pull/196.
Code stacked on: https://github.com/opencompl/ssa/pull/234

---

Closes https://github.com/opencompl/ssa/pull/196 (This PR fully integrates the features of https://github.com/opencompl/ssa/pull/196 into main.)

### DIFF
--- a/SSA/Projects/FullyHomomorphicEncryption/Rewrites.lean
+++ b/SSA/Projects/FullyHomomorphicEncryption/Rewrites.lean
@@ -82,7 +82,8 @@ infixl:50 "**" => irreduciblePow
 
 variable {q : Nat} {n : Nat} [hq : Fact (q > 1)]
 
--- code generator does not support recursor 'Decidable.rec' yet, consider using 'match ... with' and/or structural recursion
+-- We mark this as noncomputable due to the presence of poly.const, which creates a value of type R.
+-- This operation is noncomputable, as we use `coe` from `Int` to `R`, which is a noncomputable instance. 
 noncomputable def lhs := [fhe_com q, n, hq| {
 ^bb0(%a : ! R):
   %one_int = "arith.const" () {value = 1}: () -> (i16)
--- a/SSA/Projects/FullyHomomorphicEncryption/Rewrites.lean
+++ b/SSA/Projects/FullyHomomorphicEncryption/Rewrites.lean
@@ -145,6 +145,6 @@ noncomputable def p1 : PeepholeRewrite (Op q n) [.polynomialLike] .polynomialLik
       apply Poly.add_f_eq
       done
     }
-/- 'p1' depends on axioms: [propext, Classical.choice, Quot.sound, ROfZComputable_stuck_term, ROfZComputable_def] -/
-#print axioms p1
+/--info: 'p1' depends on axioms: [propext, Classical.choice, Quot.sound, ROfZComputable_stuck_term, ROfZComputable_def] -/
+#guard_msgs in #print axioms p1
 end ExampleModulo
--- a/SSA/Projects/FullyHomomorphicEncryption/Rewrites.lean
+++ b/SSA/Projects/FullyHomomorphicEncryption/Rewrites.lean
@@ -138,11 +138,7 @@ noncomputable def p1 : PeepholeRewrite (Op q n) [.polynomialLike] .polynomialLik
       /- ⊢ a + ((Ideal.Quotient.mk (Ideal.span {f q n})) ((Polynomial.monomial (2**n)) 1) + 1) = a -/
       intros a
       unfold irreduciblePow
-      --have hgenerator : f 2 3 = (Polynomial.monomial 8 1) + 1  := by simp [f, Polynomial.X_pow_eq_monomial]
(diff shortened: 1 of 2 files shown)


alexkeizer:
To clarify: the axiom trick is needed because of the reduction we do in the `fhe_com` elaborator?

bollu:
@alexkeizer that's right.

bollu:
@alexkeizer actually, on second thought, I think there are two places we agressively reduce: one is in `simp_peephole`/`change_mlir_context`, and one is at the `fhe_com`. I think it's the aggressiveness of `simp_peephole` / `change_mlir_context` that used to cause stuff to explode.

bollu:
@alexkeizer My hope is that the example is self-contained enough for someone (nudge nudge @tobiasgrosser ) to investigate if it's possible to eliminate the need for the axiom.

alexkeizer:
Good to know: we could (should) change `change_mlir_context` to only reduce to a sort of transitive whnf, where it is of the form `_ :: _ :: ... :: []` (with the `cons` being made explicit), but the elements don't have to be reduced. That would solve your issue as well, then?

bollu:
@alexkeizer that might, but I would not take my word for it. I'd have to play with the example again to remember precisely what caused me to need an `axiom` instead of just an `implemented_by`.

alexkeizer:
I'd like to give it a try: reducing less aggressively seems like a win regardless of if it fixes this exact issue.

SSA/Projects/FullyHomomorphicEncryption/Rewrites.lean
```diff
@@ -72,3 +72,79 @@ noncomputable def p1 : PeepholeRewrite (Op q n) [.polynomialLike, .polynomialLik
...
+section ExampleModulo
+
+@[irreducible]
+def irreduciblePow (q n : Nat) : Nat := q^n
+infixl:50 "**" => irreduciblePow
+
+
+variable {q : Nat} {n : Nat} [hq : Fact (q > 1)]
+
+-- code generator does not support recursor 'Decidable.rec' yet, consider using 'match ... with' and/or structural recursion
```
alexkeizer, line 85: What does this comment refer to?

SSA/Projects/FullyHomomorphicEncryption/Rewrites.lean
```diff
@@ -72,3 +72,79 @@ noncomputable def p1 : PeepholeRewrite (Op q n) [.polynomialLike, .polynomialLik
...
+  "return" (%v1) : (! R) -> ()
+}]
+
+def rhs := [fhe_com q, n, hq | {
+^bb0(%a : ! R):
+  "return" (%a) : (! R) -> ()
+}]
+
+/- 'lhs' depends on axioms: [propext, Quot.sound, Classical.choice, ROfZComputable_stuck_term] -/
+#print axioms lhs
```
alexkeizer, line 103:
```suggestion
/--info: 'lhs' depends on axioms: [propext, Quot.sound, Classical.choice, ROfZComputable_stuck_term] -/
#guard_msgs in #print axioms lhs
```

This way, `lake build` is less noisy. Plus, we wouldn't have to manually compare axioms anymore; a change in axioms would become a build/CI error

SSA/Projects/FullyHomomorphicEncryption/Rewrites.lean
```diff
@@ -72,3 +72,79 @@ noncomputable def p1 : PeepholeRewrite (Op q n) [.polynomialLike, .polynomialLik
...
+      unfold irreduciblePow
+      have hgenerator : f q n - (1 : Polynomial (ZMod q)) = (Polynomial.monomial (R := ZMod q) (2^n : Nat) 1)  := by simp  [f, Polynomial.X_pow_eq_monomial]
+      rw [← hgenerator]
+      have add_congr_quotient : ((Ideal.Quotient.mk (Ideal.span {f q n})) (f q n - 1) + 1)  = ((Ideal.Quotient.mk (Ideal.span {f q n})) (f q n )) := by simp
+      rw [add_congr_quotient]
+      apply Poly.add_f_eq
+      done
+    }
+/- 'p1' depends on axioms: [propext, Classical.choice, Quot.sound, ROfZComputable_stuck_term, ROfZComputable_def] -/
+#print axioms p1
```
alexkeizer, line 149:
```suggestion
/--info: 'p1' depends on axioms: [propext, Classical.choice, Quot.sound, ROfZComputable_stuck_term, ROfZComputable_def] -/
#guard_msgs in #print axioms p1
```

alexkeizer (commented):
I'd say this is fine to merge after you've addressed my (minor) comments

SSA/Projects/FullyHomomorphicEncryption/Rewrites.lean
bollu: It explains why the code is marked `noncomputable`. The `Decidable.rec` is from the `poly.const`, IIRC.

bollu: Addressed, I changed the comment.

PR #238, opened 2024-04-17
refactor: make `change_mlir_context` reduce less aggressively

alexkeizer:
In particular, don't needlessly reduce individual types in a context

This might help mitigate an issue in #237, where we had to introduce an axiom to prevent reduction


bollu:
LGTM

PR #241, opened 2024-04-17
chore: remove use of axiom in FHE that was a workaround for aggressive reducibility

bollu:
Closes https://github.com/opencompl/ssa/issues/240

https://github.com/opencompl/ssa/pull/239 toned down how aggressive our unfolding is (to be Reducibility.default) [thanks @alexkeizer !], Thus, we can now use `irreducible_def` to hide the implementation of `R.ofZComputable`. This eliminates our reliance on axioms.
This generally improves the trustworthiness of our development.

### DIFF
--- a/SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
+++ b/SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
@@ -1,11 +1,10 @@
-/- := rfl
+/-
 Syntax definitions for FHE, providing a custom [fhe_com|...] with syntax sugar.
 
-Authors: Andrés Goens<andres@goens.org>
+Authors: Andrés Goens<andres@goens.org>, Siddharth Bhat<siddu.druid@gmail.com>
 -/
 import SSA.Core.MLIRSyntax.EDSL
 import SSA.Projects.FullyHomomorphicEncryption.Basic
-import Mathlib.Tactic.IrreducibleDef
 
 open MLIR AST Ctxt
 open Polynomial -- for R[X] notation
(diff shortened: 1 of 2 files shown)


SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
```diff
@@ -1,10 +1,11 @@
-/-
+/- := rfl
```
tobiasgrosser, line 1: What is this?

SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
```diff
@@ -105,11 +91,39 @@ def ROfZComputable_impl (z : ℤ) : R q n :=
...
+2. It is equal to the noncomputable definition `(coe z : R q n)`
+3. It does not unfold when using `simp_peephole`, so we can rewrite using the equation
+     `ROfZComputable_def` after calling `simp_peephole`.
+
+What we really want is to have the following:
+
+```lean
 @[implemented_by ROfZComputable_impl]
-axiom ROfZComputable_stuck_term (q n : Nat) (z : ℤ) : R q n
+irreducible_noncomputable_def ROfZComputable_stuck_term (q n : Nat) (z : ℤ) :=
```
alexkeizer, line 106:
```suggestion
noncomputable irreducible_def ROfZComputable_stuck_term (q n : Nat) (z : ℤ) :=
```
`irreducible_noncomputable_def` is not a thing, but `irreducible_def` does take modifiers just like a normal `def`

SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
```diff
@@ -105,11 +91,39 @@ def ROfZComputable_impl (z : ℤ) : R q n :=
...
+
+However, the problem is that `irreducible_noncomputable_def` does not exist, so we fake it,
+by creating a computable `_implemented_by_shim`, and then making this shim an `irreducible_def`.
+-/
+
+@[implemented_by ROfZComputable_impl]
+def ROfZ_computable_stuck_term_implemented_by_shim (q n : Nat) (z : ℤ) := (↑ z : R q n)
+
+irreducible_def ROfZComputable_stuck_term (q n : Nat) (z : ℤ) :=
+  ROfZ_computable_stuck_term_implemented_by_shim q n z
```
alexkeizer, line 122:
```suggestion
@[irreducible] def ROfZComputable_stuck_term (q n : Nat) (z : ℤ) :=
  ROfZ_computable_stuck_term_implemented_by_shim q n z
```
Unless I'm missing a reason why plain `@[irreducible] def` doesn't work, but it seems to do fine on machine when I tested locally. This way, we don't have to rely on the Mathlib version.

alexkeizer (commented):
Nice job getting rid of the axiom!

SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
bollu: Actually, I don't get it, how do I access the definition?

SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
bollu: Thanks, stray incorrect code.

SSA/Projects/FullyHomomorphicEncryption/Syntax.lean
```diff
@@ -105,11 +90,30 @@ def ROfZComputable_impl (z : ℤ) : R q n :=
   }
   R.fromPoly p
 
-@[implemented_by ROfZComputable_impl]
-axiom ROfZComputable_stuck_term (q n : Nat) (z : ℤ) : R q n
+/-#
+Explanation of what's going on:
+- We want to have a `cstComputable : Z -> R`, which achieves three properties:
+1. It is computable, since the meta code needs it to be computable for `[fhe_icom|..]` to elaborate.
+2. It is equal to the noncomputable definition `(coe z : R q n)`
```
alexkeizer, line 97: > It is equal to the noncomputable definition `(coe z : R q n)`

Does it have to be def-eq? My guess would be no, since this equality used to be an axiom, so I get the feeling we can get rid of the `implemented_by` as well, if we can prove that `ROfZComputable_impl` is equal to `coe`.

alexkeizer (approved):
LGTM. I do have the feeling we should get rid of the `implemented_by` as well, but I'll conceded that this is a non-trivial amount of extra effort, and the PR as-is is already a clear improvement

PR #245, opened 2024-04-18
chore: drop two unnecessary lemmas from simp_peephole

tobiasgrosser:
These were added in 121e9c7fa316edf8fc1824da136cb6a2cb3ba08c in an attempt to reduce the delta of the FHE branch, but were really not needed.

### DIFF
--- a/SSA/Core/Tactic.lean
+++ b/SSA/Core/Tactic.lean
@@ -49,20 +49,6 @@ elab "change_mlir_context " V:ident : tactic => do
 
 end
 
-@[simp]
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
This PR reduces the delta between https://github.com/opencompl/ssa/pull/191 and main.

alexkeizer:
Thanks for picking this up! Could you delete the theorems themselves as well? Then I'd say it's good to go

alexkeizer:
Perfect :+1:

PR #246, opened 2024-04-18
change: redefine alive dsl in terms of generic `elabIntoCom`

alexkeizer:
This PR uses `elabIntoCom` to simplify the definition of `alive_com`. This means first `elabIntoCom` will normalize the `Com` (which still has object metavariables), and only afterwards do we instantiate the metavariables and reduce it *again*.

To reduce the impact of this double reduction, we also weaken the reduction done by `elabIntoCom` analogously to `ctxtNf`; we recursively call `whnf` on the body of a `Com.lete`, instead of fully reducing. This is enought for `FHE` and `PaperExamples`. Alive, however, does need full reduction to get rid of the meta-var instantiation code, so the second reduction in `alive_com` is still `reduce`.

Minor testing suggests this version of `alive_com` is still faster than what we currenlty have in main: `and_sequence_30_lhs` (of `ScalingTest`) in the current branch takes about 175000 heartbeats compared to 210000 in main (9305ece)

### DIFF
--- a/SSA/Core/MLIRSyntax/EDSL.lean
+++ b/SSA/Core/MLIRSyntax/EDSL.lean
@@ -13,6 +13,35 @@ namespace SSA
 open Qq Lean Meta Elab Term
 open MLIR.AST
 
+/-- `ctxtNf` reduces an expression of type `Ctxt _` to something in between whnf and normal form.
+`ctxtNf` recursively calls `whnf` on the tail of the list, so that the result is of the form
+  `a₀ :: a₁ :: ... :: aₙ :: [] `
+where each element `aᵢ` is not further reduced -/
+partial def ctxtNf (as : Expr) : MetaM Expr := do
+  let as ← whnf as
+  match_expr as with
+    | List.cons _ a as =>
+        let as ← ctxtNf as
+        mkAppM ``Ctxt.snoc #[as, a]
+    | _ => return as
+
+/-- `comNf` reduces an expression of type `Com` to something in between whnf and normal form.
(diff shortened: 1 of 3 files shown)


tobiasgrosser:
Nice. Do you expect any impact on:

```lean
 theorem alive_Select_858  (w : Nat)   : alive_Select_858_src w  ⊑ alive_Select_858_tgt 1  := by
   unfold alive_Select_858_src alive_Select_858_tgt
   stop
   simp_alive_peephole -- fails to clear the context
```

bollu:
LGTM, I'm trying to merge it into the other PR to see if it simplifies as expected, but I assume it does?

bollu:
@alexkeizer feel invited to pull it in

alexkeizer:
> Nice. Do you expect any impact on:
> 
> ```lean
>  theorem alive_Select_858  (w : Nat)   : alive_Select_858_src w  ⊑ alive_Select_858_tgt 1  := by
>    unfold alive_Select_858_src alive_Select_858_tgt
>    stop
>    simp_alive_peephole -- fails to clear the context
> ```

Sadly, no, I haven't been able to fix what's going on there. I tried `reduceAll` (which is `reduce` with `skipProofs` and `skipTypes` set to false), but that broke a lot of other things.
This PR is mostly about cleaning up the base line, before digging deeper

alexkeizer:
Coincidentally, this change does seem to fix the huge slowdown we saw in #191. I merged locally, and got about 

> LGTM, I'm trying to merge it into the other PR to see if it simplifies as expected, but I assume it does?

The other being #191 ? I've got that merged locally

PR #247, opened 2024-04-19
Broken alive autogenerated in case of mismatched types (alternative branch)

alexkeizer:
By normalising variables to `toSnoc _ <| toSnoc _ <| ... <| last _ _`, we enables reduction of the types far better than the "proper" normal form (i.e., `Subtype.mk`, where the type of the variable is determined by the proof and thus not easily reduced).

This fixes the bug we saw in #224, but in the process has to resort to some pretty aggressive reduction, breaking FHE once more.


tobiasgrosser:
@alexkeizer, should we close this and re-open in case work on this proceeds?

alexkeizer:
Closing this, since we went the alternative route of adding simp sets to control the reduction

PR #248, opened 2024-04-22
chore: use rfl in proofs where possible

tobiasgrosser:


SSA/Core/ErasedContext.lean
```diff
@@ -304,10 +304,10 @@ def Valuation.ofPair [TyDenote Ty] {t₁ t₂ : Ty} (v₁: ⟦t₁⟧) (v₂ : 
 
 @[simp]
 theorem Valuation.ofPair_fst [TyDenote Ty] {t₁ t₂ : Ty} (v₁: ⟦t₁⟧) (v₂ : ⟦t₂⟧) :
-  (Ctxt.Valuation.ofPair v₁ v₂) ⟨0, by simp⟩ = v₁ := rfl
+  (Ctxt.Valuation.ofPair v₁ v₂) ⟨0, by rfl⟩ = v₁ := rfl
```
bollu, line 307: This should not matter, because it's at the *pattern matching* side of things, where proof irrelevance indeed holds (although @alexkeizer said that when he was working on `mathlib`, he has seen examples where even this is not true).

SSA/Core/ErasedContext.lean
```diff
@@ -304,10 +304,10 @@ def Valuation.ofPair [TyDenote Ty] {t₁ t₂ : Ty} (v₁: ⟦t₁⟧) (v₂ : 
...
 @[simp]
 theorem Valuation.ofPair_snd [TyDenote Ty] {t₁ t₂ : Ty} (v₁: ⟦t₁⟧) (v₂ : ⟦t₂⟧) :
-  (Ctxt.Valuation.ofPair v₁ v₂) ⟨1, by simp⟩ = v₂ := rfl
+  (Ctxt.Valuation.ofPair v₁ v₂) ⟨1, by rfl⟩ = v₂ := rfl
```
bollu, line 310: Same as above:

> This should not matter, because it's at the *pattern matching* side of things, where proof irrelevance indeed holds (although @alexkeizer said that when he was working on `mathlib`, he has seen examples where even this is not true).

bollu (approved):

SSA/Core/ErasedContext.lean
alexkeizer: Just to clarify: I've had people tell me in mathlib reviews that they observed proofs being relevant when it comes to rewrite lemmas, so this is third-hand info, basically folklore at this point

PR #249, opened 2024-04-22
Change our use of Functor.map (via <$>) to a more targeted use of ..

tobiasgrosser:
RegionSignature.map and Signature.map. This makes proof states easier to read, and allows for unfolding one of the instances of Functor.map at a time.

This is pulled out of https://github.com/opencompl/ssa/pull/225 as an independent useful change.

Co-authored-by: Siddharth Bhat <siddu.druid@gmail.com>

### DIFF
--- a/SSA/Core/Framework.lean
+++ b/SSA/Core/Framework.lean
@@ -503,12 +503,17 @@ theorem Lets.denote_getExpr {Γ₁ Γ₂ : Ctxt Ty} : {lets : Lets Op Γ₁ Γ�
 
 section Map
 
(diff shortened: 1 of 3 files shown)


tobiasgrosser:
@bollu authored this change, so we should get a review from @alexkeizer or @goens.

goens (changes requested):
The code lgtm, just a bit worried how this will interact with #191 (I guess we'll also merge this with that PR if merged?)

I have one comment about the metadata though: If @bollu authored this change, maybe he should be in the commits? At least with a co-authored-by tag (see https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors).

Issue #250, opened 2024-04-22
Add default arguments to OpSignature to make Id default


bollu:
In #191, a dialect consists of `Op`, `Ty`, *and* a monad `m`, leading to rather verbose variable/argument lists.

# Possible Solutions

## Make `Id` default
Often, when writing pure dialects, it is overhead to mention the monad. Consider making `Id` the default monad. 

## Introduce a `Dialect` structure
This structure would be defined as 
```
structure Dialect 
  Op : Type
  Ty : Type
  m : Type -> Type
```

This way, we can have classes like `OpSignature`/`OpDenote`, but also datastructures like `Com/Expr` take just a `d : Dialect` argument. Consider, for example, the following code paraphrased from #191 
```lean
variable (Op : Type) {Ty : Type} {m : Type → Type} [OpSignature Op Ty m]
mutual 
inductive Expr : (Γ : Ctxt Ty) → (eff : EffectKind) → (ty : Ty) → Type := ...
```
after introducing a `Dialect` struct, this would be shortened to
```lean
variable (d : Dialect) [OpSignature d]
mutual 
inductive Expr : (Γ : Ctxt d.Ty) → (eff : EffectKind) → (ty : d.Ty) → Type := ...
```

Besides being less verbose, it has the benefit of making it much easier to change what exactly constitutes a dialect in the future, since we only have to add/remove a field of the `Dialect` definition, instead of having to change every mention of `OpSignatere`,`OpDenote`, `Com`, `Expr`, `Lets`, etc. to include the new/removed variable (as we had to do in #191). In particular, making this change before #191 is merged, would make that diff substantially smaller.

alexkeizer:
I agree with the problem, not sure I like the solution.

How about making a
```lean
structure Dialect 
  Op : Type
  Ty : Type
  m : Type -> Type
```

That would also make future refactors easier, without implicit/default magic

bollu:
I do not have a good sense of the tradeoffs, and would prefer to refactor this after #191 has been merged — we can try it out and see how it feels?

alexkeizer:
Then I'd have a project management nit: this and #252 should be a single GH issue (since it's one "issue" with the code-base), which then mentions both possible solutions, so that there is a single clear home for discussion.
I'll close #252, and edit this issue to include the `Dialect` solution

alexkeizer:
> I do not have a good sense of the tradeoffs, and would prefer to refactor this after #191 has been merged — we can try it out and see how it feels?

[...]

PR #254, opened 2024-04-23
refactor: bundle `Op` and `Ty` into a `Dialect` struct

alexkeizer:
Morally, we currently identify a dialect with it's `Op` type, which leads to things like 
```lean
abbrev ExceptM  Op [OpSignature Op Ty] := Except (TransformError Ty)
```
Notice how `ExceptM` doesn't actually need `Op`, or its `OpSignature` instance, indeed a comment just above points out:
> Even though we technically only need to know the `Ty` type, our typeclass hierarchy is fully based on `Op`.
> It is thus more convenient to incorporate `Op` in these types, so that there will be no ambiguity errors.

Since we're already pulling in all arguments to `OpSignature` even when we don't need them, it makes sense to bundle `Op` and `Ty` explicitly into a structure 
```lean
structure Dialect where
  Op : Type
  Ty : Type
```
This PR implements this change, implementing one of the alternatives as presented in #250.

The diff is rather huge, but the change is almost completely mechanical, using find/replace.

A notable exception to this is that `deriving Repr` in the definition of `Lets` no longer works, so I've had to manually implement `Repr` instead. Presumably the derivation meta-code knows to insert `[Repr Op] [Repr Ty]` bounds when `Op` and `Ty` are arguments, but it doesn't know to insert `[Repr d.Op]` and `[Repr d.Ty]` bounds when `Lets` has an argument `d : Dialect`.


tobiasgrosser:
This looks like a good idea. Can you maybe write 1-2 paragraphs of context, potentially even with an example of why this simplifies changes in the future or even reduces the diff of https://github.com/opencompl/ssa/pull/191? I think it is obvious to you, but being explicit here would help me.

AFAIU this pull out a large mechanical change from https://github.com/opencompl/ssa/pull/191, which is perfect as having noisy being 100% mechanical is typically a nice property for a PR.

alexkeizer:
> This looks like a good idea. Can you maybe write 1-2 paragraphs of context, potentially even with an example of why this simplifies changes in the future or even reduces the diff of #191? I think it is obvious to you, but being explicit here would help me.
> 
> AFAIU this pull out a large mechanical change from #191, which is perfect as having noisy being 100% mechanical is typically a nice property for a PR.

You beat me to the punch ;) I was writing a description, but got a bit distracted by #250.
Let's keep discussion of the broader context and trade-offs in #250, I opened this PR mostly because I was curious if anything would break.

SSA/Core/Framework.lean
```diff
@@ -15,19 +15,24 @@ open TyDenote (toType)
...
 abbrev RegionSignature Ty := List (Ctxt Ty × Ty)
 
 structure Signature (Ty : Type) where
   sig     : List Ty
   regSig  : RegionSignature Ty
   outTy   : Ty
 
-class OpSignature (Op : Type) (Ty : outParam (Type)) where
-  signature : Op → Signature Ty
+class OpSignature (d : Dialect) where
```
goens, line 30: Shouldn't this be `DialectSignature` then?

goens (approved):
LGTM (modulo one naming suggestion)

SSA/Core/Framework.lean
alexkeizer: I've also renamed `OpDenote` -> `DialectDenote`

bollu:
@alexkeizer I'm still a little hesitant, given the `simp [(X)]` thing we discovered. I'm happy to merge this, but I wish to note my lingering fear :P

SSA/Core/Tactic.lean
```diff
@@ -120,4 +143,5 @@ macro "simp_peephole" "[" ts: Lean.Parser.Tactic.simpLemma,* "]" "at" Γv:ident
 /-- `simp_peephole` with no extra user defined theorems. -/
 macro "simp_peephole" "at" Γv:ident : tactic => `(tactic| simp_peephole [] at $Γv)
 
+
```
bollu, line 146:
```suggestion
```

alexkeizer:
This now builds locally, so CI should be happy soon as well.

@goens could you give it another look, now that everything is ported over?

alexkeizer:
> @alexkeizer I'm still a little hesitant, given the `simp [(X)]` thing we discovered. I'm happy to merge this, but I wish to note my lingering fear :P

Yeah, some of my optimism has been converted into fear as well :sweat_smile:.
I don't really see an alternative way to reduce verbosity, though, so I'm still cautiously optimistic about going forward.

In particular, since `d : Dialect` is (reducibly) def-eq to `\<d.Op, d.Ty\>` we can always unbundle a particular problematic definition. For example, I defined `Scf.Op` as follows:
```lean
inductive Scf.Op (Op' Ty' : Type) [TyDenote Ty'] [DialectSignature ⟨Op', Ty'⟩] [DialectDenote ⟨Op', Ty'⟩] : Type _
  ...
```

SSA/Core/Framework/Dialect.lean
```diff
@@ -0,0 +1,17 @@
+/-
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
+import SSA.Core.ErasedContext
```
tobiasgrosser, line 4: Is this import needed?

tobiasgrosser (approved):
At least code noise wise this seems to go in the right direction.

SSA/Core/Tactic.lean
```diff
@@ -97,15 +97,38 @@ macro "simp_peephole" "[" ts: Lean.Parser.Tactic.simpLemma,* "]" "at" Γv:ident
...
+        use the *default* instance to determine whether a simp-lemma applies to the current goal.
+        Writing `simp [(X)]`, on the other hand, is equivalent to writing `simp [@X _ _ _]`
+          (for as many underscores as `X` takes arguments, implicit or explicit).
+        The parentheses seems to enable `simp` to unify typeclass arguments as well, and thus the
+          simp-lemma applies even for non-standard instances.
+
+        One caveat: `simp [(X)]` only works if `X` is a lemma, *not* if `X` is a definition to be
+        unfolded. Thus, we replace `HVector.denote` with its equation lemmas
+          `(HVector.denote_cons)` and `(HVector.denote_nil)` -/
+        (HVector.denote_cons), (HVector.denote_nil),
```
goens, line 131: wow, that's cursed! thanks for the thorough documentation :)

goens (approved):
lgtm!

SSA/Core/Framework/Dialect.lean
alexkeizer: Good call, it is not

PR #257, opened 2024-04-24
feat: do not leak the type definition in alive

tobiasgrosser:
This change ensures that simp_alive_ssa does not unfold the type definition of LLVM bitvectors, such that our semantics do not leak before we unfold them.


goens (commented):
lgtm

PR #258, opened 2024-04-24
chore: include InstCombine wrappers in build

tobiasgrosser:
And fix it with respect to the recent dialect introduction.


tobiasgrosser:
Merged as trivial.

Issue #169, opened 2024-01-26
Alive example mistranslation


alexkeizer:
The alive script generates the following example case that failed to elaborate:
```lean
def alive_AndOrXor_698_src  (w : Nat)   :=
[alive_icom ( w )| {
^bb0(%a : _, %b : _, %d : _):
  %v1 = "llvm.and" (%a,%b) : (_, _) -> (i1)
  %v2 = "llvm.and" (%a,%d) : (_, _) -> (i1)
  %v3 = "llvm.mlir.constant" () { value = 0 : i1 } :() -> (i1)
  %v4 = "llvm.icmp.eq" (%v1,%v3) : (_, _) -> (i1)
  %v5 = "llvm.mlir.constant" () { value = 0 : i1 } :() -> (i1)
  %v6 = "llvm.icmp.eq" (%v2,%v5) : (_, _) -> (i1)
  %v7 = "llvm.and" (%v4,%v6) : (i1, i1) -> (i1)
  "llvm.return" (%v7) : (i1) -> ()
}]

def alive_AndOrXor_698_tgt  (w : Nat)  :=
[alive_icom ( w )| {
^bb0(%a : _, %b : _, %d : _):
  %v1 = "llvm.or" (%b,%d) : (_, _) -> (i1)
  %v2 = "llvm.and" (%a,%v1) : (_, _) -> (i1)
  %v3 = "llvm.and" (%a,%b) : (_, _) -> (_)
  %v4 = "llvm.and" (%a,%d) : (_, _) -> (_)
  %v5 = "llvm.mlir.constant" () { value = 0 : _ } :() -> (_)
  %v6 = "llvm.icmp.eq" (%v3,%v5) : (_, _) -> (i1)
  %v7 = "llvm.mlir.constant" () { value = 0 : _ } :() -> (_)
  %v8 = "llvm.icmp.eq" (%v4,%v7) : (_, _) -> (i1)
  %v9 = "llvm.mlir.constant" () { value = 0 : i1 } :() -> (i1)
  %v10 = "llvm.icmp.eq" (%v2,%v9) : (_, _) -> (i1)
  "llvm.return" (%v10) : (i1) -> ()
}]
```

The issue is that the program claims a return type of `i1` where it really should be the meta-variable `_`.
That is, it should have generated:

```lean
def alive_AndOrXor_698_src  (w : Nat)   :=
[alive_icom ( w )| {
^bb0(%a : _, %b : _, %d : _):
[...]

PR #260, opened 2024-04-29
Re gen alive

goens:
This documents and re-generates the `AliveAutoGenerated.lean`.

Since the file has some manual intervention, it currently wont work well.


goens:
There's essentially two non-trivial issues with `AliveAutoGenerated.lean` and the current flow:
  1. There's a bug in the "unification" that we do in the translation that results in Issue #169. We can either:
     a. fix the bug or
     b. keep the manual edits somehow (e.g. a sed script?) or 
     c. stop using the flow to regenerate.
  2. In the alive generating script, commit https://github.com/opencompl/alive/commit/8563d42a changed the flow we used to generate the statements (and it now won't work), because it's generating a different format. We can:
    a. Revert the changes in that commit to allow us to re-generate the statements (and do it in two steps like we used to).
    b. Give up on the automatic statement generation and just do those by hand.
  Note that there's been a lot of manual commits on `AliveStatements.lean`, which was auto-generated by this flow. If we want to revive this part of the flow, we'll need to also manage those separately.

tobiasgrosser:
Thank you. This is great. It seems the AliveAutoGenerated.lean file has not massively changed. That is expected. I guess my objective was to replace `simp_alive_peephole` with `simp_alive_ssa` in `AliveAutoGenerated`and then re-generate the `AliveStatements.lean`. My hope would be that then AliveAutoGenerated is independent as well as the signatures in AliveStatements are independent of any changes to our semantics, which should make these files a lot more stable.

We could also commit the script to update AliveStatements.lean into this repo, if this would make updates easier and it is indeed independent of alive. Why did @bollu change the script in https://github.com/opencompl/alive/commit/8563d42a?

goens:
this is subsumed by #309

PR #264, opened 2024-04-30
feat: upstream alive_simplifyMulDivRem290

tobiasgrosser:
This one is again pulled out from https://github.com/opencompl/ssa/pull/167


bollu:
LGTM, I think it would be nicer to go via the more general lemma `(A <<< B).toNat = A.toNat * 2^B.toNat`, but I'm happy to leave it as-is.

PR #272, opened 2024-05-07
chore: move non-ssa related tactics into TacticAuto

tobiasgrosser:
This change will allow us to include only TacticAuto.lean, but not Tactic.lean into AliveStatement.lean. As a result, we do not include Com.Refinement and the `⊑` symbol does not expect Com terms, but can work on LLVM.IntW level.

This patch is in support of https://github.com/opencompl/ssa/pull/269 and https://github.com/opencompl/ssa/pull/260.


SSA/Projects/InstCombine/AliveStatements.lean
```diff
@@ -1,3 +1,4 @@
+
```
bollu, line 1:
```suggestion

```

Stray newline?

bollu:
LGTM modulo my comment.

SSA/Projects/InstCombine/AliveStatements.lean
tobiasgrosser: This file was added by accident. I dropped it now.

PR #275, opened 2024-05-07
feat: include SSA/Core.lean and add SSA/Core/MLIRSyntax.lean

bollu:
This cleans up our import story.


tobiasgrosser:
Merged as trivial.

PR #276, opened 2024-05-07
Optimize the kernels in AliveStatements

tobiasgrosser:
This gives us 48 out of 90 statements proven (42 sorry-ed).

### DIFF
--- a/SSA/Projects/InstCombine/TacticAuto.lean
+++ b/SSA/Projects/InstCombine/TacticAuto.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 import Mathlib.Tactic.Ring
 import Std.Data.BitVec
 import Mathlib.Data.BitVec.Lemmas
+import SSA.Projects.InstCombine.ForLean
 
 import SSA.Projects.InstCombine.LLVM.EDSL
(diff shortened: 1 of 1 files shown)


bollu:
@tobiasgrosser you seem to be adding `simp` again, do we not want `simp only [theorem1, ..., theoremN]` or `simp only [simp_set]`?

tobiasgrosser:
Let's start with the simps. If it becomes clearer what these actually are, we can factor this into a simp_set.

> @tobiasgrosser you seem to be adding `simp` again, do we not want `simp only [theorem1, ..., theoremN]` or `simp only [simp_set]`?

Is there a way to add `ofInt_eq_ofNat` via a simp set? I tried, but I cannot add the `()`. 😢

tobiasgrosser:
Otherwise, I feel `negOne_eq_allOnes : -1#w = allOnes w` is inconsistent in lean. We should either rewrite to allOnes or state our theorems as -1#w.

tobiasgrosser:
I feel in absence of new ideas of to cleanup this code (with respect to the questions above) this is ready to merge.

tobiasgrosser:
I think we should just move ahead with this. I don't think I want to add another zulip discussion cycle.

PR #279, opened 2024-05-07
chore: reduce imports

tobiasgrosser:
Avoid including 'Tactic.lean' as this pulls in Refinement.lean, which currently pollutes the refinement relation and prevents the use of ForLean.lean in AliveStatements.


tobiasgrosser:
Merging as trivial

PR #284, opened 2024-05-08
update to 4.8.0-rc4

tobiasgrosser:


bollu:
Lgtm, the increase in max heartbeats for the shift lemmas is interesting, and we should keep a close eye on this.

PR #288, opened 2024-05-09
feat: run update_alive_statements in parallel

tobiasgrosser:
This should speedup the CI a bit.


tobiasgrosser:
This doubles the CI performance of this step from 7m to 3m.

bollu:
LGTM.

PR #290, opened 2024-05-09
feat: make CI faster by merging all lake exe cache get calls.

bollu:
This should prevent the repeated 'building Mathlib' in the logs.


bollu:
Does not really help. These changes make the time to be (https://github.com/opencompl/ssa/actions/runs/9020267548/job/24785045534):

> succeeded 1 minute ago in 12m 23s


While the last commit on CI has actions running at (https://github.com/opencompl/ssa/actions/runs/9018352452/job/24778773153)

> succeeded 2 hours ago in 14m 41s


--- 

However, @tobiasgrosser , note that comparing with Lean's CI time is not apples to apples, since Lean CI does not depend on mathlib, nor does it build using lake.

tobiasgrosser:
Thank you for looking into this. I am not the biggest fan of merging all together. This makes performance analysis harder and the speedup we get is rather minor. In particular, we cannot get perf numbers for some of the individual compilation things. TBH, I would leave it like this for now until we find a way to really speed things up.

bollu:
I agree. I shall close the PR, since it gave me what I wanted: a roofline on the available performance from CI.

PR #291, opened 2024-05-09
chore: split CI in fast and slow

tobiasgrosser:


bollu:
LGTM

bollu:
Thanks for the split, I agree this makes a whole lot of sense.

PR #293, opened 2024-05-09
chore: enable fail on warning

bollu:
This forces us to fix all warnings.


bollu:
we need to figure out what to do with sorrys

tobiasgrosser:
We only have two sorries left in proofs that we may not need. Should we just drop them and replace them with `proofwanted`? This this PR can go in.

bollu:
@tobiasgrosser that's a good idea, would you mind making the change?

tobiasgrosser:
> @tobiasgrosser that's a good idea, would you mind making the change?

I replaced the first: https://github.com/opencompl/ssa/pull/363.

Not sure what about the second one. Should we comment that one?

tobiasgrosser:
I commented the last sorry. Let's see if that makes this PR green.

tobiasgrosser:
This did not make it for months. Let's close it for now.

PR #294, opened 2024-05-10
feat: theorems about Ctxt and MVar

tobiasgrosser:
These theorems will allow us to reason about the context and MVar with less unfolding about low-level theorems. While uses of these theorems are still worked on, the theorems seem to be generally useful so lets upstream them already to keep later patches small.

Co-Authored-By : Siddharth Bhat<siddu.druid@gmail.com>

These are factored out of https://github.com/opencompl/ssa/pull/225.

### DIFF
--- a/SSA/Core/ErasedContext.lean
+++ b/SSA/Core/ErasedContext.lean
@@ -34,7 +34,7 @@ def empty : Ctxt Ty := []
 instance : EmptyCollection (Ctxt Ty) := ⟨Ctxt.empty⟩
 instance : Inhabited (Ctxt Ty) := ⟨Ctxt.empty⟩
 
-@[simp] lemma empty_eq : (∅ : Ctxt Ty) = [] := rfl
+lemma empty_eq : (∅ : Ctxt Ty) = [] := rfl
 
 @[match_pattern]
 def snoc : Ctxt Ty → Ty → Ctxt Ty :=
@@ -55,6 +55,16 @@ def get? : Ctxt Ty → Nat → Option Ty :=
 /-- Map a function from one type universe to another over a context -/
 def map (f : Ty₁ → Ty₂) : Ctxt Ty₁ → Ctxt Ty₂ :=
   List.map f
+
+@[simp]
+lemma map_nil (f : Ty → Ty') :
+  map f ∅ = ∅ := rfl
+
+@[simp]
+lemma map_cons (Γ : Ctxt Ty) (t : Ty) (f : Ty → Ty') :
+  map f (Γ.cons t) = (Γ.map f).cons (f t) := rfl
+
+theorem map_snoc (Γ : Ctxt Ty) : (Γ.snoc a).map f = (Γ.map f).snoc (f a) := rfl
 
 instance : Functor Ctxt where
   map := map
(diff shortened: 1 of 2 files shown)


bollu:
@tobiasgrosser LGTM, and to stick to @goens ' principles, I add myself to the  `Co-Authored-By` set.

SSA/Core/Util/ConcreteOrMVar.lean
```diff
@@ -26,6 +26,12 @@ instance : OfNat (ConcreteOrMVar Nat φ) n := ⟨.concrete n⟩
 
 namespace ConcreteOrMVar
 
+/- Terse notation for concrete and mvar. -/
+namespace Notation
+scoped notation "i" n => ConcreteOrMVar.concrete n
+scoped notation "i?" f => ConcreteOrMVar.mvar f
```
bollu, line 32: I'm not totally sold on the value of these notations, though I write them myself. I wanted to mimic LLVM syntax, so it looks like `i 1` and `i? 0` to stand for concrete numbers and mvars, but I'm not a huge fan. @alexkeizer 's second opinion is much appreciated.

bollu:
LGTM from my end.

SSA/Core/Util/ConcreteOrMVar.lean
tobiasgrosser: I dropped the notation for now, given that you are uncertain here.

tobiasgrosser:
> @tobiasgrosser LGTM, and to stick to @goens ' principles, I add myself to the `Co-Authored-By` set.

I feel this needs to actually be part of the commit. I force pushed a commit with the correct tag and will push now. Thank you for the reminder.

SSA/Core/Util/ConcreteOrMVar.lean
```diff
@@ -48,5 +48,22 @@ def instantiate (as : Vector α φ) : ConcreteOrMVar α φ → α
...
+  simp [instantiate]
+  simp [Vector.get]
+
+@[simp]
+lemma instantiate_mvar_zero' :
+    (mvar (φ := 1) ⟨0, by simp⟩).instantiate (Subtype.mk [w] (by simp)) = w := rfl
+
+@[simp]
+lemma instantiate_mvar_zero'' :
+    (mvar (φ := 1) 0).instantiate (Subtype.mk [w] h1) = w := rfl
```
alexkeizer, line 67: The difference between these three lemmas is not very clear to me, and obfuscated by different spellings of identical concepts (i.e., `ConcreterOrMVar.instantiate` vs dot-notation `(...).instantiate` and `Subtype.mk` vs magic braces `⟨_, _⟩`). I'd like to see this be more uniform, and I'd like to see a docstring to explain the differences

SSA/Core/ErasedContext.lean
```diff
@@ -56,6 +56,16 @@ def get? : Ctxt Ty → Nat → Option Ty :=
 def map (f : Ty₁ → Ty₂) : Ctxt Ty₁ → Ctxt Ty₂ :=
   List.map f
 
+@[simp]
+lemma map_nil (f : Ty → Ty') :
+  map f ∅ = ∅ := rfl
+
+@[simp]
+lemma map_cons (Γ : Ctxt Ty) (t : Ty) (f : Ty → Ty') :
+  map f (Γ.cons t) = (Γ.map f).cons (f t) := rfl
```
alexkeizer, line 65: There's some inconsistent use of dot-notation here (`map f _` vs `_.map f`).

Also, I think `cons` here is referring to `List.cons`, which we've alliased as `Ctxt.snoc`.
In the spirit of remaining higher-level, I think we should not be adding lemmas for `List.snoc`, to nudge us towards not unfolding `Ctxt.snoc` in the first place

alexkeizer (commented):
I missed this PR before it was merged, but still wanted to share some thoughts

PR #301, opened 2024-05-13
core: make AliveStatements.lean robust against multiple goals

tobiasgrosser:
https://github.com/opencompl/ssa/pull/225 will fix three statements, one of which will introduce multiple goals in `AliveStatements.lean`. This PR makes `AliveStatements.lean` robust against multiple worries by using `all_goals`.

Thank you @alexkeizer for proposing this.


bollu:
LGTM, cute use of `all_goals` to handle both the "no goals" and the "multiple goals" case.

PR #302, opened 2024-05-13
chore: disable unreachable tactic warning in AliveStatements

tobiasgrosser:


bollu:
LGTM, though I'm not a fan of the global disable. @tobiasgrosser can you confirm that this does not "leak into" any module that imports this one? Ie, setting it here does not disable unreachable checking for all other modules that import this one?

tobiasgrosser:
Confirmed. It does not leak.

PR #305, opened 2024-05-14
chore: express lemma in terms of '+' instead of .add

tobiasgrosser:
This ensures that the theorem solves our theorem in AliveStatements.lean. This brings us from 50 to 51 proven lemma.


eurquhart1 (approved):
looks good to me

PR #309, opened 2024-05-17
chore: document alive generation process

goens:
This takes (and fixes) the documentation changes from #260 . It documents how to (re)generate the alive rewrites and the corresponding statements.


bollu:
LGTM, thanks for the documentation andres!

PR #310, opened 2024-05-19
chore: use functional induction principle

tobiasgrosser:


SSA/Core/Framework.lean
```diff
@@ -671,8 +671,9 @@ section Lemmas
 @[simp] lemma Com.denoteLets_ret : (.ret v : Com d Γ eff t).denoteLets = fun V => pure V := by
   funext V; simp [denoteLets]
 
-theorem Com.denoteLets_eq {com : Com d Γ eff t} : com.denoteLets = com.toLets.denote := by
-  simp only [toLets]; induction com using Com.rec' <;> simp [Lets.denote_lete]
+theorem Com.denoteLets_eq {com : Com d Γ eff t} : com.denoteLets Γv = com.toLets.denote Γv := by
+  induction Γ, eff, t, com, Γv using Com.denoteLets.induct
+  <;> simp only [toLets] <;> simp [Lets.denote_lete]
```
alexkeizer, line 676: I'd say that in this case, the original proof is actually nicer than the functional induction proof, which makes sense given that the functions involved are all perfectly structurally recursive.

bollu:
I leave this to @alexkeizer 's executive decision.

tobiasgrosser:
Let's keep it the way it is. This was mostly an experiment to see if I could find code that became easier.

PR #330, opened 2024-05-22
Remove reduce in LLVM eDSL

tobiasgrosser:
This change will reduce our dependency on leans unfolding in the reduction. It might also speed up compilation times.


tobiasgrosser:
This has been resolved in https://github.com/opencompl/ssa/pull/356.

PR #333, opened 2024-05-23
chore: build TestLLVMOps

tobiasgrosser:
While at it we also improve the formatting. We also enable a test case for "llvm.not". Even though LLVM does not support it, our parser does so we should also test this functionality. In the future we may want to remove it.


tobiasgrosser:
This one is already tested via the LLVM Enumerator.

PR #336, opened 2024-05-23
chore: fix tests

tobiasgrosser:
This fixed https://github.com/opencompl/ssa/issues/334.

### DIFF
--- a/SSA/Projects/InstCombine/Test.lean
+++ b/SSA/Projects/InstCombine/Test.lean
@@ -48,7 +48,8 @@ def bb0 : Region 0 := [mlir_region|
 
 
 open InstCombine
-def Γn (n : Nat) : Context φ :=
+open InstcombineTransformDialect
+def Γn (n : Nat) : Ctxt (MetaLLVM φ).Ty :=
   Ctxt.ofList <| .replicate n (.bitvec 32)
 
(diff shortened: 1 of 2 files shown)


bollu:
LGTM now.

SSA/Projects/InstCombine/Test.lean
```diff
@@ -87,19 +88,14 @@ def ops' := [op0, op1, op2, op3, op4]
...
-#eval (mkCom bb0).toOption
+def com := mkCom (d := InstCombine.MetaLLVM 0) bb0 |>.toOption |>.get (by rfl)
 
-
-#check ExceptT.run
-
-def com := mkCom bb0 |>.toOption |>.get (by rfl)
-
-#eval com
+#reduce com
```
alexkeizer, line 93: I think we should either drop the `#reduce` here altogether, or put it behind a `#guard_msgs` line to make sure whatever the `#reduce` is checking is automatically tested (and not polluting the output of `lake build`)

tobiasgrosser: This has been fixed at: https://github.com/opencompl/ssa/pull/337. I had planned to resolve this in this PR, but we merged it before doing so.

PR #337, opened 2024-05-23
chore: use guard_msgs to avoid info noise

tobiasgrosser:


SSA/Projects/InstCombine/Test.lean
```diff
@@ -64,12 +86,42 @@ def opRet : Op 0 := [mlir_op| "llvm.return"(%4) : (i32) -> ()]
...
+info: Except.ok ⟨EffectKind.pure, ⟨i32, InstCombine.MOp.binary
+    (ConcreteOrMVar.concrete 32)
+    (InstCombine.MOp.BinaryOp.and)[[%0, ,, %2]]⟩⟩
+-/
+#guard_msgs in #eval mkExpr    (Γn 4) op3    ["2", "1", "0", "arg0"]
+
+/--
+info: Except.ok ⟨EffectKind.pure, ⟨i32, InstCombine.MOp.binary
+    (ConcreteOrMVar.concrete 32)
+    (InstCombine.MOp.BinaryOp.add)[[%0, ,, %1]]⟩⟩
```
tobiasgrosser, line 117: Interestingly, here the printer seems to be broken as it yields a `, ,,`. I suggest to commit the broken guard_msg and to open an issue for this.

PR #339, opened 2024-05-24
chore: add simpler tests that compare our macro syntax with the com t…

tobiasgrosser:
…ests

These tests are useful to debug issues that may arise due to our macros not reducing the way they should.


tobiasgrosser:
This PR will facilitate the debugging of https://github.com/opencompl/ssa/pull/330.

bollu:
LGTM

PR #347, opened 2024-05-26
chore: do not make Ctxt.Var.toMap a coercion

tobiasgrosser:
Before this change toSnoc and toMap could not be distinguished visually, which made debugging unnecessarily hard. Also, we really want to eliminate toMap to ensure that all meta-abstractions are liminated before we try to remove the SSA semantics.


bollu:
This is fine by me, so LGTM. I want a post-merge review from @alexkeizer in case this was added for some reason I don't know / remember.

alexkeizer:
I think it's fine to remove. It was added as a coercion because it represents the straightforward way of using a variable `v : Var \G t` where a variable of context `f <$> \G` is expected, so the system might as well do the transformation for you.

I agree that it's unfortunate that this obfuscates the proof goals, and explicitly writing `toMap` is not that much added effort.

PR #351, opened 2024-05-27
chore: put all simps on a new line

tobiasgrosser:


bollu:
I disagree with this change. `:= by simp` is very standard lean convention:

```
bollu@user-117-162 ~/2/e/l/l/src (getLsb_sshiftRight)> rg ":= by simp" | wc -l
     153
bollu@user-117-162 ~/2/e/l/l/src (getLsb_sshiftRight)> rg ":= by simp" | head
Init/PropLemmas.lean:theorem or_rotate : a ∨ b ∨ c ↔ b ∨ c ∨ a := by simp only [or_left_comm, Or.comm]
Init/PropLemmas.lean:@[simp] theorem forall_eq' {a' : α} : (∀ a, a' = a → p a) ↔ p a' := by simp [@eq_comm _ a']
Init/PropLemmas.lean:@[simp] theorem exists_and_right : (∃ x, p x ∧ b) ↔ (∃ x, p x) ∧ b := by simp [And.comm]
Init/PropLemmas.lean:@[simp] theorem exists_eq_left' : (∃ a, a' = a ∧ p a) ↔ p a' := by simp [@eq_comm _ a']
Init/PropLemmas.lean:    (∀ b a, f a = b → p b) ↔ ∀ a, p (f a) := by simp [forall_comm]
Init/PropLemmas.lean:    (∀ b a, b = f a → p b) ↔ ∀ a, p (f a) := by simp [forall_comm]
Init/PropLemmas.lean:theorem decide_eq_true_iff (p : Prop) [Decidable p] : (decide p = true) ↔ p := by simp
Init/Grind/Norm.lean:@[grind_norm] theorem eq_true_eq (p : Prop) : (p = True) = p := by simp
Init/Grind/Norm.lean:@[grind_norm] theorem eq_false_eq (p : Prop) : (p = False) = ¬p := by simp
Init/Grind/Norm.lean:@[grind_norm↓] theorem not_forall (p : α → Prop) : (¬∀ x, p x) = ∃ x, ¬p x := by simp
```

bollu:
Aha, I see, this was done due to @alexkeizer 's comments asking for consistency at https://github.com/opencompl/ssa/pull/350#discussion_r1616180639.

OK, I'm happy to merge this for consistency *in this file*, but I want to write on the record that I prefer ` := by simp` on the same line as the statement.

alexkeizer:
My rule of thumb is that having `by simp` on the same line is preferred if the theorem/lemma is simple enough that it all fits on one line. But, once we start using a second line to put the type/statement on, it tends to be complicated enough that it's best to put the proof on its own line as well to visually distinguish it.

I don't think this is universally agreed up-on, though. I think I've seen mathlib theorems that span multiple lines but don't have their proof in its own line

bollu:
> My rule of thumb is that having by simp on the same line is preferred if the theorem/lemma is simple enough that it all fits on one line. But, once we start using a second line to put the type/statement on, it tends to be complicated enough that it's best to put the proof on its own line as well to visually distinguish it.



I agree with this. However, the particular change made here is a simple theorem whose entire proof is `by simp`, so I think it's fair to keep it all in one line.

alexkeizer:
It's not, though? Put on one line, the theorem would be
```lean
def three_inst_stmt (e : LLVM.IntW w) : @BitVec.Refinement (BitVec w) (LLVM.not (LLVM.not (LLVM.not e))) (LLVM.not (LLVM.not (LLVM.not e))) := by simp
```
which is about 150 columns wide.

My point is that I don't like theorems which take multiple lines, but don't have a dedicated line for their proof. In this case, even the statement itself already needs multiple lines if we limit ourselves to 100 columns per line, as we generally do.
I'm saying I explicitly *don't* like short proofs  being on the same line as a complicated, multi-line theorem statement, as in:
```lean
def three_inst_stmt (e : LLVM.IntW w) : 
  @BitVec.Refinement (BitVec w) (LLVM.not (LLVM.not (LLVM.not e))) 
    (LLVM.not (LLVM.not (LLVM.not e))) := by simp
```

I much prefer bumping the proof to the next line, even if the proof itself is simple. 
in:
```lean
def three_inst_stmt (e : LLVM.IntW w) : 
    @BitVec.Refinement (BitVec w) (LLVM.not (LLVM.not (LLVM.not e))) (LLVM.not (LLVM.not (LLVM.not e))) := by 
  simp
```
I find this helps me visually distinguish when a `rfl` or `by simp` is part of the theorem statement (e.g., when subtypes are involved), or when it's a trailing proof.

PR #356, opened 2024-05-29
chore: disable reduction in LLVM eDSL

tobiasgrosser:
This change is needed to update to lean 4.9. We commit this as a separate change, such that for future debugging sessions there is a commit where easily both versions can be inspected.

### DIFF
--- a/SSA/Projects/InstCombine/ForLean.lean
+++ b/SSA/Projects/InstCombine/ForLean.lean
@@ -84,6 +84,7 @@ lemma one_shiftLeft_mul_eq_shiftLeft {A B : BitVec w} :
   simp only [bv_toNat, Nat.mod_mul_mod, one_mul]
   rw [Nat.mul_comm]
 
+@[simp]
 def ofInt_zero_eq (w : Nat) : BitVec.ofInt w 0 = 0#w := rfl
 def ofNat_zero_eq (w : Nat) : BitVec.ofNat w 0 = 0#w := rfl
 def toInt_zero_eq (w : Nat) : BitVec.toInt 0#w = 0 := by
@@ -272,8 +273,7 @@ def one_sdiv { w : Nat} {a : BitVec w} (ha0 : a ≠ 0) (ha1 : a ≠ 1)
       unfold BitVec.sdiv
       simp [udiv_one_eq_self, msb_one, ofInt_one_eq_ofNat_one]
       by_cases h : BitVec.msb a <;> simp [h]
-      · simp [BitVec.ofInt_zero_eq]
-        simp only [neg_eq_iff_eq_neg, BitVec.neg_zero]
+      · simp only [neg_eq_iff_eq_neg, BitVec.neg_zero]
         apply BitVec.udiv_one_eq_zero
         apply BitVec.gt_one_of_neq_0_neq_1
         · rw [neg_neq_iff_neq_neg]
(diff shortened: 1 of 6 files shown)


bollu:
I'm happy to enable this once the CI goes green.

SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -249,7 +249,6 @@ elab "[alive_icom (" mvars:term,* ")| " reg:mlir_region "]" : term => do
     if ssa.alive_icom_reduce.get (← getOptions)
     then reduce com
```
bollu, line 250:
```suggestion
    then do 
      logWarningAt reg "simplifying proof state with reduction is deprecated, as it unstable across Lean versions. Recall that reduction in Lean is slow and heuristic based, and the heuristics can change across Lean versions. Suggestion: Find appropriate simp lemmas to simplify the proof state."
      reduce com
```

alexkeizer:
I think this is fine to merge, but I'm a bit scared of the implications. Presumably our use of `reduce` in the AST -> ICom transformations is similarly fragile. However, if we did the same there that means we'd have to think about the appropriate simp sets to reduce the `TransformDialect` machinery for every new dialect. This changes our story from "You only have to implement a few typeclasses to get pretty syntax" to "You have to implement a few typeclasses, and then spent a non-trivial amount of time engineering a simp-set to ensure it reduces properly".

For now, it's fine to just let the `LLVM` dialect be a bit of a special case, but if we start running into issues with the other `reduce`, I would strongly suggest we revisit the decision to do all of this in object code. If meta variables and AST transformations were happening on the level of Lean `Expr`s, we wouldn't have to deal with any of these simp-sets or reduction problems (and I strongly suspect it'd be significantly faster, too)

SSA/Projects/InstCombine/ForLean.lean
```diff
@@ -84,7 +84,12 @@ lemma one_shiftLeft_mul_eq_shiftLeft {A B : BitVec w} :
   simp only [bv_toNat, Nat.mod_mul_mod, one_mul]
   rw [Nat.mul_comm]
 
+@[simp]
 def ofInt_zero_eq (w : Nat) : BitVec.ofInt w 0 = 0#w := rfl
+
+@[simp]
+def ofInt_one_eq (w : Nat) : BitVec.ofInt w 1 = 1#w := rfl
```
alexkeizer, line 91:
```suggestion
@[simp]
lemma ofInt_one_eq (w : Nat) : BitVec.ofInt w 1 = 1#w := rfl
```
It should be `lemma` or `theorem`, but `def` is definitely wrong.

This should probably be a generic lemma about `OfNat.ofNat`, e.g., 
```lean
@[simp] lemma ofInt_ofNat (w n : Nat) : BitVec.ofInt w (OfNat.ofNat n) = n#w := by sorry
```

SSA/Projects/InstCombine/ForLean.lean
```diff
@@ -282,7 +293,7 @@ def one_sdiv { w : Nat} {a : BitVec w} (ha0 : a ≠ 0) (ha1 : a ≠ 1)
         · rw [neg_neq_iff_neq_neg]
           rw [← negOne_eq_allOnes] at hao
           assumption
-      · simp [BitVec.ofInt_zero_eq]
+      ·
         apply BitVec.udiv_one_eq_zero
```
alexkeizer, line 297: whitespace
```suggestion
      · apply BitVec.udiv_one_eq_zero
```

SSA/Projects/InstCombine/ForLean.lean
tobiasgrosser: I dropped this change.

PR #358, opened 2024-05-29
chore: canonicalize constants

tobiasgrosser:


tobiasgrosser:
I pulled this PR out of https://github.com/opencompl/ssa/pull/356, as this seems to be a self-contained change that can be verified rather mechanically. In https://github.com/opencompl/ssa/pull/356, I made some of these statements new simp-lemma on BitVec which caused unrelated changes in BitVec.

tobiasgrosser:
> Alive Statistics: 30 / 93 (63 failed)
Nice. These still need to be fixed.

tobiasgrosser:
Nice. This is now ready to review (and merge). It should reduce a lot of the noise we saw in https://github.com/opencompl/ssa/pull/356.

SSA/Projects/InstCombine/ForStd.lean
```diff
@@ -131,4 +131,11 @@ theorem bitvec_minus_one : BitVec.ofInt w (Int.negSucc 0) = (-1 : _root_.BitVec
     rw [zeroBitwidth]
     simp
 
+theorem bitvec_minus_one' : BitVec.ofInt w (-1) = (-1 : BitVec w) := by
+  have x : -1 = Int.negSucc 0 := by simp
+  rw [x, bitvec_minus_one]
+
+
+
+
```
bollu, line 140:
```suggestion
theorem bitvec_minus_one' : BitVec.ofInt w (-1) = (-1 : BitVec w) := by
  have x : -1 = Int.negSucc 0 := by simp
  rw [x, bitvec_minus_one]
  
```

SSA/Projects/InstCombine/ForLean.lean
```diff
@@ -299,22 +298,18 @@ def sdiv_one_one' (h : 1 < w) : BitVec.sdiv 1#w 1#w = 1#w := by
   apply BitVec.eq_of_toNat_eq
   simp [hone]
 
-def sdiv_one_one : BitVec.sdiv (BitVec.ofInt w 1) 1 = 1 := by
+def sdiv_one_one : BitVec.sdiv 1#w 1 = 1 := by
```
bollu, line 301: This is not in simp-normal-form for sure, right? It's depending on the `1 : Nat` to get coerced into `1#w : BitVec w`? A good way to check that the theorem statement is in simp normal form is to write 

```suggestion
def sdiv_one_one : BitVec.sdiv 1#w 1 = 1 := by simp
```

and see if `simp` complains about the theorem statement being in normal form. I would imagine that the canonical theorem statement is

```lean
def sdiv_one_one : BitVec.sdiv 1#w 1#w = 1#w := by simp
```

SSA/Projects/InstCombine/ForLean.lean
```diff
@@ -123,9 +123,10 @@ lemma ofInt_ofNat (w n : Nat) :
     BitVec.ofInt w (OfNat.ofNat n) = BitVec.ofNat w n :=
   rfl
 
+lemma ofInt_ofNat' : BitVec.ofInt w (OfNat.ofNat (α := ℤ) x ) = x#w := rfl
```
bollu, line 126: Giving types here should just work:

```suggestion
lemma ofInt_ofNat' {x : Nat} : BitVec.ofInt w (OfNat.ofNat x) = x#w := rfl
```

I am once again suspicious of this, though. Doesn't the `OfNat` simp away?

bollu (changes requested):
I think the theorem statements need to be in `simp` normal form before we can merge this. If they are not being normalized correctly, then we ought to write the correct simps to go between `Int/Nat/OfNat.ofNat`

SSA/Projects/InstCombine/ForStd.lean
tobiasgrosser: @bollu, what is the change you are suggesting here?

bollu: @tobiasgrosser aha,  I wanted to nuke the extra newlines, but it seems like github's "suggestion" algorithm strips trailing newline, so it looks like I'm suggesting we add a newline..

SSA/Projects/InstCombine/ForLean.lean
tobiasgrosser: Resolved.

SSA/Projects/InstCombine/ForStd.lean
tobiasgrosser: Done.

SSA/Projects/InstCombine/ForLean.lean
tobiasgrosser: We already have `ofInt_ofNat`, but it does not apply. The problem is here:

```
+++ b/SSA/Projects/InstCombine/AliveHandwrittenLargeExamples.lean
@@ -232,6 +232,8 @@ def alive_simplifyMulDivRem805' (w : Nat) :
   simp
   split_ifs with c
   simp
+  -- TODO: This one does not work: simp [(BitVec.ofInt_ofNat)]
+  simp [(BitVec.ofInt_ofNat')]
```

alexkeizer: I'm very confused, `n#w` is a macro that expands to `BitVec.ofNat`, so `ofInt_ofNat` and `ofInt_ofNat'` should be the same theorem (as in, syntactically equal statements, post elaboration)

bollu: The upstream issue tracking this is https://github.com/leanprover/lean4/issues/2867

bollu:
Lgtm

bollu (approved):

SSA/Projects/InstCombine/ForLean.lean
```diff
@@ -120,9 +120,11 @@ theorem Nat.one_mod_two_pow_succ_eq {n : Nat} : 1 % 2 ^ n.succ = 1 := by
 
 @[simp]
 lemma ofInt_ofNat (w n : Nat) :
-    BitVec.ofInt w (OfNat.ofNat n) = BitVec.ofNat w n :=
+    BitVec.ofInt w (no_index (OfNat.ofNat n)) = BitVec.ofNat w n :=
   rfl
 
+lemma ofInt_ofNat' : BitVec.ofInt w (OfNat.ofNat (α := ℤ) x ) = x#w := rfl
+
 -- @[simp]
```
alexkeizer, line 128: It seems you didn't remove the primed version, but this one should not be needed after the `no_index` fix, right?

tobiasgrosser: Right. Sorry for this and thank you for the post-commit review: https://github.com/opencompl/ssa/pull/new/clean_unneeded_stmt.

PR #359, opened 2024-05-30
feat: LLVM pretty syntax based on macro elaboration

bollu:
This revives our old way of building custom syntax, by elaborating pretty syntax into generic syntax. Thanks @alexkeizer and @tobiasgrosser for the elbow grease :)

PR #361, opened 2024-05-30
Fix: Make the generic MLIR parser hygiene resistant

alexkeizer:
This changes various parts of the MLIR generic parser to inspect the `rawVal` of an identifier, rather than the `val`. This raw value is literally the substring of the input value that corresponds to the identifier, and is thus not affected by hygiene.

This becomes especially relevant if we want to generate MLIR syntax with macros, since we don't want our types and attributes, which use identifiers in the parser, to be affected by hygiene. In particular, this patch removes the need for a `set_option hygiene false` in #359

### DIFF
--- a/SSA/Core/MLIRSyntax/GenericParser.lean
+++ b/SSA/Core/MLIRSyntax/GenericParser.lean
@@ -665,27 +665,24 @@ attrVal10Float
 # MLIR ATTRIBUTE
 -/
 
-declare_syntax_cat mlir_attr_entry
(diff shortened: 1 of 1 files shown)


tobiasgrosser (approved):
I am not an expert on this code, but from my example it looks great.

bollu:
Sweet, thanks @alexkeizer !

alexkeizer:
I messed up a bit when opening this PR, and apparently we've now merged the changes here into the `new_syntax2` branch, rather than the main branch. I'll fix it

PR #364, opened 2024-05-31
refactor: downgrade MLIR generic parser from syntax categories to closed syntax

alexkeizer:
Currently almost all parts of the generic MLIR parser uses syntax categories, meaning these parts of the (generic!) grammar are in theory extensible.

In theory this can be very convenient: if a dialect wanted to customize *only* their operands it could choose to hook into just that part of the parser. In practice, I would expect that even the low level elements (such as `mlir_op_operand`) being extensible is an anti-pattern:
* Such extensions would not be scoped to just that dialect, it would affect parsing of *all* generic MLIR syntax
* The parser doesn't really seem to be written for such extensions anyway. I tried to come up with an example of extending just the operands, but the most obvious ways that came to mind didn't actually work (```macro "&" n:num : mlir_op_operand => `(mlir_op_operand| % $n:num)```).
* Controlling which parts of the grammar are extensible would make it easier to eventually write an elaborator that outputs a meta-level AST, instead of a Lean expression of type `AST`,  if we eventually decide to go that way.

Thus, I propose we keep `mlir_op` (and potentially `mlir_type`) as the top-level syntax category that can be hooked into, but make low-level elements of the grammar closed. This PR is still a draft, but it show-cases what this change would look like for the case of `mlir_op_operand`.

This matches Lean itself, where `term` and `command` are extensible, but things like `def` are closed syntax.
Also note that this is exactly the level at which the current LLVM pretty syntax operates: that only extends `mlir_op` and would still work as-is after this refactor.

### DIFF
--- a/SSA/Core/MLIRSyntax/EDSL.lean
+++ b/SSA/Core/MLIRSyntax/EDSL.lean
@@ -15,6 +15,7 @@ namespace SSA
 
 open Qq Lean Meta Elab Term
 open MLIR.AST
+open MLIR.EDSL
 
 /-- `ctxtNf` reduces an expression of type `Ctxt _` to something in between whnf and normal form.
 `ctxtNf` recursively calls `whnf` on the tail of the list, so that the result is of the form
@@ -57,7 +58,7 @@ elab "[foo_com| " reg:mlir_region "]" : term => SSA.elabIntoCom reg q(FooOp)
 --     ^^^^^^^                                                        ^^^^^
 ```
 -/
-def elabIntoCom (region : TSyntax `mlir_region) (d : Q(Dialect)) {φ : Q(Nat)}
+def elabIntoCom (region : TSyntax ``mlir_region) (d : Q(Dialect)) {φ : Q(Nat)}
     (_dialectSignature : Q(DialectSignature $d) := by exact q(by infer_instance))
     (_transformTy      : Q(TransformTy $d $φ)     := by exact q(by infer_instance))
     (_transformExpr    : Q(TransformExpr $d $φ)   := by exact q(by infer_instance))
--- a/SSA/Core/MLIRSyntax/EDSL.lean
+++ b/SSA/Core/MLIRSyntax/EDSL.lean
@@ -58,7 +58,7 @@ elab "[foo_com| " reg:mlir_region "]" : term => SSA.elabIntoCom reg q(FooOp)
 --     ^^^^^^^                                                        ^^^^^
 ```
 -/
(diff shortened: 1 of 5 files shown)


alexkeizer:
@goens @bollu @tobiasgrosser I'm curious what you think about this refactor?

tobiasgrosser:
I think this is a great idea.

bollu:
@alexkeizer thanks for thinking about this, I'm generally happy with this change in terms of API surface area.

I don't quite understand:

>  Controlling which parts of the grammar are extensible would make it easier to eventually write an elaborator that outputs a meta-level AST, instead of a Lean expression of type AST, if we eventually decide to go that way.

Could you elaborate on this? Is it that each syntax category is a construction for our meta-level-AST, and having many constructors makes this unmanagable?

alexkeizer:
> @alexkeizer thanks for thinking about this, I'm generally happy with this change in terms of API surface area.
> 
> I don't quite understand:
> 
> > Controlling which parts of the grammar are extensible would make it easier to eventually write an elaborator that outputs a meta-level AST, instead of a Lean expression of type AST, if we eventually decide to go that way.
> 
> Could you elaborate on this? Is it that each syntax category is a construction for our meta-level-AST, and having many constructors makes this unmanagable?

So what I'm thinking about is having a function like ```TSyntax `mlir_op_operand -> MetaM SSAVal ```. If `mlir_op_operand` is closed, we can just match on exactly the syntax we define, and nothing else is possible. If it's a category, though, we're promising the syntax is extensible, we would then have to somehow hook into macro expansion to make that promise actually work.
Of course, we could just say that, even though it's a category, it's not *actually* intended to be extensible and throw some kind of error if we encounter unexpected syntax, but I'd rather codify this.

goens:
oh that's pretty cool! I didn't know about this difference (between closed syntax and extensible syntax categories).

I think it makes a lot of sense, thanks for the very thorough explanation @alexkeizer!

About the meta level ast, I also don't quite understand the point. Even if we choose to do things at elaboration time, what's the drawback of using our (current) AST data structure?

Or do you mean just that we can do (full) pattern matching on the closed syntax categories, so it's about the `TSyntax mlirsoandso` types? When talking about these self-made compilers from custom syntax, Mac once asked us why we wanted our own data structure and not just used `TSyntax`. Maybe using closed syntax is what we were missing for that, if I'm understanding you correctly

tobiasgrosser:
@alexkeizer, let me know if you would like to revive this.


No update since a while. Let's close this to keep our PRs clean.

PR #367, opened 2024-06-03
feat: support leaving out type annotations in LLVM's pretty EDSL

alexkeizer:
The diff is a bit larger than I'd like, because I had to move the `macro`s back to `macro_rules`, since the former doesn't handle optional syntax particularly well (see [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Pass.20optional.20syntax.20in.20macro/near/442254013))


tobiasgrosser (approved):
Very pretty, indeed!

PR #368, opened 2024-06-03
refactor: put all EDSL tests in sections

alexkeizer:
A simple cleanup of the GenericParser file:
* Make all test definitions `private`, so they don't get leaked
* Put all those definitions in (collapsible!) `Test` sections throughout the file
* Group tests a bit more together, to make them easier to distinguish from functionanilty


bollu:
LGTM, thanks for the cleanup!

PR #369, opened 2024-06-04
chore: update alive to get its output close to the AliveAutogenerated

tobiasgrosser:
For now, we keep the AliveAutoGenerated.lean file that we evolved manually, but the delta to the alive output is small and contains only:

- a couple of flipped arguments
- Alive generates `: _` instead of no types
- Alive dropped accidentally some `w`
- Alives unification is broken as it sometimes introduces i1 where it should not


tobiasgrosser:
@alexkeizer, this PR is broken as the use of `-1` in a constant expression cannot be parsed.

SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
```diff
@@ -23,6 +23,16 @@ syntax "llvm.sub"     : InstCombine.bin_op_name
...
+syntax "llvm.icmp.eq"     : InstCombine.bin_op_name
+syntax "llvm.icmp.ne"     : InstCombine.bin_op_name
+syntax "llvm.icmp.slt"     : InstCombine.bin_op_name
+syntax "llvm.icmp.sle"     : InstCombine.bin_op_name
+syntax "llvm.icmp.sgt"     : InstCombine.bin_op_name
+syntax "llvm.icmp.sge"     : InstCombine.bin_op_name
+syntax "llvm.icmp.ult"     : InstCombine.bin_op_name
+syntax "llvm.icmp.ule"     : InstCombine.bin_op_name
+syntax "llvm.icmp.ugt"     : InstCombine.bin_op_name
+syntax "llvm.icmp.uge"     : InstCombine.bin_op_name
```
alexkeizer, line 35: I suspect this is broken, since the `bin_op` macro generates `($t, $t) -> $t` as the type, but `icmp` is `($t, $t) -> i1`.
Hence the comment two lines below about `icmp` probably needing its own macro

alexkeizer:
I took the liberty of merging #370 into this branch, which fixes the -1 issue. 

Surprisingly some of the `icmp` test cases do work, but others indeed fail like I expected.
Similarly, we should still add syntax for select

alexkeizer:
Ah, the reason some `icmp` work but others break is because some of the type ascriptions, in particular the wrong return type ascriptions currently generated for icmp, are flat-out being ignored. (this explains why some cases work when they really shouldn't)

Similarly, in a constant, only the type ascription on the `value` attribute seems to matter, with the return type being ignored. Tobias' search/replace seems to have taken this previously return type for the concise syntax, exposing the test cases where this return type was actually wrongly annotated

README.md
```diff
@@ -1,3 +1,4 @@
+
 ![Core library build status](https://github.com/opencompl/ssa/actions/workflows/ci.yml/badge.svg)
```
tobiasgrosser, line 2:
```suggestion
![Core library build status](https://github.com/opencompl/ssa/actions/workflows/ci.yml/badge.svg)
```

SSA/Projects/InstCombine/AliveAutoGenerated.lean
```diff
@@ -720,37 +720,35 @@ theorem alive_AndOrXor_144  (w : Nat)   : alive_AndOrXor_144_src w  ⊑ alive_An
...
-  %v3 = llvm.mlir.constant 0
-  %v4 = llvm.icmp.eq %v1, %v3
-  %v5 = llvm.mlir.constant 0
-  %v6 = llvm.icmp.eq %v2, %v5
+  %v1 = llvm.and %a, %b : i1
+  %v2 = llvm.and %a, %d : i1
+  %v3 = llvm.mlir.constant 0 : _
+  %v4 = llvm.icmp.eq %v1, %v3 : i1
+  %v5 = llvm.mlir.constant 0 : _
+  %v6 = llvm.icmp.eq %v2, %v5 : i1
```
tobiasgrosser, line 731: This is the first one that breaks. @goens, any idea why the `(w : Nat)` is being removed?

This code is working:

```
def alive_AndOrXor_698_src (w : Nat)  :=
[alive_icom (w)| {
^bb0(%a : _, %b : _, %d : _):
  %v1 = llvm.and %a, %b : _
  %v2 = llvm.and %a, %d : _
  %v3 = llvm.mlir.constant 0 : _
  %v4 = llvm.icmp.eq %v1, %v3 : _
  %v5 = llvm.mlir.constant 0 : _
  %v6 = llvm.icmp.eq %v2, %v5 : _
  %v7 = llvm.and %v4, %v6 : i1
  llvm.return %v7 : i1
}]

def alive_AndOrXor_698_tgt (w : Nat) :=
[alive_icom (w)| {
^bb0(%a : _, %b : _, %d : _):
  %v1 = llvm.or %b, %d : _
  %v2 = llvm.and %a, %v1 : _
  %v3 = llvm.and %a, %b : _
  %v4 = llvm.and %a, %d : _
  %v5 = llvm.mlir.constant 0 : _
  %v6 = llvm.icmp.eq %v3, %v5 : _
  %v7 = llvm.mlir.constant 0 : _
  %v8 = llvm.icmp.eq %v4, %v7 : _
  %v9 = llvm.mlir.constant 0 : _
  %v10 = llvm.icmp.eq %v2, %v9 : _
  llvm.return %v10 : i1
}]
```

goens: is this being generated with the python script? In this case, the heuristic we use to find out wether to add the w broke with the "pretty" syntax and it seems to still not work on all cases. I think this might relate to the unification bug.

tobiasgrosser: Tobias Grosser: Right, this is generated with the python script. I tried to make the PR as simple as possible. There is also a choice for us to drop the ": _" in case no type is given, to further reduce the diff (or me adding it back on main).

Tobias Grosser: For some reason, alive marks the and over a, b and d to have bitwidth 1, yet the block arguments still have type _.

Tobias Grosser: Maybe it droppes the w because the uses in the and are marked as i1?

tobiasgrosser:
It seems the alive script has still some inconsistencies. We should eventually fix this and then update AliveAutogenerated.lean. For now, we just document the inconsistencies in the commit message and make sure our repo references the alive generator with the smallest diff.

PR #377, opened 2024-06-06
chore: use beautiful LLVM syntax in InstCombine/Test.lean

tobiasgrosser:
We also split out the cmp_op syntax, as for some unknown reason the merged syntax category yields the wrong return type (i1) for non-compare operations.

### DIFF
--- a/SSA/Core/ErasedContext.lean
+++ b/SSA/Core/ErasedContext.lean
@@ -1,7 +1,6 @@
 /-
 Released under Apache 2.0 license as described in the file LICENSE.
 -/
-import Mathlib.Data.Erased
 import Mathlib.Data.Finset.Basic
 import SSA.Core.HVector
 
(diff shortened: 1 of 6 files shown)


tobiasgrosser:
@alexkeizer, this breaks due to us falling in the `i1` return-type case of icmp:

```lean
macro_rules
  | `(mlir_op| $resName:mlir_op_operand = $name:InstCombine.cmp_op_name $x, $y $[: $t]?) => do
    let some opName := extractOpName name.raw
      | Macro.throwUnsupported
    let t ← t.getDM `(mlir_type| _)
    `(mlir_op| $resName:mlir_op_operand = $opName ($x, $y) : ($t, $t) -> (i1) )
```

Any idea why this is broken?

alexkeizer:
I'll investigate

alexkeizer:
I'm confused as to why this is happening, but undoing the merging of syntax declarations for `bin_op` and `cmp_op` that I suggested before seems to fix it.

PR #378, opened 2024-06-06
feat: make alive_auto cover again all proof-automation

tobiasgrosser:


SSA/Projects/InstCombine/TacticAuto.lean
```diff
@@ -83,3 +83,13 @@ macro "alive_auto": tactic =>
         try solve | (simp [bv_ofBool])
       )
    )
+
+macro "alive_auto": tactic =>
+  `(tactic|
+      (
+        simp_alive_undef
+        simp_alive_ops
+        simp_alive_case_bash
```
alexkeizer, line 92: Is there a tactic that succeeds iff there is exactly one goal (but does nothing)?
This might be useful to only case bash if that case bashing succeeds in eliminating all `none` cases, and leave the goal unchanged otherwise

bollu: Not that I know of :( I searched for a bit in this gigantic tactic list: 

https://raw.githubusercontent.com/haruhisa-enomoto/mathlib4-all-tactics/main/all-tactics.md

Probably best to write our own:

```lean
syntax (name := onlyGoal) "ensure_only_goal" : tactic

elab "ensure_only_goal" : tactic =>
  Lean.Elab.Tactic.withMainContext do
    match (← Lean.Elab.Tactic.getGoals) with
    | [_g] => pure ()
    | [] => throwError "expected exactly one goal, found no goals."
    | g₁ :: g₂ :: gs =>
      throwError m!"expected exactly one goal, found at least two goals '{g₁}' '{g₂}'."
```

alexkeizer: FYI: `elab` is like `macro` in that it defines it own syntax, no need to predefine it. I've taken this tactic and pushed it

tobiasgrosser:
Nice. I feel this can be merged.

alexkeizer:
> Alive Statistics: 55 / 93 (38 failed)

Curiously, we've now proven one extra theorem than before (on main). Do we have a good way to figure out which one this is? I'm curious why this happened

PR #389, opened 2024-06-07
chore: rename alive_icom to llvm

tobiasgrosser:
We mixed alive and llvm and used them overall inconsistently. The llvm dialect syntax really is not specific to alive. Hence, let's rename it to llvm.


tobiasgrosser:
This moves us closer to using LLVM everywhere: https://github.com/opencompl/ssa/issues/219.

bollu:
LGTM. feel invited to merge once CI passes

PR #393, opened 2024-06-07
fix: antiquotations in pretty syntax

alexkeizer:
Implements pretty quasiquotations for the LLVM dialect, so that we can write:
```lean
$v0 = llvm.mlir.constant ${2**n}
```
where `n` is a Lean variable

### DIFF
--- a/SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
+++ b/SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
@@ -134,4 +134,19 @@ private def pretty_select (w : Nat) :=
 example : pretty_test         = prettier_test_generic 32 := rfl
 example : pretty_test_generic = prettier_test_generic    := rfl
 
+
+/-! ## antiquotations test -/
+
+private def antiquot_test (x) := -- antiquotated constant value in generic syntax
+  [llvm| {
+    %0 = "llvm.mlir.constant"() { value = $(.int (x : Nat) (.i _ 32)) } : () -> (i32)
+    llvm.return %0 : i32
+  }]
+private def antiquot_test_pretty (x : Nat) := -- antiquotated constant value in pretty syntax
+  [llvm| {
+    %0 = llvm.mlir.constant $(x) : i32
+    llvm.return %0 : i32
+  }]
+example : antiquot_test = antiquot_test_pretty := rfl
+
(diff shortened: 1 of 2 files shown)


alexkeizer:
This is particularly relevant for FHE, since we do use antiquotations in the pretty examples in the paper.
I propose we move *away* from Lean's concept of antiquotations, and instead bake our own version of embedding Lean expressions into the generic MLIR parser. 

In practice, this could look something like adding the following:
```lean
syntax "$" noWs "{" term "}" (":" mlir_type)? : mlir_attr_val
```
This gives the power of using arbitary terms where an `mlir_attr_val` is expected. If the generic parser has this, it would not be too hard to then support writing (pretty) MLIR code like the following:
```lean
%two_to_the_n = arith.const ${2**n} : index
```
Where `n` is a Lean variable.

@bollu, is this roughly what you had in mind?

SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
```diff
@@ -73,11 +73,18 @@ macro_rules
...
 macro_rules
   | `(mlir_op| $res:mlir_op_operand = llvm.mlir.constant $x $[: $t]?) => do
-    let t ← t.getDM `(mlir_type| _)
-    `(mlir_op| $res:mlir_op_operand = "llvm.mlir.constant"() {value = $x:neg_num : $t} : () -> ($t) )
+      let t ← t.getDM `(mlir_type| _)
+      `(mlir_op| $res:mlir_op_operand = "llvm.mlir.constant"() {value = $x:neg_num : $t} : () -> ($t) )
+  | `(mlir_op| $res:mlir_op_operand = llvm.mlir.constant ${ $x:term } $[: $t]?) => do
+      let t ← t.getDM `(mlir_type| _)
+      let x ← `(MLIR.AST.AttrValue.int $x [mlir_type| $t])
+      `(mlir_op| $res:mlir_op_operand = "llvm.mlir.constant"() {value =  $$($x) } : () -> ($t) )
```
bollu, line 87:
```suggestion
      `(mlir_op| $res:mlir_op_operand = "llvm.mlir.constant"() {value = $$($x) } : () -> ($t) )
```

bollu (approved):
LGTM

PR #395, opened 2024-06-08
WIP

tobiasgrosser:


bollu:
@tobiasgrosser Since this was to debug / try a particular proof strategy, I'd suggest moving the contents of this PR into `Experiments/` and leaving it there.

PR #396, opened 2024-06-08
chore: migrate to lakefile.toml

goens:
With Lean 4.8 we now can use toml for the lakefile (which is the preferred variant now), so this PR migrates our build to toml.


goens:
It seems we need to wait until `lake install` lands for docgen to work with this, see https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/v4.2E8.2E0-rc1.20discussion/near/437200014

tobiasgrosser:
@goens, can this be merged?

goens:
> @goens, can this be merged?

I'm not sure to be honest, it looks like we might, according to the mathlib example: https://github.com/leanprover-community/mathlib4/pull/14229 but I haven't looked at this new docgen workflow carefully

tobiasgrosser:
I updated this to master. I just tested our `./docgen.sh`script but it does not seem to work. Maybe @hargoniX knows how to ensure doc-gen4 will still work.

PR #397, opened 2024-06-08
feat: pretty fhe syntax

alexkeizer:


tobiasgrosser:
Nice. This looks great.

bollu:
LGTM, though may I suggest renaming `uniform_op` to `uni_type_op`? That's at least clearer to me in what way the ops are uniform!

SSA/Projects/FullyHomomorphicEncryption/PrettySyntax.lean
```diff
@@ -0,0 +1,53 @@
...
+syntax mlir_op_operand " = " "arith.const" "$" noWs "{" term "}" " : " mlir_type : mlir_op
+syntax mlir_op_operand " = " "arith.const" neg_num " : " mlir_type : mlir_op
+macro_rules
+  | `(mlir_op| $v:mlir_op_operand = arith.const $x:neg_num : $t) =>
+      `(mlir_op| $v:mlir_op_operand = "arith.const" () {value = $x:neg_num } : () -> ($t))
+  | `(mlir_op| $v:mlir_op_operand = arith.const ${ $x:term } : $t) => do
+      let ctor := mkIdent ``MLIR.AST.AttrValue.int
+      let x ← `($ctor $x [mlir_type| i64])
+      --                         ^^^^^^^^^^^^^^^^ Is this right? It seems we just ignore the
+      --                                          type annotation
```
bollu, line 21: I don't understand this, I would have assumed we use `$t`?

SSA/Projects/FullyHomomorphicEncryption/PrettySyntax.lean
```diff
@@ -0,0 +1,53 @@
...
+  | `(mlir_op| $v:mlir_op_operand = poly.monomial $xs,* : ($ts,*) -> $t) =>
+      `(mlir_op| $v:mlir_op_operand = "poly.monomial" ($xs,*) : ($ts,*) -> $t)
+
+
+section Test
+variable {q n} [h : Fact (q > 1)]
+
+private def fhe_test_one_lhs := [fhe_com q, n, h | {
+  ^bb0(%a : ! R):
+  --        ^^^ TODO: should `: R` also work here?
```
bollu, line 36: Not sure I understand where the `!` leaks in from!

SSA/Projects/FullyHomomorphicEncryption/PrettySyntax.lean
alexkeizer: So that doesn't work, because we sometimes have `arith.const`s of type `index`, and something somewhere expects only `i<num>` or `f<num>`.

I took this from the examples currently in `FHE/Rewrites.lean`, e.g.:
```lean
noncomputable def lhs := [fhe_com q, n, hq| {
^bb0(%a : ! R):
  %one_int = "arith.const" () {value = 1}: () -> (i16)
  %two_to_the_n = "arith.const" () {value = $((2**n : Int))}: () -> (index)
  %x2n = "poly.monomial" (%one_int,%two_to_the_n) : (i16, index) -> (! R)
  %oner = "poly.const" () {value = 1}: () -> (! R)
  %p = "poly.add" (%x2n, %oner) : (! R, ! R) -> (! R)
  %v1 = "poly.add" (%a, %p) : (! R, ! R) -> (! R)
  "return" (%v1) : (! R) -> ()
}]
```
Here, the `value` attribute of `arith.const` is *not* given a type attribute, which means it default to `i64` (it does *not* take the type from the type annotation of the op as a whole).
I copied that behaviour, but with some reservations, hence, the comment

SSA/Projects/FullyHomomorphicEncryption/PrettySyntax.lean
goens: the issue with that was something to do with the mlir type syntax:
```lean
syntax "!" str : mlir_type
syntax "!" ident : mlir_type
syntax ident: mlir_type
syntax "_" : mlir_type

macro_rules
  | `([mlir_type| $x:ident ]) => do
        let (.ident _ xstr _ _) := x.raw | Macro.throwUnsupported
        -- ^^ We use `rawVal` rather than `val`, so that we're not affected by hygiene
        if xstr == "index"
        then
          `(MLIRType.index)
        else if xstr.front == 'i' || xstr.front == 'f'
        then do
          let xstr' := (xstr.drop 1).toString
          match xstr'.toInt? with
          | some _ =>
            let lit := Lean.Syntax.mkNumLit xstr'
            if xstr.front == 'i'
            then `(MLIRType.int .Signless $lit)
            else `(MLIRType.float $lit)
          | none =>
              Macro.throwErrorAt x $ "cannot convert suffix of i/f to int: " ++ xstr.toString
        else Macro.throwErrorAt x $ "expected i<int> or f<int>, found: " ++ xstr.toString
  | `([mlir_type| ! $x:str ]) => `(MLIRType.undefined $x)
  | `([mlir_type| ! $x:ident ]) => `(MLIRType.undefined $(Lean.quote x.getId.toString))
  -- Hardcoded meta-variable
  | `([mlir_type| _ ]) => `(MLIRType.int (φ := 1) Signedness.Signless (Width.mvar ⟨0, by simp⟩))
```

I used the `!` because that's the only way to get a generic type name in here (for dialect-specific types). This seems to match the MLIR documentation: https://mlir.llvm.org/docs/LangRef/#dialect-types

goens:
I'm a bit confused, what's the point of this PR if it doesn't change the code in `Rewrites.lean`?

bollu:
@goens We now do in fact update Rewrites.lean.

bollu:
@goens review requested please. This is a little time sensitive, since I'd like to have it to make a docker container so I can upload a camera-ready with full confidence by end of day. I might merge and ask for a post-commit review if I don't hear back from you in ~3hours.

goens:
> @goens We now do in fact update Rewrites.lean.

I see that you did in [fb92b7d](https://github.com/opencompl/ssa/pull/397/commits/fb92b7d1b521e5cd5c7f7b16ec6610e54d2fc7d1) (which was committed after my comment/review). That looks better

goens:
LGTM, thanks!

PR #401, opened 2024-06-16
chore: avoid freestanding simp lemmas in Framework

tobiasgrosser:
This will facilitate the lean 4.9 update


tobiasgrosser:
Merge as trivial.

PR #403, opened 2024-06-17
Update README with new title, logo in the title bar, and mildly amended description

bollu:


alexkeizer:
I wouldn't use pixel sizes, see https://github.com/opencompl/lean-mlir/pull/404 for a responsive way to resize the image

goens:
lgtm

bollu:
@alexkeizer certainly, happy to use `vh`.

bollu:
Thanks @alexkeizer , switched to using `vh`.

alexkeizer:
> @alexkeizer certainly, happy to use `vh`.

You could also use `em`, which refers to the standard font size and might be more appropriate here since you're making it part of the header

PR #404, opened 2024-06-17
chore: resize logo in Readme

alexkeizer:
This ensures the logo always takes up 30% of the viewport's height, making it look decently sized on both desktop and mobile. (See https://github.com/opencompl/lean-mlir/blob/resize-readme-logo/README.md for a preview).


bollu:
@alexkeizer I personally prefer https://github.com/opencompl/lean-mlir/pull/404 over this one, since I add it to the h1 header at  a favicon-ish size.

PR #410, opened 2024-06-24
feat: rewritePeepholeRecursively

bollu:
@tobiasgrosser spotted that 'rewritePeephole' only works at one level of the IR,
and will not recurse into regions.
We write a variant ('rewritePeepholeRecursively').
This first calls 'rewritePeephole' on the 'Com'.
Then, it recurses into each let-binding of the 'Com' to call 'rewritePeepholeRecursively' on all region arguments.
This ensures that the rewrite is applied to all occurrences of the lhs in all (nested) regions.

This supercedes https://github.com/opencompl/lean-mlir/pull/408


tobiasgrosser (approved):
lgtm, but check the newlines

bollu:
Newlines fixed, marked as merge when ready.

PR #413, opened 2024-06-28
feat: FBIP proof

bollu:
Formally verified proof of our FBIP model, using SSA's infrastructure for contexts, variable bindings, ...


tobiasgrosser:
This is just an empty PR. Let's close it.

PR #414, opened 2024-06-29
chore: FHE/Basic

tobiasgrosser:
- Turn non-terminating 'simp' into 'simp only''
- Add whitespace after 'rw'


tobiasgrosser:
Merge as trivial.

PR #416, opened 2024-06-29
chore: update to Lean-nightly 2024-06-28

tobiasgrosser:


tobiasgrosser:
This currently breaks at:

```lean
/-- Transport a variable from `Γ` to any mapped context `Γ.map f` -/
def toMap : Var Γ t → Var (Γ.map f) (f t)
  | ⟨i, h⟩ => ⟨i, by
      simp only [get?, map, List.getElem?_map, Option.map_eq_some']
      exact ⟨t, h, rfl⟩
    ⟩
```

and the error
```
invalid constructor ⟨...⟩, insufficient number of arguments, constructs 'Eq.refl' does not have explicit fields, but #3 provided
```

tobiasgrosser:
invalid constructor ⟨...⟩, insufficient number of arguments, constructs 'Eq.refl' does not have explicit fields, but #3 provided

tobiasgrosser:
Merged as trivial.

PR #433, opened 2024-07-06
Drop effects

tobiasgrosser:


tobiasgrosser:
This was a debugging PR that is not needed anymore.

PR #439, opened 2024-07-08
bugfix: Fix formatting/pretty-printing of MLIR in the OPT tool

AtticusKuhn:
Ideally, we want the OPT tool (without any optimizations applied) to output the same MLIR as it was inputted. Before, the OPT tool did not display (pretty print) MLIR using the same syntax as the parser.

I changed the `Repr` class of `MOp` to bring the display of MLIR in line with the parser.

Note that I have only tried this on one example: `test/LLVMDialect/InstCombine/bb0.mlir`, because OPT doesn't seem to work on other examples in `test/LLVMDialect/InstCombine/**`.

### DIFF
--- a/SSA/Core/Framework.lean
+++ b/SSA/Core/Framework.lean
@@ -245,7 +245,7 @@ mutual
 
   /-- Format string for sequence of assignments and return in a Com. -/
   partial def comReprAux (prec : Nat) : Com d Γ eff t → Format
-    | .ret v => f!"llvm.return {reprPrec v prec} : ({repr t}) -> ()"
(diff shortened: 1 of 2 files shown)


AtticusKuhn:
Here is the bash command that I used to test if the pretty-printer used the same syntax as the parser.
```bash
diff <(cat test/LLVMDialect/InstCombine/bb0.mlir) <(lake exe opt test/LLVMDialect/InstCombine/bb0.mlir)
```

tobiasgrosser:
Nice. I guess we need to change the guard_msg statements to fix the compilation. Let's discuss more tomorrow.

PR #441, opened 2024-07-09
Fix: update test cases for `test/LLVMDialect/InstCombine`

AtticusKuhn:
Previously, the test cases inside the folder `test/LLVMDialect/InstCombine` were old and did not work with `mlir-opt` (Tobias said that this was due to the changing syntax of MLIR, I believe).

I updated the test cases by downloading the llvm project, and then converting the `*.ll`` files to `*.mlir` files using the following script:

```bash
# Source directory
SRC_DIR="llvm-project-main/llvm/test/Transforms/InstCombine/"

# Destination directory
DEST_DIR="lean-mlir/test/LLVMDialect/InstCombine/"

# Loop through all .ll files in the source directory
for file in "$SRC_DIR"*.ll; do
    # Get the filename without the path
    filename=$(basename "$file")

    # Convert .ll to .mlir
    output_file="${filename%.ll}.mlir"

    echo "Converting $filename to $output_file"

    # Perform the conversion and save to the destination directory
    mlir-translate -import-llvm "$file" > "$DEST_DIR$output_file"

    # Check if the conversion was successful
    if [ $? -eq 0 ]; then
        echo "Successfully converted $filename to $output_file"
    else
        echo "Failed to convert $filename"
    fi
done
```

Note that Lean-Opt can still not handle these test cases, because it needs the operations to be in quotation marks. 

The primary difference is that before, `mlir-opt` did not work on the test cases, and now it does.


AtticusKuhn:
I changed the naming scheme of the files so that it would make a nicer git diff. 

The script is now:

```bash
# Source directory
SRC_DIR="llvm-project-main/llvm/test/Transforms/InstCombine/"

# Destination directory
DEST_DIR="lean-mlir/test/LLVMDialect/InstCombine/"

# Loop through all .ll files in the source directory
for file in "$SRC_DIR"*.ll; do
    # Get the filename without the path
    filename=$(basename "$file")

    # Convert .ll to .mlir
    output_file="${filename}.mlir"

    echo "Converting $filename to $output_file"

    # Perform the conversion and save to the destination directory
    mlir-translate -import-llvm "$file" > "$DEST_DIR$output_file"

    # Check if the conversion was successful
    if [ $? -eq 0 ]; then
        echo "Successfully converted $filename to $output_file"
    else
        echo "Failed to convert $filename"
    fi
done
```

tobiasgrosser (approved):
This looks great.

PR #443, opened 2024-07-09
Bugfix: Change Syntax so that Parentheses are Required for `llvm.mlir.constant`

### LINKED ISSUE #440: Bug: LLVM syntax macro does not accept parentheses around llvm.mlir.constant but mlir-opt does
The following program does not compile inside the LLVM syntax macro:

```lean
def alive_DivRemOfSelect_src (w : Nat) :=
  [llvm (w)| {
  ^bb0(%c: i1, %y : _, %x : _):
    %c0 = llvm.mlir.constant(0 : i32)
    %v1 = llvm.select %c, %y, %c0
    %v2 = llvm.udiv %x,  %v1
    llvm.return %v2
  }]
```

results in error:

```
unexpected token '('; expected '$' or numeral
```

Instead. the macro is expecting something more like:

```lean
def alive_DivRemOfSelect_src (w : Nat) :=
  [llvm (w)| {
  ^bb0(%c: i1, %y : _, %x : _):
    %c0 = llvm.mlir.constant 0
    %v1 = llvm.select %c, %y, %c0
    %v2 = llvm.udiv %x,  %v1
    llvm.return %v2
  }]
```

I consider this to be an error because mlir-opt does accept parentheses after a constant, and I believe that the goal of lean-opt is to mirror the behavior of mlir-opt as closely as possible.

AtticusKuhn:
Here is a reference in the LLVM  manual:

[llvm.mlir.constant (LLVM::ConstantOp)](https://mlir.llvm.org/docs/Dialects/LLVM/#llvmmlirconstant-llvmconstantop)

```bnf
operation ::= `llvm.mlir.constant` `(` $value `)` attr-dict `:` type($res)
```

```llvm
// Integer constant, internal i32 is mandatory
%0 = llvm.mlir.constant(42 : i32) : i32

// It's okay to omit i64.
%1 = llvm.mlir.constant(42) : i64

// Floating point constant.
%2 = llvm.mlir.constant(42.0 : f32) : f32

// Splat dense vector constant.
%3 = llvm.mlir.constant(dense<1.0> : vector<4xf32>) : vector<4xf32>
```

In the llvm reference manual, parentheses are required after `llvm.mlir.constant`, but the syntax did not require parentheses. Thus, I changed the syntax.

It's my intuition that the Lean version of LLVM should stay as close to the real version of LLVM as possible.

This closes #440


tobiasgrosser:
Nice. There are some build failures, but this looks overall great. @alexkeizer might be able to comment on the lean code more.

alexkeizer:
> Integer constant, internal i32 is mandatory
Interesting, I think we do allow the internal `i32` to be omitted, as in
```
%0 = llvm.mlir.constant(42) : i32
```

Like you say, we should stick with the official grammar, so that's something we should probably address too (at some point, not necessarily in this PR)

SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
```diff
@@ -47,13 +47,12 @@ macro_rules
 
 open MLIR.AST
 
-syntax mlir_op_operand " = " "llvm.mlir.constant" neg_num (" : " mlir_type)? : mlir_op
-syntax mlir_op_operand " = " "llvm.mlir.constant" ("$" noWs "{" term "}") (" : " mlir_type)? : mlir_op
+syntax mlir_op_operand " = " "llvm.mlir.constant(" neg_num (" : " mlir_type)? ")" (" : " mlir_type) : mlir_op
+syntax mlir_op_operand " = " "llvm.mlir.constant(" ("$" noWs "{" term "}") ")" (" : " mlir_type)?   : mlir_op
```
alexkeizer:
```suggestion
syntax mlir_op_operand " = " "llvm.mlir.constant" "(" neg_num (" : " mlir_type)? ")" (" : " mlir_type) : mlir_op
syntax mlir_op_operand " = " "llvm.mlir.constant" "(" ("$" noWs "{" term "}") ")" (" : " mlir_type)?   : mlir_op
```
I'd split this up into separate tokens, so that whitespace is allowed between them.
For example, `llvm.mlir.constant (42) : i32` would not parse with the original version

SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
```diff
@@ -47,13 +47,12 @@ macro_rules
...
 macro_rules
-  | `(mlir_op| $res:mlir_op_operand = llvm.mlir.constant $x $[: $t]?) => do
-      let t ← t.getDM `(mlir_type| _)
-      `(mlir_op| $res:mlir_op_operand = "llvm.mlir.constant"() {value = $x:neg_num : $t} : () -> ($t) )
-  | `(mlir_op| $res:mlir_op_operand = llvm.mlir.constant ${ $x:term } $[: $t]?) => do
+  | `(mlir_op| $res:mlir_op_operand = llvm.mlir.constant( $x $[: $t]?) : $typ ) => do
+      `(mlir_op| $res:mlir_op_operand = "llvm.mlir.constant"() {value = $x:neg_num : $typ} : () -> ($typ) )
```
alexkeizer: This implementation will completely ignore the internal type annotation `$t`.
We should check that if `t` is given, it's the same as `typ`.
And if `t` is not given, we should check that `typ` is either `f64` or `u64`, to be completely compliant with the LLVM docs

alexkeizer (commented):
The build fails because there is another file, `AliveAutoGenerated`, which has a collection of mlir programs in the DSL too.
It's a bit slow, so we don't build it by default, you have to run `lake build AliveExamples` to explicitly build them.

As the  name suggests, though, these examples are (mostly) automatically generated. Let's talk a bit more about how to fix this tomorrow

tobiasgrosser:
I like this direction. @AtticusKuhn, it might be nice to close this on Monday.

AtticusKuhn:
>  I like this direction. @AtticusKuhn, it might be nice to close this on Monday.

I realised that I don't actually need this PR to do
https://github.com/opencompl/lean-mlir/pull/452
because I can use alternate syntax.
But still, I agree that it would be good to close this on Monday.

tobiasgrosser:
Finishing this PR would be great, but don't hesitate to just close it if this is not relevant anymore.

alexkeizer:
I'm a bit hesitant to change `AliveAutoGenerated` manually. I think the best course of action at the moment is to just allow both the old syntax (without parens) alongside the proper LLVM syntax (with parens). That way, we can leave `AliveAutoGenerated` as is.

AtticusKuhn:
> I'm a bit hesitant to change `AliveAutoGenerated` manually. I think the best course of action at the moment is to just allow both the old syntax (without parens) alongside the proper LLVM syntax (with parens). That way, we can leave `AliveAutoGenerated` as is.

That's probably a good idea. I could change the script that generates `AliveAutoGenerated`, but it's no extra work just to allow both syntaxes.

AtticusKuhn:
Anyway, I've made the parentheses syntax optional, so both

```mlir
    %c0 = llvm.mlir.constant 0
```

and

```mlir
    %c0 = llvm.mlir.constant(0 : i32) : i32
```
are allowed

SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
```diff
@@ -80,7 +96,7 @@ private def pretty_test :=
 private def pretty_test_generic (w : Nat) :=
   [llvm (w)|{
   ^bb0(%arg0: _):
-    %0 = llvm.mlir.constant 8 : _
+    %0 = llvm.mlir.constant 8
```
alexkeizer: unneeded change (w.r.t. the `: _` being removed)

SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
```diff
@@ -99,7 +115,7 @@ private def prettier_test_generic (w : Nat) :=
 
 private def neg_constant (w : Nat) :=
   [llvm (w)| {
-    %0 = llvm.mlir.constant -1
+    %0 = llvm.mlir.constant(-1) : _
```
alexkeizer: We vary a bit whether we use `: _` or not, best to keep it consistent with the existing instructions around it

SSA/Projects/InstCombine/PaperExamples.lean
```diff
@@ -21,7 +21,7 @@ open BitVec
 theorem shift_mul:
     [llvm (w)| {
   ^bb0(%X : _, %Y : _):
-    %c1 = llvm.mlir.constant 1
+    %c1 = llvm.mlir.constant(1) : _
```
alexkeizer: See previous comments

alexkeizer (commented):
LGTM, a few minor comments left. Please address those and feel free to merge

PR #454, opened 2024-07-11
Allow Numeric Block Ids

AtticusKuhn:
Reference from the MLIR standard:

https://mlir.llvm.org/docs/LangRef/#identifiers-and-keywords

```
// Identifiers
bare-id ::= (letter|[_]) (letter|digit|[_$.])*
bare-id-list ::= bare-id (`,` bare-id)*
value-id ::= `%` suffix-id
alias-name :: = bare-id
suffix-id ::= (digit+ | ((letter|id-punct) (letter|id-punct|digit)*))
symbol-ref-id ::= `@` (suffix-id | string-literal) (`::` symbol-ref-id)?
value-id-list ::= value-id (`,` value-id)*
// Uses of value, e.g. in an operand list to an operation.
value-use ::= value-id (`#` decimal-literal)?
value-use-list ::= value-use (`,` value-use)*
```

Before, out parser did not allow numeric block ids. This change allows numeric block ids.

### DIFF
--- a/SSA/Core/MLIRSyntax/GenericParser.lean
+++ b/SSA/Core/MLIRSyntax/GenericParser.lean
@@ -478,20 +478,43 @@ macro_rules
 macro_rules
   | `([mlir_ops| $$($q)]) => `(coe $q)
 
+/--
+Reference from the MLIR standard
+https://mlir.llvm.org/docs/LangRef/#identifiers-and-keywords
+// Identifiers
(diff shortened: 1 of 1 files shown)


AtticusKuhn:
This was implemented as pair coding with @alexkeizer

AtticusKuhn:
The key line is

```
suffix-id ::= (digit+ | ((letter|id-punct) (letter|id-punct|digit)*))
```

Crucially, a suffix-id may consist of a sequence of digits.

PR #471, opened 2024-07-16
Feat: add overflow flags to lean semantics

### LINKED ISSUE #470: Bug: lean-mlir does not support overflow flags
See
https://grosser.zulipchat.com/#narrow/stream/248449-Project---Lean4/topic/Overflow.20Flags.20in.20LLVM
For more information,

AtticusKuhn:
See
https://github.com/opencompl/lean-mlir/issues/470
and
https://grosser.zulipchat.com/#narrow/stream/248449-Project---Lean4/topic/Overflow.20Flags.20in.20LLVM
for more information


SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -34,6 +34,9 @@ def and {w : Nat} (x y : IntW w) : IntW w := do
 The ‘or’ instruction returns the bitwise logical inclusive or of its two
 operands.
 -/
+
+structure OrParams where
```
goens: why is this a `structure`? do you expect the params to increase? if it's only a `Bool` in LLVM then making this a `structure` sounds like unnecessary complexity

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -42,9 +45,11 @@ def or? {w : Nat} (x y : BitVec w) : IntW w :=
 theorem or?_eq : LLVM.or? a b  = .some (BitVec.or a b) := rfl
 
 @[simp_llvm_option]
-def or {w : Nat} (x y : IntW w) : IntW w := do
+def or {w : Nat} (x y : IntW w) (params : OrParams := {}) : IntW w := do
   let x' ← x
   let y' ← y
+  let g ← and x y
+  guard (¬params.disjoint ∨ g = 0)
```
goens: this feels a bit hard to read, I'm curious, why not use `if`?

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -70,18 +75,25 @@ The value produced is the integer sum of the two operands.
...
 Because LLVM integers use a two’s complement representation, this instruction is appropriate for both signed and unsigned integers.
 -/
+
+structure AddParams where
+  nuw : Bool := false
+  nsw : Bool := false
+
 @[simp_llvm]
-def add? {w : Nat} (x y : BitVec w) : IntW w :=
+def add? {w : Nat} (x y : BitVec w)  : IntW w :=
```
goens:
```suggestion
def add? {w : Nat} (x y : BitVec w) : IntW w :=
```
(why add this whitespace?)

goens (commented):
Overall it seems like going in the right direction! I'm finding the guards with a bunch of negations somewhat confusing. If you feel these are the most idiomatic way because they reflect the document's semantics, maybe you can add a comment (quoting from the language ref) to at least justify these guards?

I think for merging this we also want to be sure that they really correspond to the LLVM behavior. Maybe you can 
- add the flags to the LLVM syntax and parse these params
- check them with the flow in `test/bruteforce-correctness/llvm.py`
This would give us more confidence your(/our) interpretation of the prose matches the implementation in LLVM

AtticusKuhn:
Thanks for the feedback. I guess it would be better to use `if_then_else` rather than `guard`, because it leads to an easier proof-state.

AtticusKuhn:
As for adding the flag to the parser, I think this would be a difficult task.

AtticusKuhn:
I've added the overflow flags to the parser.
It seems that the `disjoint` and `exact` flags are not even implemented by MLIR.
See
https://github.com/llvm/llvm-project/issues/99353

SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -116,12 +116,26 @@ def mkExpr (Γ : Ctxt (MetaLLVM φ).Ty) (opStx : MLIR.AST.Op φ) :
...
-        | "llvm.udiv"   => pure <| Sum.inl .udiv
+        | "llvm.mul"    =>do
+          let att := opStx.attrs.getAttr "overflowFlags"
+          match att with
+            | .none =>  pure <| Sum.inl (MOp.BinaryOp.add false false)
+            | .some y => match y with
+              | (.opaque_ "llvm.overflow" "nsw") => pure <| Sum.inl (MOp.BinaryOp.mul true false)
+              | (.opaque_ "llvm.overflow" "nuw") => pure <| Sum.inl (MOp.BinaryOp.mul false true)
+              | (.opaque_ "llvm.overflow" s ) =>throw <| .generic s!"flag {s} not allowed"
+              | _ => throw <| .generic s!"flag not allowed"
```
alexkeizer, line 127: It would be helpful to include the flag that was present in the error. Something like `unexpected attribute `$att``

SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -116,12 +116,26 @@ def mkExpr (Γ : Ctxt (MetaLLVM φ).Ty) (opStx : MLIR.AST.Op φ) :
...
-
-        | "llvm.mul"    => pure <| Sum.inl .mul
-        | "llvm.sub"    => pure <| Sum.inl .sub
-        | "llvm.sdiv"   => pure <| Sum.inl .sdiv
-        | "llvm.udiv"   => pure <| Sum.inl .udiv
+        | "llvm.mul"    =>do
+          let att := opStx.attrs.getAttr "overflowFlags"
+          match att with
+            | .none =>  pure <| Sum.inl (MOp.BinaryOp.add false false)
+            | .some y => match y with
```
alexkeizer: Minor nit: since we use `attr` as an abbreviation for attribute elsewhere, it's best to be consistent and not use `att` as an alternative abbreviation as well.

Single letter variables can be fine when the scope is limited like here, but it's better to use a more descriptive name. In this case, we can use `?` as a suffix to explicate which variable is the option, and which is the actual attribute

```suggestion
          let attr? := opStx.attrs.getAttr "overflowFlags"
          match attr? with
            | .none =>  pure <| Sum.inl (MOp.BinaryOp.add false false)
            | .some attr => match attr with
```

SSA/Projects/InstCombine/AliveHandwrittenLargeExamples.lean
```diff
@@ -134,6 +135,7 @@ def alive_simplifyMulDivRem805 (w : Nat) :
       rw [LLVM.sdiv?_denom_zero_eq_none]
       apply Refinement.none_left
     case neg =>
+      simp
```
alexkeizer, line 138: non-terminal simp (if a simp is not the last element of a proof, it should be `simp only` to reduce chances of the proof breaking when the global simp-set changes, you can use `simp?` to "squeeze" the simp)

SSA/Projects/InstCombine/ComWrappers.lean
```diff
@@ -191,7 +191,7 @@ def udiv {Γ : Ctxt _} (w : ℕ) (l r : Nat)
       := by get_elem_tactic) :
     Expr InstCombine.LLVM Γ .pure (InstCombine.Ty.bitvec w) :=
   Expr.mk
-    (op := InstCombine.MOp.udiv w)
+    (op := InstCombine.MOp.udiv  false w)
```
alexkeizer, line 194:
```suggestion
    (op := InstCombine.MOp.udiv false w)
```

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -42,10 +42,13 @@ def or? {w : Nat} (x y : BitVec w) : IntW w :=
 theorem or?_eq : LLVM.or? a b  = .some (BitVec.or a b) := rfl
 
 @[simp_llvm_option]
-def or {w : Nat} (x y : IntW w) : IntW w := do
+def or {w : Nat} (x y : IntW w) (disjoint : Bool := false) : IntW w := do
   let x' ← x
   let y' ← y
-  or? x' y'
+  if  disjoint ∧  x' &&&  y' ≠ 0 then
```
alexkeizer, line 48: Please be mindful of whitespace 
```suggestion
  if disjoint ∧ x' &&& y' ≠ 0 then
```

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -78,11 +86,23 @@ def add? {w : Nat} (x y : BitVec w) : IntW w :=
...
 @[simp_llvm_option]
-def add {w : Nat} (x y : IntW w) : IntW w := do
-  let x' ← x
-  let y' ← y
-  add? x' y'
-
+def add {w : Nat} (x y : IntW w) (params : AddParams := {}) : IntW w := do
+  let x ← x
+  let y ← y
+  if (params.nsw ∧  (x.toInt + y.toInt) < -(2^(w-1)) ∧ (x.toInt + y.toInt) ≥ 2^w) ∨ ( params.nuw ∧  (x.toNat + y.toNat) ≥ 2^w) then
```
alexkeizer, line 92: This line you also have some inconsistent whitespace

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -91,15 +111,24 @@ Because LLVM integers use a two’s complement representation, this instruction
 @[simp_llvm]
 def sub? {w : Nat} (x y : BitVec w) : IntW w :=
   pure <| x - y
+structure SubParams where
+  nuw : Bool := false
+  nsw : Bool := false
```
alexkeizer, line 116: Given that `AddParams` is the same as `SubParams` is the same as `MulParams`, it makes sense to unify those into a single `OverflowFlags` structure

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -42,10 +42,13 @@ def or? {w : Nat} (x y : BitVec w) : IntW w :=
 theorem or?_eq : LLVM.or? a b  = .some (BitVec.or a b) := rfl
 
 @[simp_llvm_option]
-def or {w : Nat} (x y : IntW w) : IntW w := do
+def or {w : Nat} (x y : IntW w) (disjoint : Bool := false) : IntW w := do
   let x' ← x
   let y' ← y
-  or? x' y'
+  if  disjoint ∧  x' &&&  y' ≠ 0 then
+    .none
```
alexkeizer, line 49: Since `Option.none` is brought into scope by the prelude, you don't need the dot here (nor the rest of the file)
```suggestion
    none
```

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -166,27 +213,41 @@ at width 2, -4 / -1 is considered overflow!
...
+def sdiv {w : Nat} (x y : IntW w) (params : SdivParams := {})  : IntW w := do
   let x' ← x
   let y' ← y
-  sdiv? x' y'
-
+  if (params.exact ∧  ¬ (x'.toNat ∣ y'.toNat)) then
+    .none
+  else
+    sdiv? x' y'
+@[simp, reducible]
```
alexkeizer, line 243: Please leave an empty line between definitions

AtticusKuhn:
See
https://github.com/opencompl/lean-mlir/pull/480

PR #476, opened 2024-07-22
Began implementing intrinsically typed regions

imbrem:


SSA/Projects/IntrinsicRegions/Framework.lean
```diff
@@ -0,0 +1,86 @@
+import Init.Data.BitVec.Basic
```
bollu, line 10:
```suggestion
/-
We develop the core theory of unstructured control flow that is intrinsically well-typed.
The key insight to intrinsically well-type a CFG is to realize that a CFG is inductively well-typed
when performing induction on the *dominator tree*. 

Author: Jad Ghaliyani, Siddharth Bhat
-/
import Init.Data.BitVec.Basic
```

SSA/Projects/IntrinsicRegions/Framework.lean
```diff
@@ -0,0 +1,86 @@
...
+  | val : {α : Ty} → (v : toType α) → Term Γ α
+  -- | op : φ → Term φ → Term φ
+  -- | let1 : Term φ → Term φ → Term φ
+  -- | pair : Term φ → Term φ → Term φ
+  -- | unit : Term φ
+  -- | let2 : Term φ → Term φ → Term φ
+  -- | inl : Term φ → Term φ
+  -- | inr : Term φ → Term φ
+  -- | case: Term φ → Term φ → Term φ → Term φ
+  -- | abort : Term φ → Term φ
```
bollu: Delete dead code. 

```suggestion
```

SSA/Core/ErasedContext.lean
```diff
@@ -404,6 +476,58 @@ theorem Valuation.reassignVar_eq_of_lookup [DecidableEq Ty]
...
+def CoValuation.ofVar {Γ : Ctxt Ty} (v : Γ.Var α) (a : toType α) : CoValuation Γ :=
+  ⟨α, v, a⟩
+
+/-- transport/pullback a valuation along a context homomorphism. -/
+def CoValuation.comap {Γi Γo : Ctxt Ty} (Γiv: Γi.CoValuation) (hom : Ctxt.Hom Γi Γo) : Γo.CoValuation where
+  t := Γiv.t
+  var := hom Γiv.var
+  val := Γiv.val
+
+def CoValuation.appendEquiv {Γ Δ : Ctxt Ty}
```
bollu, line 644:
```suggestion
/-- A covaluation `(Γ ++ Δ)` is either a covaluation `Γ` or a covaluation `Δ` -/
def CoValuation.appendEquiv {Γ Δ : Ctxt Ty}
```

bollu (approved):
LGTM after changes I've made, but also ping @tobiasgrosser. @alexkeizer  on Zulip.

SSA/Core/ErasedContext.lean
```diff
@@ -18,6 +18,29 @@ notation "⟦" x "⟧" => TyDenote.toType x
...
+  elim : (toType a → γ) → (toType b → γ) → toType (C.coprod a b) → γ
+
+/-
+TODO: implement a thing `left? : toType (C.coprod α β) → Option (toType α)` using `elim`. This will
+allow us to cleanup case_inl by writing `(ha : (l : toType  α) ∈ a.evaluate VΓ |>.left?)`
+-/
+
+/-
+TODO: add lawful instance of TyHasCoprodDenote for proving properties about the universal property.
+-/
```
alexkeizer: Would it be easier to have `LawfulTyHasCoprod` simply state that for all `a, b : Ty`, it is the case that `toType (C.coprod a b)` is equivalent to `Sum (toType a) (toType b)`, then you would not need `TyHasCoprodDenote` at all anymore

bollu: I don't think that suffices, because my iso can be non-natural. I can do the terrible trick I showed you last time, where I match on `a : Ty` and build something super non-canonical.

SSA/Projects/IntrinsicRegions/Framework.lean
bollu: @imbrem unresolved comment?

SSA/Core/ErasedContext.lean
alexkeizer: > my iso can be non-natural

I'm not sure what natural means in this context? 

You're right that this equivalence is stronger than `TyHasCoprodDenote` as currently stated, but surely a Lawful version would impy the equivalence?
Normally, you'd separate out the laws into a separate typeclass to facilitate programming where you might not necessarily care about things being 100% lawful, but that hardly seems applicable in our setting; if you're just programming you'd probably use untyped datastructures, rather than our highly dependent ones.

tobiasgrosser:
No update since a while. Let's close this to keep our PRs clean.

PR #477, opened 2024-07-22
feat: make context generalization in simp_peephole tactic optional

alexkeizer:
This PR separates the framework simplification done in `simp_peephole` from the valuation massaging, by making the final `at \Gv:ident` part of the tactic syntax optional.

This came up in a pair programming session with @ymherklotz, where we had already massaged the valuation to be in terms of Lean variables using Valuation.ofPair, so we did not really need to pass a valuation, but we did want the simplification.


tobiasgrosser (approved):
Thank you @alexkeizer.

PR #479, opened 2024-07-23
chore: only use of_bool_tactic if it solves the goal 100%

tobiasgrosser:
This allows us to reintroduce 'simp only [BitVec.toNat_eq]' as a mean to solve ofBool goals, which reduces the failing equations from 34 to 29.


AtticusKuhn:
I minorly shortened the code, but otherwise, it looks good to me.

AtticusKuhn (approved):

PR #480, opened 2024-07-23
Feat: Add Overflow Semantics to Addition in LLVM

AtticusKuhn:
This is a smaller, scaled back version of 
https://github.com/opencompl/lean-mlir/pull/471

```
I am wondering if we can scope it a bit smaller and just add overflow flags for addition. I feel this would allow us to iterate quickly on the right implementation and then expand it to all ops.
```

https://grosser.zulipchat.com/#narrow/stream/446584-Project---Lean4---BitVectors/topic/Overflow.20Flags.20in.20LLVM/near/453334239


tobiasgrosser:
I am surprised to see still a need to print the default false. Is this unavoidable.

AtticusKuhn:
> I am surprised to see still a need to print the default false. Is this unavoidable.

I was also surpised that Lean prints `MOp.BinaryOp.add w false`, i.e. Lean still prints the default parameter. I'm not sure why this is.

AtticusKuhn:
Maybe someone with more knowledge of the Lean internals can comment as to why this is?

tobiasgrosser:
It might be useful to build a minimal example and ask on the official lean zulip.

AtticusKuhn:
> It might be useful to build a minimal example and ask on the official lean zulip.

Good idea, I posted this on the Lean Zulip
https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Why.20does.20Lean.204.20Print.20Default.20Values.3F/near/453447387

AtticusKuhn:
>  I am surprised to see still a need to print the default false. Is this unavoidable.


The pretty printer only considers omitting the last optional argument. I got around this by writing
```lean4
inductive MOp.BinaryOp : Type
  | add (nswnuw  : Bool × Bool := ⟨ false , false ⟩ )
```

SSA/Projects/InstCombine/Test.lean
```diff
@@ -197,7 +197,7 @@ info: ⟨[MTy.bitvec (ConcreteOrMVar.concrete 32)],
                 (Com.ret ⟨0, ⋯⟩)))))⟩⟩⟩
 -/
 #guard_msgs in #reduce com
-
+#reduce com
 theorem com_Γ : com.1 = (Γn 1) := by rfl
```
tobiasgrosser: Is this change unrelated?

AtticusKuhn: yes

tobiasgrosser (approved):
This looks great to me. Maybe some comments from a parsing expert such as @bollu .

SSA/Projects/InstCombine/Base.lean
```diff
@@ -275,7 +306,7 @@ def Op.denote (o : LLVM.Op) (op : HVector TyDenote.toType (DialectSignature.sig
   | Op.lshr _      => LLVM.lshr   (op.getN 0) (op.getN 1)
   | Op.ashr _      => LLVM.ashr   (op.getN 0) (op.getN 1)
   | Op.sub _       => LLVM.sub    (op.getN 0) (op.getN 1)
-  | Op.add _       => LLVM.add    (op.getN 0) (op.getN 1)
+  | Op.add w nsw nuw       => LLVM.add    (op.getN 0) (op.getN 1) ⟨ nsw ,  nuw ⟩
```
bollu: We should either make the whitespace consistent, or give up on the alignment

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -78,11 +78,25 @@ def add? {w : Nat} (x y : BitVec w) : IntW w :=
...
 @[simp_llvm_option]
-def add {w : Nat} (x y : IntW w) : IntW w := do
+def add {w : Nat} (x y : IntW w) (nswnuw : Bool × Bool := ⟨ false , false ⟩ ) : IntW w := do
   let x' ← x
   let y' ← y
-  add? x' y'
-
+  let nsw := nswnuw.1
+  let nuw := nswnuw.2
+  if (nsw ∧  (x'.toInt + y'.toInt) < -(2^(w-1)) ∧ (x'.toInt + y'.toInt) ≥ 2^w) ∨ (nuw ∧  (x'.toNat + y'.toNat) ≥ 2^w) then
```
bollu:
```suggestion
  if (nsw ∧  ((x'.toInt + y'.toInt) < -(2^(w-1)) ∨ (x'.toInt + y'.toInt) ≥ 2^w)) ∨ (nuw ∨  (x'.toNat + y'.toNat) ≥ 2^w) then
```

Pretty sure we need an `∨` instead of an `∨`: It's an overflow if it goes either over or under the range.

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -78,11 +78,25 @@ def add? {w : Nat} (x y : BitVec w) : IntW w :=
...
+
+set_option allowUnsafeReducibility true
+@[simp, reducible]
+theorem add_reduce (x y : IntW w) :  add x y = match x , y with
+  | .none , _ => .none
+  | _ , .none => none
+  | .some a , .some b  => .some (a + b) := by
+  rcases x
+  all_goals (cases y)
+  all_goals (try simp ; try rfl)
```
bollu, line 99:
```suggestion
set_option allowUnsafeReducibility true
@[simp, reducible]
theorem add_reduce (x y : IntW w) : add x y = match x , y with
  | .none , _ => .none
  | _ , .none => none
  | .some a , .some b  => .some (a + b) := by
  rcases x
  all_goals (cases y)
  all_goals (try simp ; try rfl)
```

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -78,11 +78,25 @@ def add? {w : Nat} (x y : BitVec w) : IntW w :=
...
+  let nsw := nswnuw.1
+  let nuw := nswnuw.2
+  if (nsw ∧  (x'.toInt + y'.toInt) < -(2^(w-1)) ∧ (x'.toInt + y'.toInt) ≥ 2^w) ∨ (nuw ∧  (x'.toNat + y'.toNat) ≥ 2^w) then
+    none
+  else
+    add? x' y'
+
+set_option allowUnsafeReducibility true
+@[simp, reducible]
+theorem add_reduce (x y : IntW w) :  add x y = match x , y with
```
bollu:
```suggestion
theorem add_reduce (x y : IntW w) : add x y = match x , y with
```

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -78,11 +78,25 @@ def add? {w : Nat} (x y : BitVec w) : IntW w :=
...
+  | .none , _ => .none
+  | _ , .none => none
+  | .some a , .some b  => .some (a + b) := by
```
bollu:
```suggestion
  | .some a , .some b => .some (a + b) := by
```

SSA/Projects/InstCombine/LLVM/Semantics.lean
bollu: Correct me if I am wrong, but I was under the impression that this can be replaced by: 
```suggestion
  rcases x <;> cases y <;> [simp, rfl]
 ```

File: SSA/Projects/InstCombine/LLVM/Semantics.lean [hunk 7]
```diff
@@ -78,11 +78,25 @@ def add? {w : Nat} (x y : BitVec w) : IntW w :=
 theorem add?_eq : LLVM.add? a b  = .some (BitVec.add a b) := rfl
 
 @[simp_llvm_option]
-def add {w : Nat} (x y : IntW w) : IntW w := do
+def add {w : Nat} (x y : IntW w) (nswnuw : Bool × Bool := ⟨ false , false ⟩ ) : IntW w := do
```
bollu:
```suggestion
def add {w : Nat} (x y : IntW w) (nswnuw : Bool × Bool := ⟨ false , false ⟩ ) : IntW w := do
```

Maybe it's just me, but I find putting the options into a `structure` to be cleaner. But IIRC, you had this argument with @goens who ruled in favour of not having a `structure`?

SSA/Projects/InstCombine/LLVM/Semantics.lean
bollu: Even better would be to take this condition and lift it to defs. Something like:

```lean
def intOverflow? (x y : BitVec w) : Prop := 
  ((x.toInt + y.toInt) < -(2^(w-1)) ∨ (x.toInt + y.toInt) ≥ 2^w))

def natOverflow? (x y : BitVec w) : Prop := (x.toNat + y.toNat) ≥ 2^w
```

And then to write the condition as:

```lean
if (nsw ∧ intOverflow? x' y')  ∨ (nuw ∧ natOverflow? x' y')
```

bollu (approved):
LGTM, but it'd be nice if we can refactor the flags.

SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -106,7 +106,15 @@ def mkExpr (Γ : Ctxt (MetaLLVM φ).Ty) (opStx : MLIR.AST.Op φ) :
...
         | "llvm.srem"   => pure <| Sum.inl .srem
-        | "llvm.add"    => pure <| Sum.inl .add
+        | "llvm.add"    =>  do
+          let attr? := opStx.attrs.getAttr "overflowFlags"
+          match attr? with
+            | .none =>  pure <| Sum.inl (MOp.BinaryOp.add ⟨ false , false ⟩ )
+            | .some y => match y with
+              | (.opaque_ "llvm.overflow" "nsw") => pure <| Sum.inl (MOp.BinaryOp.add ⟨ true ,false⟩ )
+              | (.opaque_ "llvm.overflow" "nuw") => pure <| Sum.inl (MOp.BinaryOp.add ⟨ false, true ⟩ )
+              | (.opaque_ "llvm.overflow" s ) => throw <| .generic s!"The overflow flag {s} not allowed. We currently support nsw (no signed wrap) and nuw (no unsigned wrap)"
```
bollu: This does not seem to handle the case where we make both `nsw` and `nuw`? Don't we need to have another case where we have both flags?

bollu (changes requested):
I think we need to fix the flag handling to handle cases where we have both `nsw` and `nuw`.

SSA/Projects/InstCombine/LLVM/Semantics.lean
alexkeizer: @bollu `intOverflow?` is a `Prop` thus the naming convention demands an UpperCamelCase name. At the risk of bike-shedding, I'd propose `AddSignedWraps` and `AddUnsignedWraps` to mirror the name of the flags (and to explicate this is about whether addition overflows)

SSA/Projects/InstCombine/LLVM/Semantics.lean
alexkeizer: Also, `rcases` without a pattern seems like it should be just a regular `cases`, no?

SSA/Projects/InstCombine/LLVM/Semantics.lean
alexkeizer: Where was this discussion? I also feel a structure is cleaner, especially when compared to having an anonymous product

SSA/Core/MLIRSyntax/GenericParser.lean
```diff
@@ -549,6 +549,13 @@ syntax "@" str : mlir_attr_val_symbol
 syntax "#" ident : mlir_attr_val -- alias
 syntax "#" strLit : mlir_attr_val -- aliass
 
+declare_syntax_cat dialect_attribute_contents
+syntax mlir_attr_val : dialect_attribute_contents
+syntax "(" dialect_attribute_contents + ")" : dialect_attribute_contents
+syntax "[" dialect_attribute_contents + "]": dialect_attribute_contents
+syntax "{" dialect_attribute_contents + "}": dialect_attribute_contents
+syntax "#" ident "<" dialect_attribute_contents ">" : mlir_attr_val -- opaqueAttr
+
```
alexkeizer: What is a `dialect_attribute_contents`, and how does it differ from a regular `mlir_attr_val`?

Since there seems to be quite some discussion around this PR already, I'd suggest taking the changes to the parser, which seem somewhat independent, out of this minimal PR.

SSA/Projects/InstCombine/AliveHandwrittenLargeExamples.lean
```diff
@@ -134,6 +135,7 @@ def alive_simplifyMulDivRem805 (w : Nat) :
       rw [LLVM.sdiv?_denom_zero_eq_none]
       apply Refinement.none_left
     case neg =>
+      simp only [add_tsub_cancel_right, false_and, or_self, ↓reduceIte,Option.some_bind]
```
alexkeizer: This change, and others in this file, seem unrelated, in the sense that I can't see why the old proof might've broken

SSA/Projects/InstCombine/Base.lean
```diff
@@ -88,6 +88,22 @@ deriving Repr, DecidableEq, Inhabited
...
+  | urem
+  | srem
+  | add (nswnuw  : Bool × Bool := ⟨ false , false ⟩ )
+  | mul
+  | sub
+  | sdiv
+  | udiv
+deriving DecidableEq, Inhabited
+/-- Homogeneous, binary operations -/
+inductive g : Type
```
alexkeizer: Leave an empty line in between definition/inductives (there are some exceptions, when items are small and very closely related, but 99/100 times you want the empty line)

alexkeizer: `g` is not a very descriptive name, which is not helped by the fact that you merely copied the docstring!
It seems to be a copy of `MOp` without the overflow flags. Document this connection, and explain why we need it.

SSA/Projects/InstCombine/Base.lean
```diff
@@ -102,7 +118,22 @@ inductive MOp.BinaryOp : Type
   | sdiv
   | udiv
 deriving Repr, DecidableEq, Inhabited
-
+def aux : MOp.BinaryOp → ℕ → Lean.Format
```
alexkeizer: Similarly, `aux` is not descriptive at all: it tells me it's auxiliary to something, but I have no idea what the main thing is! This would better be names as `repr`, or maybe `reprAux` if you want to emphasize it's only a part of the `repr` implementation.

SSA/Projects/InstCombine/Base.lean
```diff
@@ -102,7 +118,22 @@ inductive MOp.BinaryOp : Type
...
+  | .ashr => reprPrec g.ashr
+  | .urem => reprPrec g.urem
+  | .srem => reprPrec g.srem
+  | .add _ => reprPrec g.add
+  | .mul => reprPrec g.mul
+  | .sub => reprPrec g.sub
+  | .sdiv => reprPrec g.sdiv
+  | .udiv => reprPrec g.udiv
+instance : Repr (MOp.BinaryOp) where
+  reprPrec op w := ((toString (aux op w)).replace "InstCombine.g" "InstCombine.MOp.BinaryOp").replace "false" ""
```
alexkeizer: This code is quite obscure, and it's not super clear to me, but I guess the purpose of this is to "hide" the default `false` values? Why do we need those to be hidden?

I guess the purpose of `g` is to use it's automatically derived `Repr` instance here? This is feels like a very roundabout way of achieving that goal. 

Consider something like the following: 
```lean
def repr : ...
  | .and => "and"
  | .or => "or"
  ...
  | .add flags =>
    let \<nsw, nuw \> := flags
    if nsw = false \land nuw = false then
      "add"
    else
      "add " ++ repr flags
```
This is much more readable, and means you can get rid of `g`, so it should be significantly simpler as well.

SSA/Projects/InstCombine/Base.lean
```diff
@@ -216,7 +247,7 @@ namespace Op
 @[match_pattern] abbrev urem   : Nat → Op := MOp.urem   ∘ .concrete
 @[match_pattern] abbrev srem   : Nat → Op := MOp.srem   ∘ .concrete
 @[match_pattern] abbrev select : Nat → Op := MOp.select ∘ .concrete
-@[match_pattern] abbrev add    : Nat → Op := MOp.add    ∘ .concrete
+@[match_pattern] abbrev add (w : Nat) (nsw : Bool := false) (nuw : Bool := false) : Op:=  MOp.add (.concrete w) nsw nuw
```
alexkeizer: line is too long, also please be consistent with whitespace
```suggestion
@[match_pattern] abbrev add (w : Nat) (nsw : Bool := false) (nuw : Bool := false) : Op := 
  MOp.add (.concrete w) nsw nuw
```

The other lines are packed close together, since they are all relatively simple, uniform, and closely related. 
This definition breaks that mold. Once a definition spans more than one line, we almost always (or at least, this is my opninion, tastes on code style do differ) want to have empty lines separating it from other defs. In this case, it's probably best to yoink `add` after `const` (notice how `icmp` and `const` are also separated, because they are also different)

SSA/Projects/InstCombine/Base.lean
```diff
@@ -127,7 +158,7 @@ namespace MOp
 @[match_pattern] def ashr   (w : Width φ) : MOp φ := .binary w .ashr
 @[match_pattern] def urem   (w : Width φ) : MOp φ := .binary w .urem
 @[match_pattern] def srem   (w : Width φ) : MOp φ := .binary w .srem
-@[match_pattern] def add    (w : Width φ) : MOp φ := .binary w .add
+@[match_pattern] def add    (w : Width φ) (nsw : Bool := false) (nuw : Bool := false) : MOp φ := .binary w (.add  ⟨ nsw , nuw ⟩ )
```
alexkeizer: Why does `BinaryOp.add` take its flags as a pair of Bools, but `MOp` takes them as two separate arguments?

SSA/Projects/InstCombine/LLVM/Enumerator.lean
```diff
@@ -184,7 +184,7 @@ def generateRawSemantics : IO Unit := do
   rows := rows.append (binopRows "and"  (fun w a b => InstCombine.Op.denote (.and w)  [a,b]ₕ))
   rows := rows.append (binopRows "or"   (fun w a b => InstCombine.Op.denote (.or w)   [a,b]ₕ))
   rows := rows.append (binopRows "xor"  (fun w a b => InstCombine.Op.denote (.xor w)  [a,b]ₕ))
-  rows := rows.append (binopRows "add"  (fun w a b => InstCombine.Op.denote (.add w)  [a,b]ₕ))
+  rows := rows.append (binopRows "add"  (fun w a b => InstCombine.Op.denote (.add w false false)  [a,b]ₕ))
```
alexkeizer: Is this change not unnecessary, since the flags both have `false` as default value?

alexkeizer (changes requested):
There are still quite some inconsistencies

SSA/Core/MLIRSyntax/GenericParser.lean
AtticusKuhn: I got this from the MLIR reference:

[Dialect Attribute Values](https://mlir.llvm.org/docs/LangRef/#dialect-attribute-values)
Similarly to operations, dialects may define custom attribute values.

```bnf
dialect-namespace ::= bare-id

dialect-attribute ::= `#` (opaque-dialect-attribute | pretty-dialect-attribute)
opaque-dialect-attribute ::= dialect-namespace dialect-attribute-body
pretty-dialect-attribute ::= dialect-namespace `.` pretty-dialect-attribute-lead-ident
                                              dialect-attribute-body?
pretty-dialect-attribute-lead-ident ::= `[A-Za-z][A-Za-z0-9._]*`

dialect-attribute-body ::= `<` dialect-attribute-contents+ `>`
dialect-attribute-contents ::= dialect-attribute-body
                            | `(` dialect-attribute-contents+ `)`
                            | `[` dialect-attribute-contents+ `]`
                            | `{` dialect-attribute-contents+ `}`
                            | [^\[<({\]>)}\0]+
```

SSA/Projects/InstCombine/LLVM/EDSL.lean
AtticusKuhn: You are correct. I hadn't considered that. This case actually does appear in our test data:

```mlir
  llvm.func @shl_add_lshr_flag_preservation(%arg0: i32, %arg1: i32, %arg2: i32) -> i32 {
    %0 = llvm.shl %arg0, %arg1 overflow<nuw>  : i32
    %1 = llvm.add %0, %arg2 overflow<nsw, nuw>  : i32
    %2 = llvm.lshr %1, %arg1  : i32
    llvm.return %2 : i32
  }
```

tobiasgrosser:
@AtticusKuhn, it seems there are still comments open. If we can get them addressed, this patch can be merged.

AtticusKuhn:
> @AtticusKuhn, it seems there are still comments open. If we can get them addressed, this patch can be merged.

Got it. I will address all the comments today.

SSA/Projects/InstCombine/AliveHandwrittenLargeExamples.lean
AtticusKuhn: I would have thought so as well, but it turns out that this line is nessecary.

tobiasgrosser:
There is a new related PR proposed for lean: https://github.com/leanprover/lean4/pull/4854.

SSA/Projects/InstCombine/AliveHandwrittenLargeExamples.lean
```diff
@@ -118,6 +119,7 @@ def alive_simplifyMulDivRem805 (w : Nat) :
   unfold MulDivRem805_lhs MulDivRem805_rhs
   simp only [simp_llvm_wrap]
   simp_alive_ssa
+  simp only [LLVM.add_reduce]
```
tobiasgrosser: We should not need to add this simp only rewrite explicitly here, instead it should be executed as either `simp_alive_ssa` or `simp_alive_undef`.

kmill:
> There is a new related PR proposed for lean: [leanprover/lean4#4854](https://github.com/leanprover/lean4/pull/4854).

This optional parameters pretty printing PR is merged and you can expect it in the next nightly.

tobiasgrosser:
> > There is a new related PR proposed for lean: [leanprover/lean4#4854](https://github.com/leanprover/lean4/pull/4854).
> 
> This optional parameters pretty printing PR is merged and you can expect it in the next nightly.

Lovely. Thank you so much.

tobiasgrosser:
The latest version of lean merged in this PR now supports multiple default arguments. Should we use this feature, or do we stay with the struct?

AtticusKuhn:
> The latest version of lean merged in this PR now supports multiple default arguments. Should we use this feature, or do we stay with the struct?

 I personally think a structure would be cleaner, but I think either method could work.

tobiasgrosser:
> > The latest version of lean merged in this PR now supports multiple default arguments. Should we use this feature, or do we stay with the struct?
> 
> I personally think a structure would be cleaner, but I think either method could work.

Then let's keep the structure if nobody else complains. @alexkeizer @bollu, it seems your `changes requested` still blocks the PR. Can you do a final pass such that this can be merged.

SSA/Projects/InstCombine/Base.lean
```diff
@@ -227,6 +264,11 @@ namespace Op
 @[match_pattern] abbrev icmp (c : IntPredicate)   : Nat → Op  := MOp.icmp c ∘ .concrete
 @[match_pattern] abbrev const (w : Nat) (val : ℤ) : Op        := MOp.const (.concrete w) val
 
+
+/-- Add is seperate from the other operations because it takes in 2 flags: nuw and nsw.-/
+@[match_pattern] abbrev add (w : Nat) (flags: AdditionFlags :=
```
tobiasgrosser:
```suggestion
/-- Add is separate from the other operations because it takes in 2 flags: nuw and nsw.-/
@[match_pattern] abbrev add (w : Nat) (flags: AdditionFlags :=
```

AtticusKuhn:
I think this PR looks good to me now.

SSA/Core/MLIRSyntax/GenericParser.lean
alexkeizer: Please don't hit resolve unless its uncontroversial. I'd still say these *syntactic* changes (how are overflow flags represented in the DSL) should be done separately from the main *semantic* change of (optionally) having UB on overflow.

That said, I won't die on this hill. If you're convinced this is the right approach, feel free to go ahead. I would still like to see a docstring explain that `dialect_attribute_contents` is a dialect-namespaced `mlir_attr_val`

SSA/Core/MLIRSyntax/GenericParser.lean
```diff
@@ -551,9 +551,37 @@ declare_syntax_cat mlir_attr_val_symbol
...
+                            | `(` dialect-attribute-contents+ `)`
+                            | `[` dialect-attribute-contents+ `]`
+                            | `{` dialect-attribute-contents+ `}`
+                            | [^\[<({\]>)}\0]+
+```
+-/
+syntax "(" dialect_attribute_contents + ")" : dialect_attribute_contents
+syntax "[" dialect_attribute_contents + "]": dialect_attribute_contents
+syntax "{" dialect_attribute_contents + "}": dialect_attribute_contents
+syntax "#" ident "<" sepBy(mlir_attr_val, ",")  ">" : mlir_attr_val
```
alexkeizer:
```suggestion
syntax "#" ident "<" mlir_attr_val,*  ">" : mlir_attr_val
```
This should be exactly the same, but a more succinct/idiomatic way to phrase it

SSA/Projects/InstCombine/Base.lean
alexkeizer: @AtticusKuhn you've still not added the newline inbetween these two definitions

SSA/Projects/InstCombine/Base.lean
alexkeizer: > ```
> /--
> The reason that I am using the admittedly hacky and ad-hoc method is that I want to preserve the guard_msgs > statements, otherwise the build will fail.
> But the default Repr instance has some fancy behavior where depending on the indentation it will sometimes wrap in parentheses.
> I think the only way to replicate this behavior is to have another class and piggy-back off its default Repr class
> -/
>```

Thanks for writing this comment to explain your rationale! You are right that this solution deserves some comment in the code, but first this would fit better here in the code review --- also, as an aside, code comments are best not written in the first person, since a future reader might not immediately know who actually wrote that comment.

More concretely, the fancy behaviour you refer to is simply behaviour of [Std.Format](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Format/Basic.html#Std.Format).  You can have a look at the [derive handler of Repr](https://github.com/leanprover/lean4/blob/c375e19f6b656fcd594cdca3a38b8578634df8cd/src/Lean/Elab/Deriving/Repr.lean#L82-L86) to see what code it generates; in particular the line: `Repr.addAppParen (Format.group (Format.nest (if prec >= max_prec then 1 else 2) ($rhs:term))) prec)` is where the parenthesization happens.

We could write similar logic by hand. See for example:
```
open Std (Format) in
def reprWithoutFlags (op : MOp.BinaryOp) (prec : Nat) : Format :=
  let op := match op with
    | .and    => "and"
    | .or     => "or"
    | .xor    => "xor"
    | .shl    => "shl"
    | .lshr   => "lshr"
    | .ashr   => "ashr"
    | .urem   => "urem"
    | .srem   => "srem"
    | .add _  => "add"
    | .mul    => "mul"
    | .sub    => "sub"
    | .sdiv   => "sdiv"
    | .udiv   => "udiv"
  Repr.addAppParen (Format.group (Format.nest
    (if prec >= max_prec then 1 else 2) f!"InstCombine.MOp.BinaryOp.{op}"))
    prec
```

Finally, and the reason I do consider this one a blocker for merging: your current `Repr` instance completely ignores flags, all four of the following lines print exactly the same (`InstCombine.MOp.BinaryOp.add`)
```
#eval repr (MOp.BinaryOp.add ⟨false, false⟩)
#eval repr (MOp.BinaryOp.add ⟨false, true⟩)
#eval repr (MOp.BinaryOp.add ⟨true, true⟩)
#eval repr (MOp.BinaryOp.add ⟨true, false⟩)
```

SSA/Projects/InstCombine/Base.lean
```diff
@@ -127,7 +165,7 @@ namespace MOp
 @[match_pattern] def ashr   (w : Width φ) : MOp φ := .binary w .ashr
 @[match_pattern] def urem   (w : Width φ) : MOp φ := .binary w .urem
 @[match_pattern] def srem   (w : Width φ) : MOp φ := .binary w .srem
-@[match_pattern] def add    (w : Width φ) : MOp φ := .binary w .add
+@[match_pattern] def add    (w : Width φ) (additionFlags: AdditionFlags := {nsw := false , nuw := false}) : MOp φ := .binary w (.add  additionFlags )
```
alexkeizer: line is too long

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -77,11 +77,33 @@ def add? {w : Nat} (x y : BitVec w) : IntW w :=
 @[simp_llvm_option]
 theorem add?_eq : LLVM.add? a b  = .some (BitVec.add a b) := rfl
 
+structure AdditionFlags where
+  nsw : Bool := false
+  nuw : Bool := false
+  deriving DecidableEq
+
+/-- Does the signed addition x + y overflow? -/
+def AddSignedWraps? (x y : BitVec w) : Bool := (x.toInt + y.toInt) < -(2^(w-1)) ∨ (x.toInt + y.toInt) ≥ 2^w
```
alexkeizer: Please have a look at https://leanprover-community.github.io/contribute/naming.html;
as a `Bool`-valued `def` this should be `lowerCamelCase`d. 

However, this really should be a `Prop` instead. In fact, if you `#print` the def you'll notice the rhs is actually a Prop and it's being coerced into a `Bool` by `decide`

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -77,11 +77,33 @@ def add? {w : Nat} (x y : BitVec w) : IntW w :=
...
+
 @[simp_llvm_option]
-def add {w : Nat} (x y : IntW w) : IntW w := do
+def add {w : Nat} (x y : IntW w) (flags : AdditionFlags := {nsw := false , nuw := false}) : IntW w := do
   let x' ← x
   let y' ← y
-  add? x' y'
+  let nsw := flags.nsw
+  let nuw := flags.nuw
+  if (nsw ∧  AddSignedWraps? x' y') ∨ (nuw ∧  AddUnSignedWraps? x' y') then
```
alexkeizer: You still have spurious extra spaces here

SSA/Projects/InstCombine/TacticAuto.lean
```diff
@@ -59,6 +60,10 @@ macro_rules
 macro "simp_alive_undef" : tactic =>
   `(tactic|
       (
+        -- Since I introduced the NSW and NUW flags to add, I need to remove them in the case where they are both
+        -- the default value of false.
+        -- I don't know why Lean isn't smart enough to see that False ∧ x = False
+        try simp only [LLVM.add_reduce]
```
alexkeizer: Could you make an [MWE](https://leanprover-community.github.io/mwe.html) that describes the problem you ran into? [`false_and`](https://leanprover-community.github.io/mathlib4_docs/Init/SimpLemmas.html#false_and) exists as a simp-lemma, so Lean should definitely do the simplification you describe.

alexkeizer (changes requested):
I'd say this PR is not ready to be merged yet, some of my earlier comments still stand

SSA/Core/MLIRSyntax/GenericParser.lean
```diff
@@ -551,9 +551,37 @@ declare_syntax_cat mlir_attr_val_symbol
...
 syntax "@" str : mlir_attr_val_symbol
 syntax "#" ident : mlir_attr_val -- alias
-syntax "#" strLit : mlir_attr_val -- aliass
+syntax "#" strLit : mlir_attr_val -- alias
 
-syntax "#" ident "<" strLit ">" : mlir_attr_val -- opaqueAttr
+declare_syntax_cat dialect_attribute_contents
+syntax mlir_attr_val : dialect_attribute_contents
+/--
+I got this from https://mlir.llvm.org/docs/LangRef/
```
alexkeizer:
```suggestion
Following https://mlir.llvm.org/docs/LangRef/, we define a `dialect-attribute`, 
which is a particular case of an `mlir-attr-val` that is namespaced to a particular dialect
```

SSA/Projects/InstCombine/Base.lean
```diff
@@ -96,12 +96,41 @@ inductive MOp.BinaryOp : Type
...
+    | .and    => "and"
+    | .or     => "or"
+    | .xor    => "xor"
+    | .shl    => "shl"
+    | .lshr   => "lshr"
+    | .ashr   => "ashr"
+    | .urem   => "urem"
+    | .srem   => "srem"
+    | .add ⟨false, false⟩ => "add"
+    | .add ⟨nsw, nuw ⟩    => toString f!"add {nsw} {nuw}"
```
alexkeizer:
```suggestion
    | .add ⟨nsw, nuw⟩    => toString f!"add {nsw} {nuw}"
```
Please be more careful with your whitespace!

alexkeizer (approved):
Thanks for addressing my comments @AtticusKuhn, I think this is good to merge now!

Nevertheless, some general remarks: I like your code style generally, but I find you tend to be a bit sloppy when it comes to whitespace (both in terms of putting newlines in between definitions, and where to put spaces inside a line). I appreciate that there is no clear style guide and things are quite vague, but do try to give your diff a thorough read yourself to catch things that are obviously wrong! It might also help to ask @Equilibris to give your code a look for general style, before waiting on me. I've left one suggestion above, please do accept that before merging.

A minor nit: comments with `/-- ... -/` are generally intended for so-called docstrings. That is, documentation of a function that will show up when you hover over a call of that function. You have some comments of this form which talk about the position of the function in the file, which is not really interesting outside of the context of that file, and should probably be a `--` or `/- ... -/` comment instead (neither of which will show up as documentation).

This last point is an exceedingly nitpicky comment, though. Feel free to merge as-is, or to address the points yourself and merge afterwards

alexkeizer:
@bollu I think you also need to "approve these changes" to unblock this PR

AtticusKuhn:
> Nevertheless, some general remarks: I like your code style generally, but I find you tend to be a bit sloppy when it comes to whitespace (both in terms of putting newlines in between definitions, and where to put spaces inside a line). I appreciate that there is no clear style guide and things are quite vague, but do try to give your diff a thorough read yourself to catch things that are obviously wrong! It might also help to ask @Equilibris to give your code a look for general style, before waiting on me. I've left one suggestion above, please do accept that before merging.


Thank you for your review. I'll be sure to be more attentive in the future.

tobiasgrosser:
Great to see this merged. 🎉

PR #494, opened 2024-07-29
chore: avoid non-terminal simp

tobiasgrosser:


alexkeizer:
@AtticusKuhn I think I've seen you merge a non-terminal simp in some PR. Not sure if Tobias or Sid already explained what we mean by non-terminal simp, but if not: see https://leanprover-community.github.io/extras/simp.html#non-terminal-simps for more details

alexkeizer:
On another note: I think mathlib has lints for non-terminal simp, we could investigate if we can ~steal~ reuse that linter for our CI

PR #510, opened 2024-08-08
Chore: tactic alive_auto

AnotherAlexHere:
Changed all 
```lean
sorry
```
 to 
```lean
try alive_auto
all_goals sorry
```


SSA/Projects/InstCombine/HackersDelight.lean
```diff
@@ -19,11 +23,14 @@ theorem not_eq_neg_sub_one :
 
 theorem not_and_eq_not_or_not :
     ~~~ (x &&& y) = ~~~ x ||| ~~~ y := by
-  sorry
+  try alive_auto
+  all_goals try alive_auto
+  all_goals sorry
```
AtticusKuhn:
```suggestion
  try alive_auto
  all_goals sorry
```

AtticusKuhn (approved):
LGTM

AtticusKuhn:
How many theorems does `alive_auto` solve?

Equilibris (commented):
LGTM, but at the same time i think a discussion should be opened on whether this is a good idea in general when some of the theorems don't solve by the tactics as the complexity is $\mathcal O (n^2)$

Equilibris (approved):

AtticusKuhn:
> LGTM, but at the same time i think a discussion should be opened on whether this is a good idea in general when some of the theorems don't solve by the tactics as the complexity is O ( n 2 )

This is a good observation. Maybe we could break up this file into smaller files to speed this up?

PR #532, opened 2024-08-13
chore: update to lean4:v4.11.0-rc2

tobiasgrosser:
Lean is now actively complaining about variable definitions that change theorem statements by only being references in proofs. We revert to the old behavior using `set_option deprecated.oldSectionVars true` to keep this PR small. We should subsequently refactor our code to effectively address the warnings about variable definitions.


SSA/Core/Framework.lean
```diff
@@ -522,24 +522,23 @@ end Lemmas
 ## `denote`
 Denote expressions, programs, and sequences of lets
 -/
-variable [TyDenote d.Ty] [DialectDenote d] [DecidableEq d.Ty] [Monad d.m] [LawfulMonad d.m]
```
goens: the change here seems very strong, it seems like we're using the typeclasses `TyDenote`, `Monad` and `DialectDenote` essentially everywhere  - shouldn't these stay as variables here? We can use sections to delimit these if there are some theorems that don't depend on them, but at least from a first glance I couldn't find examples, and even if there was a few, I think we'd still prefer manually excluding those cases since these typeclasses are so ubiquitous.

SSA/Core/Framework.lean
```diff
@@ -616,6 +615,9 @@ open EffectKind (liftEffect)
 /- Equation lemma to unfold `denote`, which does not unfold correctly due to the presence
   of the coercion `ty_eq` and the mutual definition. -/
 theorem Expr.denote_unfold
+    [TyDenote d.Ty]
```
goens: why are these all in a separate line, in contrast to the others?

SSA/Core/Framework.lean
```diff
@@ -1315,7 +1330,7 @@ theorem Lets.denote_getPureExprAux [LawfulMonad d.m] {Γ₁ Γ₂ : Ctxt d.Ty} {
         Valuation.snoc_last]
       cases eff <;> simp
 
-theorem Lets.denote_getExpr [LawfulMonad d.m] {Γ₁ Γ₂ : Ctxt d.Ty}
+theorem Lets.denote_getExpr [TyDenote d.Ty] [Monad d.m] [DialectDenote d] [LawfulMonad d.m] {Γ₁ Γ₂ : Ctxt d.Ty}
```
goens: This being the last theorem with these typeclasses would make a great end of the `section` that has these typeclasses, no?

SSA/Projects/FullyHomomorphicEncryption/Basic.lean
```diff
@@ -52,8 +52,11 @@ noncomputable def f : (ZMod q)[X] := X^(2^n) + 1
 --   done
 
 def ZMod.toInt (x : ZMod q) : Int := ZMod.cast x
+
+variable [ hqgt1 : Fact (q > 1)] in
```
goens: why is this used as a scoped variable? we can just include the assumption directly right (before the `:`) below

SSA/Projects/FullyHomomorphicEncryption/Basic.lean
```diff
@@ -632,7 +646,7 @@ theorem R.representative_fromTensor_eq_fromTensor' (tensor : List Int) :
     R.representative q n (R.fromTensor tensor) =
       R.representative' q n (R.fromTensor' q tensor)  %ₘ (f q n) := by
   simp only [representative]
-  rw [fromTensor_eq_fromTensor'_fromPoly];
+  simp [fromTensor_eq_fromTensor'_fromPoly ];
```
goens: interesting, what happened here, did `rw` break? do you know why? (and why is there an extra whitespace at the end of the theorem name)?

SSA/Projects/FullyHomomorphicEncryption/Basic.lean
```diff
@@ -31,7 +31,7 @@ open Polynomial -- for R[X] notation
 We assume that `q > 1` is a natural number (not necessarily a prime) and `n` is a natural number.
 We will use these to build `ℤ/qℤ[X]/(X^(2^n) + 1)`
 -/
-variable (q t : Nat) [ hqgt1 : Fact (q > 1)] (n : Nat)
```
goens: it looks like this assumption `hqgt1` is essentially used everywhere - why do we need to factor it out?

goens (commented):
It seems this PR is essentially getting rid of most variables, and then including them individually in every theorem below. Is this what we want? I would have thought we want to organize our theorems well in sections with variables, and only add explicit typeclass annotations very sparingly when we need them as a one-off (e.g. the `LawfulMonad` instances in `Framework.lean`)

SSA/Projects/DCE/DCE.lean
```diff
@@ -225,7 +224,7 @@ def arglistDeleteVar? {Γ: Ctxt d.Ty} {delv : Γ.Var α} {Γ' : Ctxt d.Ty} {ts :
         ⟩
 
 /- Try to delete a variable from an Expr -/
-def Expr.deleteVar? (DEL : Deleted Γ delv Γ') (e: Expr d Γ .pure t) :
+def Expr.deleteVar? [DialectSignature d] [DialectDenote d] [Monad d.m] (DEL : Deleted Γ delv Γ') (e: Expr d Γ .pure t) :
```
AtticusKuhn: Are all these changes really neccessary? I think they make the code more verbose. I added back the instances to the variables declaration and removed them from this definition, and there was no error when I ran it locally.

As far as I understand it, definitions were not affected by the change to "variables", but perhaps I am mistaken on this.

tobiasgrosser:
Thank you, @goens and @AtticusKuhn, I avoided most changes to keep this PR small. We can then open subsequent PRs to actually address the semantic changes.

tobiasgrosser:
The first two cleanup PRs are https://github.com/opencompl/lean-mlir/pull/538 and https://github.com/opencompl/lean-mlir/pull/539.

PR #537, opened 2024-08-15
Chore: prove that addition on BitStreams corresponds to addition on BitVectors

AtticusKuhn:
There are currently two sorries in the file `BitStream.lean`. This proof removes one sorry, in the case of addition by proving that addition on BitVectors agrees with addition on BitStreams up to the first w bits.


SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -373,12 +373,41 @@ variable {w : Nat} {x y : BitVec w} {a b a' b' : BitStream}
...
+    (decide p).xor (decide q) = decide (p ≠ q) := by
+  cases dp
+  <;> (rename_i pt ; simp [pt])
+
 theorem ofBitVec_add : ofBitVec (x + y) ≈ʷ (ofBitVec x) + (ofBitVec y)  := by
-  sorry
+  intros n a
+  have add_lemma : ⟨BitVec.carry (n + 1) x y false , (x + y).getLsb n⟩ = (ofBitVec x).addAux (ofBitVec y) n := by
+    induction' n with n ih
+    · simp [addAux, BitVec.adcb, ofBitVec,a, BitVec.getLsb, Nat.two_mod_mod, BitVec.carry, ← Bool.decide_and, decide_eq_decide, xor_decide, Nat.parity_and x.toNat y.toNat, Nat.add_odd_iff_neq x.toNat y.toNat]
```
Equilibris: Should maybe some of these be made to be simp lemmas if they are all required like this?

SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -373,12 +373,41 @@ variable {w : Nat} {x y : BitVec w} {a b a' b' : BitStream}
...
+    2 ≤ n % 2 + m % 2 ↔ n % 2 = 1 ∧ m % 2 = 1 := by
+  omega
+
+theorem Nat.add_odd_iff_neq (n m : Nat) :
+    (n + m) % 2 = 1 ↔ (n % 2 = 1) ≠ (m % 2 = 1) := by
+  cases Nat.mod_two_eq_zero_or_one n
+  <;> cases Nat.mod_two_eq_zero_or_one m
+  <;> (
+    rename_i mparity nparity
+    simp [mparity, nparity, Nat.add_mod]
```
Equilibris: I would prefer `simp_all [Nat.add_mod]` than a `rename_i` then a `simp`

Equilibris (approved):
LGTM, only minor changes to proofs and sub looks like a simple thing to prove.

SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -373,12 +373,41 @@ variable {w : Nat} {x y : BitVec w} {a b a' b' : BitStream}
 
 local infix:20 " ≈ʷ " => EqualUpTo w
 
--- TODO: These sorries are difficult, and will be proven in a later Pull Request.
+-- TODO: This sorry is difficult, and will be proven in a later Pull Request.
 theorem ofBitVec_sub : ofBitVec (x - y) ≈ʷ (ofBitVec x) - (ofBitVec y)  := by
   sorry
 
+theorem Nat.two_mod_mod (x : Nat) (w : Nat) (_ : 0 < w) : x % 2 ^ w % 2 = x % 2 := by
```
alexkeizer:
```suggestion
theorem Nat.mod_two_pow_mod_two (x : Nat) (w : Nat) (_ : 0 < w) : x % 2 ^ w % 2 = x % 2 := by
```

This, and the other `Nat` lemmas, should be moved to a separate file, maybe `SSA/Experimental/Bits/ForLean.lean` to mark them as candidates for upstreaming

SSA/Experimental/Bits/Fast/BitStream.lean
alexkeizer: Note that `simp[*, Nat.add_mod]` also exists (`simp_all` is more like `simp[*] at *`, i.e., the former will use all local hypotheses to simplify the goal only, while the latter will simplify everywhere).

I agree that `rename_i` is a bit ugly, and brittle!

SSA/Experimental/Bits/Fast/BitStream.lean
alexkeizer: Or, you could make simp lemmas for applied versions of some of the things you are unfolding (if there is an "obviously good" simplfication to be made)

AtticusKuhn:
Thanks for your feedback.

SSA/Experimental/Bits/Fast/ForLean.lean
```diff
@@ -0,0 +1,28 @@
+import Mathlib.Algebra.Group.Defs
```
tobiasgrosser: Can we move this into SSA/Projects/InstCombine/ForLean.lean instead?

AtticusKuhn: Yes, that sounds like a good idea.

tobiasgrosser (approved):
Otherwise, this looks good-to-go.

SSA/Experimental/Bits/Fast/ForLean.lean
tobiasgrosser: Also, do we need these mathlib includes? (its ok if we do)

AtticusKuhn:
Great, thanks for your feedback.

SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -373,12 +374,22 @@ variable {w : Nat} {x y : BitVec w} {a b a' b' : BitStream}
 
 local infix:20 " ≈ʷ " => EqualUpTo w
 
--- TODO: These sorries are difficult, and will be proven in a later Pull Request.
+-- TODO: This sorry is difficult, and will be proven in a later Pull Request.
 theorem ofBitVec_sub : ofBitVec (x - y) ≈ʷ (ofBitVec x) - (ofBitVec y)  := by
   sorry
```
alexkeizer, line 380: By the way, I believe this one should not actually be that hard either. Similar to `neg`, we can in fact reduce this fairly easily to known facts:
```lean
theorem ofBitVec_sub : ofBitVec (x - y) ≈ʷ (ofBitVec x) - (ofBitVec y)  := by
  have : x - y = x + ~~~y + 1#w := by
    simp only [BitVec.sub_toAdd, BitVec.neg_eq_not_add, BitVec.ofNat_eq_ofNat]
    ac_rfl
  calc
  _ ≈ʷ ofBitVec (x + ~~~y + 1#w)            := by rw [this]
  _ ≈ʷ ofBitVec _ + ofBitVec _              := ofBitVec_add
  ...
```

I tried stitching the proof together and failed, but I believe @AtticusKuhn might know how to get the proper congruences to make this all go through (it sure would be nice to have generalized rewriting!)

AtticusKuhn: You are correct, proving "ofBitVec_sub" is not very difficult:

```lean
theorem ofBitVec_sub : ofBitVec (x - y) ≈ʷ (ofBitVec x) - (ofBitVec y)  := by
  calc
  _ ≈ʷ (@ofBitVec w (x + (- y))) := by rw [BitVec.sub_add_neg]
  _ ≈ʷ ((@ofBitVec w x) + (@ofBitVec w (-y))) := ofBitVec_add
  _ ≈ʷ ((@ofBitVec w x) + (- @ofBitVec w (y))) := add_congr equal_up_to_refl ofBitVec_neg
  _ ≈ʷ (ofBitVec x) - (ofBitVec (y)) := by rw [← sub_add_neg]
```

However, I find proving "a - b = a + (-b) to be significantly more difficult:


```lean
theorem sub_add_neg' {a b : BitStream} : a - b = a + (-b) := by
  have sub_add_lemma (i : Nat) :
      let y := b.negAux
      let x := a.addAux (fun n => (y n).1) i
      a.subAux b i = ⟨x.2, !(y i).2 &&  !x.1 ⟩  := by
    induction' i with i ih
    · simp [subAux,addAux,negAux, BitVec.adcb]
      cases (b 0 ) ;
      · simp
      · simp
    · simp [subAux,addAux,negAux, BitVec.adcb]
      rw [ih]
      simp
      cases a (i + 1)
      all_goals cases b (i + 1)
      all_goals simp
      all_goals constructor
      all_goals sorry
  ext i
  simp only [HAdd.hAdd, HSub.hSub, Neg.neg, Sub.sub, BitStream.sub,Add.add, BitStream.add]
  unfold neg
  simp [sub_add_lemma i]
```

tobiasgrosser: Could you push a draft PR such that we see where things become difficult?

AtticusKuhn: > Could you push a draft PR such that we see where things become difficult?

That sounds like a good idea, I've done that here:
https://github.com/opencompl/lean-mlir/pull/554/files

PR #538, opened 2024-08-16
chore: drop use of oldSectionVars from ErasedContext

tobiasgrosser:


AtticusKuhn (approved):
Looks good to me. I like how this is a small and atomic PR, so it is easy to review. I am curious, however, why "[TyDenote Ty]" was included before, if it wasn't necessary?

tobiasgrosser:
> Looks good to me. I like how this is a small and atomic PR, so it is easy to review. I am curious, however, why "[TyDenote Ty]" was included before, if it wasn't necessary?

I think these were already unnecessary, but we overlooked them as lean was not warning about them. So this is a strict cleanup. I updated the commit message.

goens:
that's a good question, had a look into this. If you hover over the types, they still have the dependency `Ctxt.Valuation.ofPair_fst {Ty : Type} [TyDenote Ty] {t₁ t₂ : Ty} (v₁ : ⟦t₁⟧) (v₂ : ⟦t₂⟧) : ofPair v₁ v₂ ⟨0, ⋯⟩ = v₁` which makes a lot of sense, since we need that so that we can write `(v₁ : ⟦t₁⟧) (v₂ : ⟦t₂⟧)`. I think the linter there is actually complaining about the section variable above (in L252). The question is whether we want to have everything in this section depend on `TyDenote`  (and in this case we do because `Valuation` needs it).

Long story short, @tobiasgrosser is right that these were redundant and it's just a cleanup, it's interesting that the linter didn't flag this for `Valuation.ofPair` (I just pushed another commit cleaning that up)

tobiasgrosser:
> that's a good question, had a look into this. If you hover over the types, they still have the dependency `Ctxt.Valuation.ofPair_fst {Ty : Type} [TyDenote Ty] {t₁ t₂ : Ty} (v₁ : ⟦t₁⟧) (v₂ : ⟦t₂⟧) : ofPair v₁ v₂ ⟨0, ⋯⟩ = v₁` which makes a lot of sense, since we need that so that we can write `(v₁ : ⟦t₁⟧) (v₂ : ⟦t₂⟧)`. I think the linter there is actually complaining about the section variable above (in L252). The question is whether we want to have everything in this section depend on `TyDenote` (and in this case we do because `Valuation` needs it).
> 
> Long story short, @tobiasgrosser is right that these were redundant and it's just a cleanup, it's interesting that the linter didn't flag this for `Valuation.ofPair` (I just pushed another commit cleaning that up)

Thank you. This is super useful. Do you feel this PR can be merged from your side?

goens:
yep, LGTM!

alexkeizer:
> that's a good question, had a look into this. If you hover over the types, they still have the dependency `Ctxt.Valuation.ofPair_fst {Ty : Type} [TyDenote Ty] {t₁ t₂ : Ty} (v₁ : ⟦t₁⟧) (v₂ : ⟦t₂⟧) : ofPair v₁ v₂ ⟨0, ⋯⟩ = v₁` which makes a lot of sense, since we need that so that we can write `(v₁ : ⟦t₁⟧) (v₂ : ⟦t₂⟧)`. I think the linter there is actually complaining about the section variable above (in L252). The question is whether we want to have everything in this section depend on `TyDenote` (and in this case we do because `Valuation` needs it).
> 
> Long story short, @tobiasgrosser is right that these were redundant and it's just a cleanup, it's interesting that the linter didn't flag this for `Valuation.ofPair` (I just pushed another commit cleaning that up)

AFAIU, `[TyDenote]` would still only be included for those theorems that actually need it, it's just that this *need* is now determined solely by the theorem statement and no longer influenced by the proof.
That is, having the variable floating around should not imply that everything in the section automatically gets the extra assumption, only those whose theorem statement mentions `tyDenote` should get it, right?

(Caveat: this is my reading of the changelog, I haven't played around with it myself. Do tell me if I misunderstood)

PR #539, opened 2024-08-16
chore: drop use of oldSectionVars from DCE

tobiasgrosser:
We move `Deleted.pushforward_Valuation_snoc` upwards and split the variable declaration as `Deleted.pushforward_Valuation_snoc` does not depend on `[DialectSignature d] [DialectDenote d] [DecidableEq d.Ty] [Monad d.m] [LawfulMonad d.m]`.


goens:
there was a few more unneeded typeclasses so I cleaned it up a bit more

PR #540, opened 2024-08-16
chore: remove oldSectionVars in FHE

goens:
introduces sections more carefully to manage typeclasses in FHE, and gets rid of oldSectionVars this way

### DIFF
--- a/SSA/Projects/FullyHomomorphicEncryption/Basic.lean
+++ b/SSA/Projects/FullyHomomorphicEncryption/Basic.lean
@@ -515,7 +515,7 @@ variable {q n : Nat}
 /-- an equivalent implementation of `fromTensor` that uses `Finsupp`
   to enable reasoning about values using mathlib's notions of
   support, coefficients, etc. -/
-noncomputable def R.fromTensorFinsupp (coeffs : List Int) : (ZMod q)[X] :=
+noncomputable def R.fromTensorFinsupp (q : Nat) (coeffs : List Int) : (ZMod q)[X] :=
   Polynomial.ofFinsupp (List.toFinsupp (coeffs.map Int.cast))
 
 theorem Polynomial.degree_toFinsupp [Semiring M] [DecidableEq M]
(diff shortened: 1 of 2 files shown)


tobiasgrosser:
This looks very nice. @bollu, a review would be much appreciated.

SSA/Projects/FullyHomomorphicEncryption/Basic.lean
```diff
@@ -607,15 +623,15 @@ theorem R.coeff_fromPoly {q n : Nat} [Fact (q > 1)] (p : (ZMod q)[X]) :
   3.
 -/
 /-- The coefficient of `fromTensor` is the same as the values available in the tensor input. -/
-theorem R.coeff_fromTensor [hqgt1 : Fact (q > 1)] (tensor : List Int)
+theorem R.coeff_fromTensor (tensor : List Int)
     (htensorlen : tensor.length < 2^n) :
     (R.fromTensor (q := q) (n := n) tensor).coeff i = (tensor.getD i 0) := by
   rw [fromTensor_eq_fromTensorFinsupp_fromPoly]
-  have hfromTensorFinsuppDegree := fromTensorFinsupp_degree q tensor
+  have hfromTensorFinsuppDegree := fromTensorFinsupp_degree (q:=q) tensor
```
bollu:
```suggestion
  have hfromTensorFinsuppDegree := fromTensorFinsupp_degree (q:=q) tensor
```

The fact that we need to supply the `q` explicitly is probably telling us to make it an explicit parameter. @goens I leave it to you as to whether you want to do this in the same PR. I'd be happy with a follow-up PR that makes the `q` explicit here!

bollu (approved):
LGTM, thanks for the refactor :) I feel we should make `q` explicit in places where we pass it implicitly because it's clear Lean's not able to infer it; I'm happy for that to be in a different PR, or if you want to pass the ball to me to do this refactor!

goens:
@bollu good point! I removed the cases for `fromTensorFinsupp_degree` which I had added in this PR to keep the diff small, we might consider a second pass where we do this for others like `R.fromTensor` itself. The thing that gives me pause there is that sometimes it does work, like on theorem `fromTensor_toTensor`, where it's cleaner with the implicits. I think in any case that one should be a separate PR

PR #551, opened 2024-08-18
test: drop no_index

tobiasgrosser:


tobiasgrosser:
This test is not needed anymore. We have a reduced example.

PR #555, opened 2024-08-19
feat: doz, max and min theorems

AnotherAlexHere:
I added some new theorems on "difference or zero".

### DIFF
--- a/SSA/Projects/InstCombine/HackersDelight.lean
+++ b/SSA/Projects/InstCombine/HackersDelight.lean
@@ -459,6 +459,105 @@ theorem div_overflow_iff_neq_and_RightShift_lt {x y : BitVec 64} {y : BitVec 32}
   try alive_auto
   all_goals sorry
 
+def signedDifferenceOrZero (x y : BitVec w) : BitVec w :=
+    if (x ≥ₛ y) then (x - y) else 0#w
+
+def unsignedDifferenceOrZero (x y : BitVec w) : BitVec w :=
+    if (x ≥ᵤ y) then (x - y) else 0#w
+
+def uMaxBitVec (x y : BitVec w) : BitVec w :=
+  if x ≥ᵤ y then x else y
+
+def uMinBitVec (x y : BitVec w) : BitVec w :=
+  if x ≤ᵤ y then x else y
+
+def sMaxBitVec (x y : BitVec w) : BitVec w :=
+  if x ≥ₛ y then x else y
+
+def sMinBitVec (x y : BitVec w) : BitVec w :=
+  if x ≤ₛ y then x else y
+
+theorem signed_max_eq_add_doz :
+    sMaxBitVec x y = y + signedDifferenceOrZero x y := by
+  try alive_auto
+  all_goals sorry
+
+theorem signed_min_eq_sub_doz :
+    sMinBitVec x y = x - signedDifferenceOrZero x y := by
+  try alive_auto
+  all_goals sorry
+
+theorem unsigned_max_eq_add_dozu :
+    uMaxBitVec x y = y + unsignedDifferenceOrZero x y := by
+  try alive_auto
+  all_goals sorry
+
+theorem unsigned_min_eq_sub_dozu :
+    uMinBitVec x y = x - unsignedDifferenceOrZero x y := by
+  try alive_auto
+  all_goals sorry
+
+def leBitmask (x y : BitVec w) : BitVec w :=
+  if y ≤ x then BitVec.ofNat w (2^w - 1) else BitVec.ofNat w 0
+
+theorem signedDifferenceOrZero_eq_sub_and_bitmask :
+   signedDifferenceOrZero x y = (x - y) &&& leBitmask x y := by
+  try alive_auto
+  all_goals sorry
+
+theorem signed_max_xor_and_bitmask_xor :
+    sMaxBitVec x y = ((x ^^^ y) &&& leBitmask x y) ^^^ y := by
+  try alive_auto
+  all_goals sorry
+
+theorem signed_min_xor_and_bitmask_xor :
+    sMinBitVec x y = ((x ^^^ y) &&& leBitmask y x) ^^^ y := by
+  try alive_auto
+  all_goals sorry
+
+def carryBitmask (x y : BitVec w) : BitVec w :=
+  if BitVec.carry w x ((~~~y + 1) - 1) false then BitVec.ofNat w (2^w - 1) else BitVec.ofNat w 0
+
+theorem unsigned_doz_sub_and_not_carry :
+    unsignedDifferenceOrZero x y = ((x - y) &&& ~~~ carryBitmask x y) := by
+  try alive_auto
+  all_goals sorry
+
+theorem unsigned_max_eq_sub_sub_and_carry :
+    uMaxBitVec x y = x - ((x - y) &&& carryBitmask x y) := by
+  try alive_auto
+  all_goals sorry
+
+theorem unsigned_min_eq_add_sub_and_carry :
+    uMinBitVec x y = y + ((x - y) &&& carryBitmask x y) := by
+  try alive_auto
+  all_goals sorry
+
+theorem signed_doz_and_not_xor_xor_and_and_rightShift {x y : BitVec 32} (h : d = x - y) :
+    signedDifferenceOrZero x y = d &&& (~~~ d ^^^ ((x ^^^ y) &&& (d ^^^ x)) >>> 31) := by
+  try alive_auto
+  all_goals sorry
+
+theorem unsigned_doz_and_not_xor_xor_and_and_rightShift {x y : BitVec 32} (h : d = x - y) :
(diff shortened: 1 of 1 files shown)


SSA/Projects/InstCombine/HackersDelight.lean
```diff
@@ -459,6 +459,105 @@ theorem div_overflow_iff_neq_and_RightShift_lt {x y : BitVec 64} {y : BitVec 32}
...
+  try alive_auto
+  all_goals sorry
+
+theorem unsigned_min_eq_sub_dozu :
+    uMinBitVec x y = x - unsignedDifferenceOrZero x y := by
+  try alive_auto
+  all_goals sorry
+
+def leBitmask (x y : BitVec w) : BitVec w :=
+  if y ≤ x then BitVec.ofNat w (2^w - 1) else BitVec.ofNat w 0
```
AtticusKuhn: This function could probably be written more succinctly as 


```suggestion
def leBitmask (x y : BitVec w) : BitVec w :=
  if y ≤ x then -1#w else 0#w
```

SSA/Projects/InstCombine/HackersDelight.lean
```diff
@@ -459,6 +459,105 @@ theorem div_overflow_iff_neq_and_RightShift_lt {x y : BitVec 64} {y : BitVec 32}
   try alive_auto
   all_goals sorry
 
+def signedDifferenceOrZero (x y : BitVec w) : BitVec w :=
+    if (x ≥ₛ y) then (x - y) else 0#w
```
AtticusKuhn:
```suggestion
    if x ≥ₛ y then x - y else 0#w
```

Unnecessary parentheses

SSA/Projects/InstCombine/HackersDelight.lean
```diff
@@ -459,6 +459,105 @@ theorem div_overflow_iff_neq_and_RightShift_lt {x y : BitVec 64} {y : BitVec 32}
...
+
+def unsignedDifferenceOrZero (x y : BitVec w) : BitVec w :=
+    if (x ≥ᵤ y) then (x - y) else 0#w
```
AtticusKuhn:
```suggestion
    if x ≥ᵤ y then x - y else 0#w
```
unnecessary parentheses

SSA/Projects/InstCombine/HackersDelight.lean
```diff
@@ -459,6 +459,105 @@ theorem div_overflow_iff_neq_and_RightShift_lt {x y : BitVec 64} {y : BitVec 32}
...
+  try alive_auto
+  all_goals sorry
+
+theorem signed_min_xor_and_bitmask_xor :
+    sMinBitVec x y = ((x ^^^ y) &&& leBitmask y x) ^^^ y := by
+  try alive_auto
+  all_goals sorry
+
+def carryBitmask (x y : BitVec w) : BitVec w :=
+  if BitVec.carry w x ((~~~y + 1) - 1) false then BitVec.ofNat w (2^w - 1) else BitVec.ofNat w 0
```
AtticusKuhn: I don't understand this function. 

Why does it have "((~~~y + 1) - 1)"? Isn't that equivalent to "~~~y". Perhaps a doc-string comment would clear this up.

AtticusKuhn: Doc-string comments are of the form

/-- 

-/

SSA/Projects/InstCombine/HackersDelight.lean
```diff
@@ -459,6 +459,105 @@ theorem div_overflow_iff_neq_and_RightShift_lt {x y : BitVec 64} {y : BitVec 32}
...
+    uMaxBitVec x y = x - ((x - y) &&& carryBitmask x y) := by
+  try alive_auto
+  all_goals sorry
+
+theorem unsigned_min_eq_add_sub_and_carry :
+    uMinBitVec x y = y + ((x - y) &&& carryBitmask x y) := by
+  try alive_auto
+  all_goals sorry
+
+theorem signed_doz_and_not_xor_xor_and_and_rightShift {x y : BitVec 32} (h : d = x - y) :
```
AtticusKuhn: Where is d defined? I find it somewhat strange that I do not see any declaration for d.

SSA/Projects/InstCombine/HackersDelight.lean
```diff
@@ -459,6 +459,105 @@ theorem div_overflow_iff_neq_and_RightShift_lt {x y : BitVec 64} {y : BitVec 32}
...
+    signedDifferenceOrZero x y = d &&& (~~~ d ^^^ ((x ^^^ y) &&& (d ^^^ x)) >>> 31) := by
+  try alive_auto
+  all_goals sorry
+
+theorem unsigned_doz_and_not_xor_xor_and_and_rightShift {x y : BitVec 32} (h : d = x - y) :
```
AtticusKuhn: If you think it would make your code cleaner, you may want to consider declaring a new section where x and y are of width 32. Perhaps something like:

```lean
section fixedWidth
variable {x y d : BitVec 32}
```

AtticusKuhn (approved):
Looks good to me.

SSA/Projects/InstCombine/HackersDelight.lean
AnotherAlexHere: Good idea!

AtticusKuhn:
Also, as a general comment, it seems that that HackersDelight.lean file is getting rather long, and this may slow down Lean in type-checking the file (because Lean type-checks files linearly from top to bottom). I would suggest splitting up HackersDelight.lean into several smaller files to increase modularity and hasten type-checking times.

tobiasgrosser:
> Also, as a general comment, it seems that that HackersDelight.lean file is getting rather long, and this may slow down Lean in type-checking the file (because Lean type-checks files linearly from top to bottom). I would suggest splitting up HackersDelight.lean into several smaller files to increase modularity and hasten type-checking times.

I think this is good to keep in mind. For now, I would prefer to keep things in a single file as this makes it easier to keep an overview of what exists.

SSA/Projects/InstCombine/HackersDelight.lean
```diff
@@ -459,6 +459,150 @@ theorem div_overflow_iff_neq_and_RightShift_lt {x y : BitVec 64} {y : BitVec 32}
...
+  try alive_auto
+  all_goals sorry
+
+theorem signed_min_eq_add_sub_and_sub_rightShift :
+    signedMinBitVec x y = y + ((x - y) &&& ((x - y) >>> 31)) := by
+  try alive_auto
+  all_goals sorry
+
+theorem lt_sdoz_or_neg_sdoz :
+    (y <ₛ x) ↔ (signedDifferenceOrZero x y ||| -(signedDifferenceOrZero x y)).getMsb 32 := by
```
AtticusKuhn: I don't know what the statement of the theorem in Hacker's Delight is, but I think it is probable that you want  "BitVec.msb" rather than "BitVec.getMsb 32"

SSA/Projects/InstCombine/HackersDelight.lean
```diff
@@ -459,6 +459,150 @@ theorem div_overflow_iff_neq_and_RightShift_lt {x y : BitVec 64} {y : BitVec 32}
...
+    if x ≥ᵤ y then x - y else 0#w
+
+def unsignedMaxBitVec (x y : BitVec w) : BitVec w :=
+  if x ≥ᵤ y then x else y
+
+def unsignedMinBitVec (x y : BitVec w) : BitVec w :=
+  if x ≤ᵤ y then x else y
+
+def signedMaxBitVec (x y : BitVec w) : BitVec w :=
+  if x ≥ₛ y then x else y
```
AtticusKuhn: Make the indentation between these definitions line up.

SSA/Projects/InstCombine/HackersDelight.lean
```diff
@@ -459,6 +459,150 @@ theorem div_overflow_iff_neq_and_RightShift_lt {x y : BitVec 64} {y : BitVec 32}
...
+def unsignedDifferenceOrZero (x y : BitVec w) : BitVec w :=
+    if x ≥ᵤ y then x - y else 0#w
+
+def unsignedMaxBitVec (x y : BitVec w) : BitVec w :=
+  if x ≥ᵤ y then x else y
+
+def unsignedMinBitVec (x y : BitVec w) : BitVec w :=
+  if x ≤ᵤ y then x else y
+
+def signedMaxBitVec (x y : BitVec w) : BitVec w :=
```
AtticusKuhn, line 474: It's up to you, but maybe I would name this definition something like "BitVec.signedMax".

AtticusKuhn (approved):
Looks good to me.

SSA/Projects/InstCombine/HackersDelight.lean
AtticusKuhn: See 
https://leanprover-community.github.io/mathlib4_docs/Init/Data/BitVec/Basic.html#BitVec.msb

PR #554, opened 2024-08-19
Chore: Prove that subtraction on BitVectors is the same as subtraction on BitStreams

AtticusKuhn:
If this PR is successful, it will remove the last sorry from "BitStream.lean".

There is one difficulty, however, and that is that I am not sure how to prove that "a - b = a + (-b)" on bitstreams. For this reason, I am making this a draft PR until I figure out how to prove that. My only idea is to come up with some auxiliary function "f" such that

```lean

/--
g is some unknown auxiliary function that will be useful in proving sub_add_neg
-/
def g (a b : BitStream) (i : Nat) : Bool := sorry

theorem g_zero {a b : BitStream} :
    (!a 0 && b 0) = a.g b 0 := by
  sorry

theorem g_succ_left {a b : BitStream} (i : ℕ) :
    xor (b (i + 1)) (a.g b i) = ((!b (i + 1)) != ((b.negAux i).2 != (a.addAux (fun n => (b.negAux n).1) i).2)) := by
  sorry

theorem g_succ_right {a b : BitStream} (i : ℕ)  :
    (!a (i + 1) && b (i + 1) || !xor (a (i + 1)) (b (i + 1)) && a.g b i) = a.g b (i + 1) := by
  sorry
```

PR #556, opened 2024-08-20
[NFC] Chore: Canonicalise Bits of BitStream.addAux to match subAux and negAux

AtticusKuhn:
Before, there was an in-congruity where subAux and negAux would return the result bit on the left and the carry bit on the right.
However, addAux was defined in such a way so that the result bit was on the right and the carry bit is on the left.

This PR swaps the bits for addAux so that addAux matches the convention of subAux and negAux.

This is a non-functional refactor. It does not add any new features.

### DIFF
--- a/SSA/Experimental/Bits/Fast/BitStream.lean
+++ b/SSA/Experimental/Bits/Fast/BitStream.lean
@@ -444,13 +444,6 @@ theorem ofBitVec_not_eqTo : ofBitVec (~~~ x) ≈ʷ ~~~ ofBitVec x := by
   intros _ a
   simp [ofBitVec, a]
 
-/--
-  Clarification: the reason that this theorem (unexpectedly) contains Prod.swap is
-  that in BitStream.negAux, the carry bit is in the left bit of the pair and
-  in BitStream.addAux, the carry bit is in the right bit of the pair.
-
-  TODO: Possibly investigate changing the behavior of BitStream.addAux?
--/
 theorem negAux_eq_not_addAux : a.negAux = (~~~a).addAux 1 := by
   funext i
   induction' i with _ ih
--- a/SSA/Experimental/Bits/Fast/BitStream.lean
+++ b/SSA/Experimental/Bits/Fast/BitStream.lean
@@ -272,39 +272,11 @@ end BitwiseOps
 /-! # Addition, Subtraction, Negation -/
 section Arith
 
-/--
-addAux' is the old version of addAux, which is kept around for backward-compatability reasons.
-To use the new version, use addAux
--/
--- def addAux' (x y : BitStream) : Nat → Bool × Bool
---   | 0 => BitVec.adcb (x 0) (y 0) false
---   | n+1 =>
---     let carry := (addAux' x y n).1
---     let a := x (n + 1)
---     let b := y (n + 1)
---     BitVec.adcb a b carry
 def addAux (x y : BitStream) (i : Nat) :  Bool × Bool :=
   let carry : Bool := match i with
     | 0 => false
     | i + 1 => (addAux x y i).2
   Prod.swap (BitVec.adcb (x i) (y i) carry)
-  -- | 0 => Prod.swap (BitVec.adcb (x 0) (y 0) false)
-  -- | n+1 =>
-  --   let carry := (addAux' x y n).1
-  --   let a := x (n + 1)
-  --   let b := y (n + 1)
-  --   Prod.swap (BitVec.adcb a b carry)
-/--
-The reason that there is a Prod.swap in the definition of addAux is that
-BitVec.adcb returns the carry bit on the left and the result bit on the right.
-
-In order to preserve the same design as subAux and negAux, we use Prod.swap
-so that the result bit is on the left and the carry bit is on the right.
-
-The un-swapped version is still availiable as addAux'
--/
--- @[simp]
--- def addAux (x y : BitStream) : Nat →  Bool × Bool := Prod.swap ∘ (addAux' x y)
 
 def add (x y : BitStream) : BitStream :=
   fun n => (addAux x y n).1
@@ -458,8 +430,8 @@ theorem ofBitVec_not_eqTo : ofBitVec (~~~ x) ≈ʷ ~~~ ofBitVec x := by
 theorem negAux_eq_not_addAux : a.negAux = (~~~a).addAux 1 := by
   funext i
   induction' i with _ ih
-  · simp [negAux, BitVec.adcb, OfNat.ofNat, ofNat, addAux]
-  · simp [negAux, BitVec.adcb, OfNat.ofNat, ofNat, addAux, ih]
+  · simp [negAux, addAux, BitVec.adcb, OfNat.ofNat, ofNat]
+  · simp [negAux, addAux, BitVec.adcb, OfNat.ofNat, ofNat, ih]
 
 theorem neg_eq_not_add : - a = ~~~ a + 1 := by
   ext _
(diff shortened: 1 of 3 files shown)


tobiasgrosser:
Nice. Can we get a quick review from @Equilibris?

SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -272,16 +272,31 @@ end BitwiseOps
 /-! # Addition, Subtraction, Negation -/
 section Arith
 
-def addAux (x y : BitStream) : Nat → Bool × Bool
+def addAux' (x y : BitStream) : Nat → Bool × Bool
```
tobiasgrosser: Why is this function renamed?

SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -272,16 +272,31 @@ end BitwiseOps
...
+The reason that there is a Prod.swap in the definition of addAux is that
+BitVec.adcb returns the carry bit on the left and the result bit on the right.
+
+In order to preserve the same design as subAux and negAux, we use Prod.swap
+so that the result bit is on the left and the carry bit is on the right.
+
+The un-swapped version is still availiable as addAux'
+-/
+@[simp]
+def addAux (x y : BitStream) : Nat →  Bool × Bool := Prod.swap ∘ (addAux' x y)
```
tobiasgrosser: Should we drop this?

SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -272,16 +272,31 @@ end BitwiseOps
...
+
+@[simp]
+theorem reduce_addAux : addAux x y = Prod.swap ∘ (addAux' x y) := by rfl
```
tobiasgrosser: Should we drop this as well?

SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -428,18 +444,11 @@ theorem ofBitVec_not_eqTo : ofBitVec (~~~ x) ≈ʷ ~~~ ofBitVec x := by
...
-  TODO: Possibly investigate changing the behavior of BitStream.addAux?
--/
-theorem negAux_eq_not_addAux : a.negAux = Prod.swap ∘ (~~~a).addAux 1 := by
+theorem negAux_eq_not_addAux : a.negAux = (~~~a).addAux 1 := by
   funext i
   induction' i with _ ih
-  · simp [negAux, addAux, BitVec.adcb, OfNat.ofNat, ofNat]
-  · simp [negAux, addAux, BitVec.adcb, OfNat.ofNat, ofNat, ih]
+  · simp [negAux, addAux, BitVec.adcb, OfNat.ofNat, ofNat, Prod.swap, addAux']
+  · simp [negAux, addAux, BitVec.adcb, OfNat.ofNat, ofNat, Prod.swap, addAux', ih]
```
tobiasgrosser: Why are now new swaps needed here? Also, why do we refer to both addAux and AddAux' in the simp lemmas?

SSA/Experimental/Bits/Propagate.lean
```diff
@@ -228,12 +228,19 @@ lemma ls_eq_propagate (b : Bool) (x : BitStream) :
   | 1 => rfl
   | n+2 => simp [propagate_succ, BitStream.concat]
 
+lemma addAux'_eq_propagateCarry (x y : BitStream) (n : ℕ) :
+    (addAux' x y n).1 = propagateCarry (λ _ => false)
+      (λ (carry : Unit → Bool) (bits : Bool → Bool) =>
+        λ _ => (bits true && bits false) || (bits true && carry ()) || (bits false && carry ()) )
+    (λ b => cond b x y) n () := by
+  induction n <;> simp [addAux', BitVec.adcb, Bool.atLeastTwo, *]
```
tobiasgrosser: Why do we have two versions of this one as well?

SSA/Experimental/Bits/Fast/BitStream.lean
AtticusKuhn: I kept around the old version in case anyone still wants to use the old version. The new version is called "addAux"

SSA/Experimental/Bits/Fast/BitStream.lean
AtticusKuhn: I agree, reduce_addAux isn't necessary.

SSA/Experimental/Bits/Fast/BitStream.lean
AtticusKuhn: Good spotting. It turns out that they aren't necessary.

AtticusKuhn:
In general there are two ways we might fix this issue:

option #1:

```lean
/--
addAux' is the old version of addAux, which is kept around for backward-compatability reasons.
To use the new version, use addAux
-/
def addAux' (x y : BitStream) : Nat → Bool × Bool
  | 0 => BitVec.adcb (x 0) (y 0) false
  | n+1 =>
    let carry := (addAux' x y n).1
    let a := x (n + 1)
    let b := y (n + 1)
    BitVec.adcb a b carry

/--
The reason that there is a Prod.swap in the definition of addAux is that
BitVec.adcb returns the carry bit on the left and the result bit on the right.

In order to preserve the same design as subAux and negAux, we use Prod.swap
so that the result bit is on the left and the carry bit is on the right.

The un-swapped version is still availiable as addAux'
-/
@[simp]
def addAux (x y : BitStream) : Nat →  Bool × Bool := Prod.swap ∘ (addAux' x y)
```

or 
option #2: 

```lean
def addAux (x y : BitStream) : Nat → Bool × Bool
  | 0 => Prod.swap (BitVec.adcb (x 0) (y 0) false)
  | n+1 =>
    let carry := (addAux' x y n).1
    let a := x (n + 1)
    let b := y (n + 1)
    Prod.swap (BitVec.adcb a b carry)
```

I have chosen option #1.

tobiasgrosser:
Why would we want to keep around the old version? Is the new version not strictly better? In case it is, I suggest to just drop the old version.

tobiasgrosser:
> The reason that there is a Prod.swap in the definition of addAux is that
BitVec.adcb returns the carry bit on the left and the result bit on the right.

In that case, would it not be better to follow the convention in `adbc` and always return the carry bit on the left in all our implementation? Or is lean inconsistent within itself?

SSA/Experimental/Bits/Propagate.lean
```diff
@@ -229,12 +229,13 @@ lemma ls_eq_propagate (b : Bool) (x : BitStream) :
...
 
 lemma addAux_eq_propagateCarry (x y : BitStream) (n : ℕ) :
-    (addAux x y n).1 = propagateCarry (λ _ => false)
+    (addAux x y n).2 = propagateCarry (λ _ => false)
       (λ (carry : Unit → Bool) (bits : Bool → Bool) =>
         λ _ => (bits true && bits false) || (bits true && carry ()) || (bits false && carry ()) )
     (λ b => cond b x y) n () := by
   induction n <;> simp [addAux, BitVec.adcb, Bool.atLeastTwo, *]
 
+
```
tobiasgrosser: Unrelated.

tobiasgrosser (approved):
I think this is good to go now. If you could rebase your other patch on this one, I can look at the actual problem.

AtticusKuhn:
Great, thanks for your feedback.

AtticusKuhn:
> I think this is good to go now. If you could rebase your other patch on this one, I can look at the actual problem.

Got it. I have done that now:

https://github.com/opencompl/lean-mlir/pull/554/files

PR #559, opened 2024-08-22
chore: prove that subtraction on BitStreams corresponds to subtraction on BitVectors

AtticusKuhn:
This PR proves the key theorem

```lean
theorem ofBitVec_sub : ofBitVec (x - y) ≈ʷ (ofBitVec x) - (ofBitVec y)  := by
  calc
  _ ≈ʷ ofBitVec (x + - y) := by rw [BitVec.sub_add_neg]
  _ ≈ʷ ofBitVec x + ofBitVec (-y) := ofBitVec_add
  _ ≈ʷ ofBitVec x + - ofBitVec y := add_congr equal_up_to_refl ofBitVec_neg
  _ ≈ʷ ofBitVec x - ofBitVec y := by rw [sub_add_neg]
```

Which shows that subtraction on BitStreams corresponds to subtraction on BitVectors.

In doing so, this PR removes the last "sorry" from the file BitStream.lean.

Note that this PR is a continuation of PR
https://github.com/opencompl/lean-mlir/pull/554

PR #562, opened 2024-08-22
chore: move Bool theorems to ForLean

tobiasgrosser:
In the process, we create a section for `Bool.` theorems and fix the naming of the theorems.


tobiasgrosser:
@AtticusKuhn, this is factored out from #559 to remove unrelated cleanups and keep the original PR focused.

PR #563, opened 2024-08-22
chore: drop unnecessary newlines from BitStream.lean

tobiasgrosser:


tobiasgrosser:
Another cleanup PR in preparation for #559

PR #589, opened 2024-09-03
chore: use w instead of w+1 in HackersDelight

tobiasgrosser:


AtticusKuhn:
@tobiasgrosser 
There was a reason that the bit-width was w + 1 instead of w.

See
https://grosser.zulipchat.com/#narrow/stream/446584-Project---Lean4---BitVectors/topic/Hacker's.20Delight.20.26.20Other.20Lemmas/near/463417442

tobiasgrosser:
AFAIU, I set them to `w+1` because the automata tactic could not support it. As your PR fixed this, we are able to drop it again, no?

PR #601, opened 2024-09-08
chore: proof of `bitvec_290__292`

### LINKED ISSUE #598: 1#w <<< b.toNat * a = a <<< b.toNat
A simple example is, bitvec_290__292, where we are left with:

  case some.some.isFalse
  w : ℕ
  a b : BitVec w
  h : ¬w ≤ b.toNat
  ⊢ 1#w <<< b.toNat * a = a <<< b.toNat

luisacicolini:
PR solving [issue 598](https://github.com/opencompl/lean-mlir/issues/598)

### DIFF
--- a/SSA/Projects/InstCombine/AliveStatements.lean
+++ b/SSA/Projects/InstCombine/AliveStatements.lean
@@ -689,8 +689,9 @@ theorem bitvec_290__292 :
   simp_alive_undef
   simp_alive_ops
   simp_alive_case_bash
-  try alive_auto
-  all_goals sorry
+  alive_auto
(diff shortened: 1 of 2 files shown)


tobiasgrosser:
Amazing. I cleaned this up a bit by:

- make this a named theorem and move it into ForLean.lean
- Using `simp` instead of `rw` to shorten the proof and only list the non-default lemmas.
- making it a simp lemma that automatically applies

We try to keep AliveStatements.lean unchanged as it is auto-generated.

luisacicolini:
got it! thank you

PR #626, opened 2024-09-20
chore: update to nightly-testing-2024-09-19

luisacicolini:
It worked without having to change anything, but somehow there is one less alive statement passing


tobiasgrosser:
Nice. Can you check which alive test case this is?

luisacicolini:
yes - so far it's not one of the recent ones and everything works in `ForLean.lean`

PR #647, opened 2024-09-23
chore: fixed `bv_auto` to prove `bv_AndOrXor_794`

luisacicolini:


luisacicolini:
peace of mind is now restored

PR #654, opened 2024-09-25
feat: `DC` dialect operators

luisacicolini:


SSA/Projects/CIRCT/DC/DC.lean
```diff
@@ -0,0 +1,184 @@
...
+  Stream.corec (β := ValueStream α × TokenStream) (x, y) fun ⟨x, y⟩ =>
+    match x 0, y 0 with
+    | some x₀, some _ => (x₀, (x.tail, y.tail))
+    | some _, none => (none, (x, y.tail)) -- wait to sync with the token stream
+    | none, some _ => (none, (x.tail, y)) -- wait to sync with the value stream
+    | none, none => (none, (x.tail, y.tail))
+
+-- weirdly enough the docs says that c must be a Value type (not token),
+-- but I don't see how this would be useful? I left stream bool now, but I don't like it that much
+-- emotionally not a valuestream but can do for now
```
ymherklotz: Hmm the docs say that `c` is a ValueStream (DC.value) right?  Which is what we want for the branch, so I think that the implementation matches the documentation/specification in this case, which is good.  It can't be a TokenStream because otherwise you wouldn't have a condition to branch on.  So I think this implementation is exactly the one specified by the specification.

SSA/Projects/CIRCT/DC/DC.lean
```diff
@@ -0,0 +1,184 @@
...
+def merge (x y : TokenStream) : TokenStream :=
+  Stream.corec (β := TokenStream × TokenStream) (x, y) fun ⟨x, y⟩ =>
+    match x 0, y 0 with
+    | some _, some _ => (some (), (x.tail, y))
+    | some _, none => (some (), (x.tail, y.tail))
+    | none, some _ => (some (), (x.tail, y.tail))
+    | none, none => (none, (x.tail, y.tail))
+
+-- same issue as branch for the condition stream
+def select (x y : TokenStream) (c : Stream Bool): TokenStream :=
```
ymherklotz: I would just change `Stream Bool` to `ValueStream Bool`, but otherwise my comment from above works here too, and I think this matches the documentation actually.

SSA/Projects/CIRCT/DC/DC.lean
```diff
@@ -0,0 +1,184 @@
...
+        (none, (x.tail, y, c))
+      else
+        (some (), (x.tail, y.tail, c))
+    | _, _, none => (none, (x, y, c.tail))
+    | none, none, some _ => (none, (x.tail, y.tail, c))
+
+def sink (x : TokenStream) : TokenStream :=
+  Stream.corec (β := TokenStream) x fun x => (none, x.tail)
+
+def source (a : Unit) : TokenStream :=
```
ymherklotz: I don't think this needs an input `a`, it can just generate an infinite stream of `some ()`.

SSA/Projects/CIRCT/DC/DC.lean
```diff
@@ -0,0 +1,184 @@
...
+    | none, some _, some c₀ =>
+      if c₀ then
+        (none, (x.tail, y, c))
+      else
+        (some (), (x.tail, y.tail, c))
+    | _, _, none => (none, (x, y, c.tail))
+    | none, none, some _ => (none, (x.tail, y.tail, c))
+
+def sink (x : TokenStream) : TokenStream :=
+  Stream.corec (β := TokenStream) x fun x => (none, x.tail)
```
ymherklotz, line 82: Yep I agree with the implementation of sink.

SSA/Projects/CIRCT/DC/DC.lean
```diff
@@ -0,0 +1,184 @@
...
+
+def join (x y : TokenStream) : TokenStream  :=
+  Stream.corec (β := TokenStream × TokenStream) (x, y) fun ⟨x, y⟩ =>
+    match x 0, y 0 with
+    | some _, some _ => (some (), (x.tail, y.tail))
+    | some _, none => (none, (x, y.tail))
+    | none, some _ => (none, (x.tail, y))
+    | none, none => (none, (x.tail, y.tail))
+
+def merge (x y : TokenStream) : TokenStream :=
```
ymherklotz: I think the type should be `TokenStream -> TokenStream -> ValueStream Bool`, where the value stream says true or false depending on if the left or right token was consumed.

SSA/Projects/CIRCT/DC/DC.lean
```diff
@@ -0,0 +1,184 @@
...
+            (some (), none, (x.tail, c.tail))
+          else
+            (none, some (), (x.tail, c.tail))
+
+def fork (x : TokenStream) : TokenStream × TokenStream  :=
+  Stream.corec₂ (β := TokenStream) x
+    fun x => Id.run <| do
+      (x 0, x 0, x.tail)
+
+def join (x y : TokenStream) : TokenStream  :=
```
ymherklotz, line 44: For now this looks good, but maybe we need to see if we can implement variadic arguments at some point, but for now we don't have to focus on that yet.

SSA/Projects/CIRCT/DC/DC.lean
```diff
@@ -0,0 +1,184 @@
...
+import SSA.Core.MLIRSyntax.EDSL
+import SSA.Projects.CIRCT.Stream.Stream
+
+namespace CIRCTStream
+namespace DC
+
+abbrev ValueStream := Stream
+abbrev TokenStream := Stream Unit
+
+def unpack (x : ValueStream α) : ValueStream α × TokenStream :=
```
ymherklotz, line 13: I think for now I'm happy with these semantics.  We are maybe missing out on a bit of the benefit of DC, because we are not really separating data and control, but I think we could refine it a bit later on.  A solution that I have started to think about is to really have an explicit separation between data and control.  Essentially, you can only have data operations between an `unpack` and a `pack`, and so you could just treat those regions differently.  Maybe we can explore that a bit, but at the same time we can use these semantics and try and prove an initial lowering from Handshake to DC.

SSA/Projects/CIRCT/Handshake/Handshake.lean
```diff
@@ -112,11 +115,71 @@ def altMerge (x y : Stream α) : Stream α :=
...
+def mux (x y : Stream α) (c : Stream Bool) : Stream α :=
+  Stream.corec (β := Stream α × Stream α × Stream Bool) (x, y, c) fun ⟨x, y, c⟩ => Id.run <| do
+    match x 0, y 0, c 0 with
+      | none, _, some true => (none, (x.tail, y, c)) -- could not pop anything
+      | some _, _, some true => (x 0, (x.tail, y, c.tail)) -- pop from x
+      | _, none, some false => (none, (x, y.tail, c)) -- could not pop anything
+      | _, some _, some false => (y 0, (x, y.tail, c.tail)) -- pop from y
+      | _, _, none => (none, (x, y, c.tail)) -- no pop
+
+-- discards any data: actually this should not return anything
```
ymherklotz, line 154: Yeah in practice this should not return anything, but to be able to consume a stream you have to produce another stream.  But as long as the stream never blocks and can always consume a stream, then I think it has the behaviour that we are looking for.

lakefile.lean
ymherklotz, line 1: I'm not sure why this file is part of this commit, should it be?

SSA/Projects/CIRCT/DC/DC.lean
luisacicolini: Yes I agree with this separation and I think the semantics we have so far is a good starting point for that, since `TokenStream` is explicitly different than `ValueStream` we can just use the two for different operations (control vs. data)

luisacicolini:
Added a small working example. I tried adding a more complex one using `DC.branch`, but somehow it's complaining about having two outputs, so for now we have this. I also changed the definition of `branch` slightly, this one seems to me more compliant with the docs.

ymherklotz (approved):
Looks good to me now, this implements all DC operations with pure streaming semantics and adds the WeakBisim relation to compare streams.

luisacicolini:
Happy to merge unless there are other issues CC: @tobiasgrosser @alexkeizer

PR #656, opened 2024-09-25
Predicate support

tobiasgrosser:


SSA/Experimental/Bits/Fast/FiniteStateMachine.lean
```diff
@@ -679,3 +679,33 @@ theorem decideIfZeros_correct {arity : Type _} [DecidableEq arity]
     intro x s h
     use x
     exact h
+
+inductive Predicate
+  ( α  : Type )
+  [ i : Fintype α ]
+  [ dec_eq : DecidableEq α ] :=
+| eq (t1 t2 : Term )
```
alexkeizer: We need to force the index to a particular value, to communicate the arity of the predicate

```suggestion
| eq (t1 t2 : Term ) : Predicate (Fin (max t1.arity t2.arity))
```
I think the `Fin (t1.xor t2).arity)` type you saw in the error unfolds to `max`,

SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -98,6 +98,15 @@ section Lemmas
...
   funext i; exact h i
 
+theorem corec_eq_corec {a : α} {b : β} {f g}
+    (R : α → β → Prop)
+    (h : ∀ a b, R a b →
+          let x := f a
+          let y := g b
+          R x.fst y.fst ∧ x.snd = y.snd) :
+    corec f a = corec g b := by
+  sorry
```
AtticusKuhn: @alexkeizer I don't understand the purpose of this theorem? And is it even true?

I feel like the correct version should be 

```lean

theorem corec_eq_corec {a : α} {b : β} {f g}
    (R : α → β → Prop)
    (thing : R a b)
    (h : ∀ a b, R a b →
          let x := f a
          let y := g b
          R x.fst y.fst ∧ x.snd = y.snd) :
    corec f a = corec g b := by
  ext i
  induction' i with i ih 
  simp [corec]
  have y := h a b thing
  simp at y
  cases' y with l r
  assumption
  sorry
```

AtticusKuhn: I filled out the proof as this:


```lean

theorem compose_first {α: Type u₁} (i : Nat) (a : α ) (f : α → α × Bool) : (f ((Prod.fst ∘ f)^[i] a)).1 = (Prod.fst ∘ f)^[i] (f a).1 :=
  match i with
    | 0 => by simp
    | i + 1=>
      let y:= @compose_first α i ((f a).1) f
      by simp [y]

theorem corec_eq_corec {a : α} {b : β} {f g}
    (R : α → β → Prop)
    (thing : R a b)
    (h : ∀ a b, R a b →
          let x := f a
          let y := g b
          R x.fst y.fst ∧ x.snd = y.snd) :
    corec f a = corec g b := by
  ext i
  have lem :  R ((Prod.fst ∘ f)^[i] (f a).1) ((Prod.fst ∘ g)^[i] (g b).1) ∧ corec f a i = corec g b i := by
    induction' i with i ih
    <;> simp [corec]
    apply h
    exact thing
    simp at h
    simp [corec] at ih
    have m := h ((Prod.fst ∘ f)^[i] (f a).1) ((Prod.fst ∘ g)^[i] (g b).1) (ih.1)
    cases' m with l r
    simp [r]
    have left : ((Prod.fst ∘ f)^[i] (f (f a).1).1)  = (f ((Prod.fst ∘ f)^[i] (f a).1)).1 := by
      rw [← @compose_first α]
    have right : ((Prod.fst ∘ g)^[i] (g (g b).1).1)  =(g ((Prod.fst ∘ g)^[i] (g b).1)).1  := by
      rw [← @compose_first β i]
    simp [left, right]
    exact l
  cases' lem with l r
  assumption
```

bollu:
I suggest we merge this in, and develop more theory upstream. Thanks a bunch @AtticusKuhn for filling in the proof! I cleaned up the PR, closed another sorry, and made the project build.

bollu:
@tobiasgrosser Do we want to keep the old implementation compiled? We are currently adding new fields, so the problem is that the old implementation does not know about `repeatBit`. Updating the old implementation to be in sync seems like a major pain point to me.

tobiasgrosser:
> I suggest we merge this in, and develop more theory upstream. Thanks a bunch @AtticusKuhn for filling in the proof! I cleaned up the PR, closed another sorry, and made the project build.

I think this is a great idea.

tobiasgrosser:
> @tobiasgrosser Do we want to keep the old implementation compiled? We are currently adding new fields, so the problem is that the old implementation does not know about `repeatBit`. Updating the old implementation to be in sync seems like a major pain point to me.

I need more context to understand the issue. What is the problem with the old implementation? In fact, do we not just build on-top of the old implementation?

bollu:
@tobiasgrosser we have two implementations, `SSA/Experiments/Bits/Decide.lean` and `SSA/Experimental/Bits/Fast/Decide.lean`. Both of these used to share the same definitions from `SSA/Experiments/Bits/Defs.lean`. But now, we added a field to `Defs` (`repeatBit`), which causes the old decision procedure in `Bits/Decide.lean` to fail.

Our automata / bitstream definitions all live in `Bits/Fast`, and we're building on top of that, which is also what the `bv_automata` tactic is based on.

bollu:
@tobiasgrosser my suggestion is to delete `Bits/{Defs.lean, Decide.lean}`, and continue work on `Bits/Fast/Decide.lean`.

tobiasgrosser:
> @tobiasgrosser we have two implementations, `SSA/Experiments/Bits/Decide.lean` and `SSA/Experimental/Bits/Fast/Decide.lean`. Both of these used to share the same definitions from `SSA/Experiments/Bits/Defs.lean`. But now, we added a field to `Defs` (`repeatBit`), which causes the old decision procedure in `Bits/Decide.lean` to fail.
> 
> Our automata / bitstream definitions all live in `Bits/Fast`, and we're building on top of that, which is also what the `bv_automata` tactic is based on.

Can you split them? E.g., just copy the definitions into Fast, and develop them independently for now?

This allows us to maintain the previous code while not being held back by the old one.

bollu:
@tobiasgrosser that's exactly what I did in the current PR.

tobiasgrosser:
Cool. Then lets merge.

tobiasgrosser:
If they are decoupled, there is no need to delete them, no? It might still be useful to have them in a repo, as updating them to the latest lean won't have a cost. However, I don't feel super strong here. Happy to get see them dropped, too.

bollu:
Let's merge without dropping then, in case it turns out to be useful to reference the old impl.

bollu:
@tobiasgrosser hit merge if you're happy.

SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -98,6 +98,45 @@ section Lemmas
...
+-/
+theorem compose_first {α: Type u₁} (i : Nat) (a : α) 
+    (f : α → α × Bool) : 
+    (f ((Prod.fst ∘ f)^[i] a)).1 = (Prod.fst ∘ f)^[i] (f a).1 :=
+  match i with
+    | 0 => by simp
+    | i + 1 => by simp [compose_first i ((f a).1) f]
+
+/--
+Coinduction principle for `corec`.
```
alexkeizer, line 113: I would call this bisimulation, rather than coinduction

SSA/Experimental/Bits/Fast/BitStream.lean
```diff
@@ -276,6 +315,16 @@ instance : Add BitStream := ⟨add⟩
 instance : Neg BitStream := ⟨neg⟩
 instance : Sub BitStream := ⟨sub⟩
 
+/-- `repeatBit xs` will repeat the first bit of `xs` which is `true`.
+That is, it will be all-zeros iff `xs` is all-zeroes,
+otherwise, there's some number `k` so that after dropping the `k` least
+significant bits, `repeatBit xs` is all-ones. -/
+def repeatBit (xs : BitStream) : BitStream :=
+  corec (b := (false, xs)) fun (carry, xs) =>
```
alexkeizer, line 323: I thought we wanted to change this name? I like `foldOr` more than `repeatBit`

SSA/Experimental/Bits/Fast/FiniteStateMachine.lean
```diff
@@ -679,3 +782,47 @@ theorem decideIfZeros_correct {arity : Type _} [DecidableEq arity]
...
+
+
+
+/--
+denote a reflected `predicate` into a `prop.
+-/
+def Predicate.denote : Predicate α → Prop
+| eq t1 t2 => t1.eval = t2.eval
+| and p q => p.denote ∧  q.denote
+| or p q => p.denote ∨  q.denote
```
alexkeizer, line 810: This is not quite right! We need to be taking in the arguments at the top-level, and pass those around.
Currently, we're denoting `eq t u` as equality of the *functions* that are denoted by `t` and `u`, meaning that we're universally quantifying each con-/disjunct separately.


```
def Predicate.denote (xs : α → BitStream) : Predicate α → Prop
| eq t1 t2 => t1.eval xs = t2.eval xs
| and p q => p.denote xs ∧ q.denote xs
| or p q => p.denote xs ∨ q.denote xs
```

SSA/Experimental/Bits/Fast/FiniteStateMachine.lean
```diff
@@ -679,3 +782,47 @@ theorem decideIfZeros_correct {arity : Type _} [DecidableEq arity]
...
+    let p := toFSM p
+    let q := toFSM q
+    composeBinary' FSM.and p q
+| .or p q =>
+    let p := toFSM p
+    let q := toFSM q
+    composeBinary' FSM.or p q
+
+theorem Predicate.toFsm_correct {k : Nat} (p : Predicate k) :
+  decideIfZeros p.toFSM = true ↔ p.denote := by sorry
```
alexkeizer, line 828: Similarly, this needs to be:
```
theorem Predicate.toFsm_correct {k : Nat} (p : Predicate k) :
  decideIfZeros p.toFSM = true ↔ (\all xs, p.denote xs) := by sorry
```

Issue #660, opened 2024-09-27
Update to Lean 2024-09-26 broke bv_automata


luisacicolini:
Updating to the new Mathlib version (2024-09-26) broke AliveStatement `bv_AddSub_1556`: 

```lean
theorem broken {x y : BitVec 1} : y - x = y ^^^ x := by
  bv_automata
```

The offending commit is likely https://github.com/leanprover/lean4-nightly/commit/a3ca15d2b2d5460f7d9d2b32ca90c4ccbda21168.

tobiasgrosser:
A branch with the broken test case is available in https://github.com/opencompl/lean-mlir/pull/662.

ineol:
I think the issue is that the statement is false (?). The error message is

```
application type mismatch
  Lean.ofReduceBool toto._nativeDecide_1 true (Eq.refl true)
argument has type
  true = true
but function has type
  Lean.reduceBool toto._nativeDecide_1 = true → broken._nativeDecide_1 = true
```
and `Lean.reduceBool toto._nativeDecide_1` reduces to `false`, which makes the function application ill-typed.

There is the same error with an obviously false statement `x = y`.

In any case it would be good to have a more informative error message.

tobiasgrosser:
The same issue arises with `BitVec 1` instead of `BitVec w`. For `BitVec 1`, bv_decide successfully proofs the relevant statement.

ineol:
As far as I understand, the `bv_automata` is a decision procedure for infinite bitstreams, so implicitly, it proves the result with arbitrary length.

Which version of Lean worked? I tried with 2024-09-20 and it has the same behavior.

luisacicolini:
works for me at e6e6f630 - maybe try with `bv_auto` instead of `bv_automata` ?

ineol:
I haven't found the source of the issue yet, but the problem is that now the `bv_automata` closes all the goals even if it fails, and produces an ill-typed proof. So in the `bv_auto` it stops the proof before `bv_decide` is able to prove the result. In the old version, it would leave the goal untouched, and `bv_auto` would try the next tactic.

ineol:
I think what was happening is that before the update, `ring_nf` was broken (it had a multiplication by 1), and produced a goal that was refused by `bv_automata` before calling `native_decide`, which let the `bv_decide` tactic close the goal.

tobiasgrosser:
[...]

PR #670, opened 2024-09-30
Make `bv_automata` fallible

ineol:
Currently, `bv_automata` does not fail if the goal is false, because `native_decide` doesn't fail, it produces an ill-typed proof.

On consequence is that it is awkward to use `bv_automata` in a tactic such as `bv_auto`, since it relied on tactics to fail to be able to try the next tactic (cf issue #660) It also give an error that is not helpful at all after the proof is complete.

This PR is a proof of concept that modifies the implementation of `native_decide` to check that the problem is solved by the decision procedure before constructing the proof object.

### DIFF
--- a/SSA/Experimental/Bits/Fast/Tactic.lean
+++ b/SSA/Experimental/Bits/Fast/Tactic.lean
@@ -8,7 +8,6 @@ open Lean Elab Tactic
 open Lean Meta
 open scoped Qq
 
-
 /-!
 # BitVec Automata Tactic
 There are two ways of expressing BitVec expressions. One is:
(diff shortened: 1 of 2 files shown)


tobiasgrosser:
This seems to be sensible. Should we merge it for now?

ineol:
I'm confused about the CI failure. The error is that it doesn't have permission to add the comment to PR with the Alive stats, but it works for other PRs...

tobiasgrosser:
I guess our CI only works for branches within the repo. Can you just create your branch within the repo and see if it goes through?

PR #677, opened 2024-10-02
feat: adding the flags for LLVM's binary operations

lfrenot:
Currently, `add` is the only LLVM operation with its flags implemented.
These changes implement the flags for LLVM's other Binary operations;
- `nsw` and `nuw` for `add`, `mul`, and `sub`.
- `exact` for `sdiv` and `udiv`


SSA/Projects/InstCombine/Base.lean
```diff
@@ -157,15 +161,25 @@ namespace MOp
 @[match_pattern] def urem   (w : Width φ) : MOp φ := .binary w .urem
 @[match_pattern] def srem   (w : Width φ) : MOp φ := .binary w .srem
 
-@[match_pattern] def mul    (w : Width φ) : MOp φ := .binary w .mul
-@[match_pattern] def sub    (w : Width φ) : MOp φ := .binary w .sub
-@[match_pattern] def sdiv   (w : Width φ) : MOp φ := .binary w .sdiv
-@[match_pattern] def udiv   (w : Width φ) : MOp φ := .binary w .udiv
 
-/- This definition is off by itself because it is different-/
+/- Theses definitions use NoWrapFlags -/
```
luisacicolini:
```suggestion
/- These definitions use NoWrapFlags -/
```

SSA/Projects/InstCombine/Base.lean
```diff
@@ -157,15 +161,25 @@ namespace MOp
...
+    (NoWrapFlags: NoWrapFlags := {nsw := false , nuw := false}) : MOp φ
+      := .binary w (.add  NoWrapFlags )
+@[match_pattern] def mul    (w : Width φ)
+    (NoWrapFlags: NoWrapFlags := {nsw := false , nuw := false}) : MOp φ
+      := .binary w (.mul  NoWrapFlags )
+@[match_pattern] def sub    (w : Width φ)
+    (NoWrapFlags: NoWrapFlags := {nsw := false , nuw := false}) : MOp φ
+      := .binary w (.sub  NoWrapFlags )
+
+/- Theses definitions use an exact flag -/
```
luisacicolini:
```suggestion
/- These definitions use an exact flag -/
```

luisacicolini (approved):
found a couple of typos but otherwise lgtm!

SSA/Projects/InstCombine/Base.lean
```diff
@@ -121,10 +121,14 @@ def reprWithoutFlags (op : MOp.BinaryOp) (prec : Nat) : Format :=
...
-    | .sdiv   => "sdiv"
-    | .udiv   => "udiv"
+    | .mul ⟨false, false⟩ => "mul"
+    | .mul ⟨nsw, nuw⟩    => toString f!"mul {nsw} {nuw}"
+    | .sub ⟨false, false⟩ => "sub"
+    | .sub ⟨nsw, nuw⟩    => toString f!"sub {nsw} {nuw}"
+    | .sdiv ⟨false⟩ => "sdiv"
+    | .sdiv ⟨true⟩ => "sdiv exact"
+    | .udiv ⟨false⟩  => "udiv"
+    | .udiv ⟨true⟩  => "udiv exact"
```
tobiasgrosser: Here, the previous alignment is broken. Either reduce the whitespace to one character or realign, I guess.

lfrenot: Thanks for the notice, I made a new commit to fix the alignment

tobiasgrosser:
Cool. This looks great to me now.

SSA/Projects/InstCombine/Base.lean
```diff
@@ -119,12 +119,16 @@ def reprWithoutFlags (op : MOp.BinaryOp) (prec : Nat) : Format :=
...
-    | .mul    => "mul"
-    | .sub    => "sub"
-    | .sdiv   => "sdiv"
-    | .udiv   => "udiv"
+    | .add  ⟨false, false⟩ => "add"
+    | .add  ⟨nsw, nuw⟩     => toString f!"add {nsw} {nuw}"
+    | .mul  ⟨false, false⟩ => "mul"
+    | .mul  ⟨nsw, nuw⟩     => toString f!"mul {nsw} {nuw}"
+    | .sub  ⟨false, false⟩ => "sub"
+    | .sub  ⟨nsw, nuw⟩     => toString f!"sub {nsw} {nuw}"
```
alexkeizer, line 127: Why do we have a `toString` here? If we want to have a string, we can write it as `s!"..."`, but here the return type is `Format`, so just returning the result of `f!"..."` should be fine, I think

PR #678, opened 2024-10-02
feat: adding the flags for LLVM's bitwise binary operations

lfrenot:
These changes implement the flags for LLVM's bitwise binary operations;
- `nsw` and `nuw` for `shl`
- `exact` for `lshr` and `ashr`
- `disjoint` for `or`.


tobiasgrosser:
Nice. I guess you need to fix the build and have @luisacicolini do a pass.

SSA/Projects/InstCombine/Base.lean
```diff
@@ -270,7 +279,12 @@ namespace Op
 @[match_pattern] abbrev icmp (c : IntPredicate)   : Nat → Op  := MOp.icmp c ∘ .concrete
 @[match_pattern] abbrev const (w : Nat) (val : ℤ) : Op        := MOp.const (.concrete w) val
 
+/- This operation is separate from the others because in takes in a flag: disjoint. -/
```
luisacicolini:
```suggestion
/- This operation is separate from the others because it takes in a flag: disjoint. -/
```

SSA/Projects/InstCombine/PaperExamples.lean
```diff
@@ -34,12 +34,10 @@ theorem shift_mul:
...
-    simp only [ge_iff_le,
-      EffectKind.return_impure_toMonad_eq, Option.pure_def, mul_eq,
-      Option.bind_eq_bind, Option.none_bind, h, ↓reduceIte, Option.none_bind,
-      Option.bind_none, Option.some_bind, Refinement.some_some, Refinement.refl]
-  simp only [Bool.false_eq_true, shiftLeft_eq', false_and, toNat_shiftLeft, toNat_ofNat, _root_.or_self,
-  ↓reduceIte, one_shiftLeft_mul, Refinement.refl]
+    simp only [Bool.false_eq_true, ge_iff_le, toNat_allOnes, false_and, _root_.or_self, ↓reduceIte,
+      shiftLeft_eq', EffectKind.return_impure_toMonad_eq, Option.pure_def, mul_eq,
+      Option.bind_eq_bind, Option.none_bind, Option.bind_none, Option.some_bind, Refinement.refl, h]
+  simp only [one_shiftLeft_mul, Refinement.refl]
```
luisacicolini: AFAIK when `simp` is final you can just leave `simp` instead of `simp only` (which shall instead be used when non-final)

luisacicolini:
apart from some small nits lgtm :)

SSA/Projects/InstCombine/PaperExamples.lean
lfrenot: OK, I will change that right now

PR #683, opened 2024-10-07
chore: update mathlib

luisacicolini:


luisacicolini:
thanks a lot for the help @tobiasgrosser !! :)

tobiasgrosser:
> thanks a lot for the help @tobiasgrosser !! :)

Thank you for the update. A small trick I used here. I have the old and the new code in separate vs code windows and then compare the proof-state line-by-line. This did not make it trivial, but it certainly helped.

luisacicolini:
got it! I shall try it like that next time

PR #686, opened 2024-10-09
feat: add `BitVec.getMsbD_(add, sub, neg)` and `Bitvec.getLsbD(sub, neg)`

luisacicolini:
Still WIP, currently thinking how to refactor `sub` theorems to have a simpler `rhs`, without the explicit negation.


luisacicolini:
there seems to be no such thing as a `borrow` for `BitVec` (provided that I am looking for the right word and it does not have another name). In this case I guess going through addition is a reasonable option.

luisacicolini:
Merged by mistake so I reverted it (https://github.com/opencompl/lean-mlir/pull/691)

PR #689, opened 2024-10-10
feat: added overflow tags to EDSL and prettyEDSL, and re-generated the test cases

lfrenot:


SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -101,11 +101,23 @@ def mkExpr (Γ : Ctxt (MetaLLVM φ).Ty) (opStx : MLIR.AST.Op φ) :
...
+          let attr? := opStx.attrs.getAttr "overflowFlags"
+          match attr? with
+            | .none =>  pure <| Sum.inl (MOp.BinaryOp.shl)
+            | .some y => match y with
+              | .opaque_ "llvm.overflow" "nsw" => pure <| Sum.inl (MOp.BinaryOp.shl ⟨true, false⟩)
+              | .opaque_ "llvm.overflow" "nuw" => pure <| Sum.inl (MOp.BinaryOp.shl ⟨false, true⟩)
+              | .list [.opaque_ "llvm.overflow" "nuw", .opaque_ "llvm.overflow" "nsw"] => pure <| Sum.inl (MOp.BinaryOp.shl ⟨true, true⟩)
+              | .list [.opaque_ "llvm.overflow" "nsw", .opaque_ "llvm.overflow" "nuw"] => pure <| Sum.inl (MOp.BinaryOp.shl ⟨true, true⟩)
+              | .opaque_ "llvm.overflow" s => throw <| .generic s!"The overflow flag {s} not allowed. We currently support nsw (no signed wrap) and nuw (no unsigned wrap)"
+              | _ => throw <| .generic s!"Unrecognised overflow flag found: {MLIR.AST.docAttrVal y}. We currently support nsw (no signed wrap) and nuw (no unsigned wrap)"
```
alexkeizer, line 120: Can we factor this out into a generic parseOverflowFlags function that we can reuse, instead of copying the same code for every instruction that uses these flags?

SSA/Projects/InstCombine/LLVM/PrettyEDSLOverflow.lean
```diff
@@ -0,0 +1,33 @@
+import SSA.Core.MLIRSyntax.PrettyEDSL
+import SSA.Projects.InstCombine.Tactic
+open Lean
+
+namespace MLIR.EDSL
+open Pretty
+
+
+declare_syntax_cat MLIR.Pretty.overflow_flag
```
alexkeizer: I think it could help to have a docstring near the top of the file, that explains what this file does.
Something like:
```
/-!
# Pretty syntax for overflow flags

This file defines the `MLIR.Pretty.overflow_op` syntax category which, just like `MLIR.Pretty.uniform_op`,
is a category of MLIR operations where the arguments and return values are all of the same type.
Additionally, `overflow_op`s may be annotated with overflow flags, like `nsw` or `nuw`, which get
translated to [insert description of the generic attributes here] in the generic syntax.
-/
```

You might want to wordsmith that suggestion a little bit

SSA/Projects/InstCombine/LLVM/PrettyEDSLOverflow.lean
```diff
@@ -0,0 +1,33 @@
...
+declare_syntax_cat MLIR.Pretty.overflow_op
```
alexkeizer, line 17: The overflow flags are LLVM specific, right? In that case, we should be putting these things in the `InstCombine` namespace, instead of the generic `MLIR.Pretty` namespace

```suggestion
declare_syntax_cat InstCombine.overflow_op
```

SSA/Projects/InstCombine/LLVM/PrettyEDSLOverflow.lean
```diff
@@ -0,0 +1,33 @@
...
+
+syntax "nsw" : MLIR.Pretty.overflow_flag
+syntax "nuw" : MLIR.Pretty.overflow_flag
+
```
alexkeizer: IIUC, the code matches on specific overflow flags, making the `overflow_flag` syntax category not really extensible.

In such cases, we can choose to define our syntax in a closed way as well, to prevent users from extending the syntax and getting obscure errors.
```suggestion
syntax overflow_flag := "nsw" <|> "nuw"
```
Notice that I gave the syntax a name (`overflow_flag`).

Then you can remove the `declare_syntax_cat MLIR.Pretty.overflow_flag` line near the top of the file.
Unlike syntax categories, named syntax declarations do pick up on the current namespace, so we don't have to explicitly namespace them ourselves (hence I dropped the `MLIR.Pretty`).

alexkeizer (commented):
Thanks @lfrenot. It looks good as-is, but I left some comment for how we can clean it up a bit more, and make it even nicer!

SSA/Projects/InstCombine/LLVM/EDSL.lean
lfrenot: I'm not sure
When I tried doing that, I had issues with the throws
But I was probably missing something

SSA/Projects/InstCombine/tests/LLVM/g2010h11h23hDistributed.lean
```diff
@@ -39,7 +39,7 @@ theorem foo_proof : foo_before ⊑ foo_after := by
   intros
   try simp
   ---BEGIN foo
-  apply foo_thm
+  all_goals (try extract_goal ; sorry)
```
tobiasgrosser, line 42: Why do we drop `apply foo_thm` here?

lfrenot: Not sure, I can try to run the script by hand to see where it fails
It may be due to the changes from the flags

tobiasgrosser: It would be interesting to know at least for one case why this now fails.

tobiasgrosser: On the other hand, I think we can merge this (after you addressed the comments from @alexkeizer). And then iterate on improving this.

lfrenot: I ran it alone on this file and got foo_thm
so I don't know

tobiasgrosser: Interesting.  Let's merge this now, but maybe take a note and debug this afterwards.

PR #709, opened 2024-10-16
feat: Fixed lshr and shl being swapped in PrettyEDSL

lfrenot:


SSA/Projects/InstCombine/scripts/test/minimal.lean
```diff
@@ -0,0 +1,102 @@
+import SSA.Projects.InstCombine.LLVM.Semantics
```
tobiasgrosser: This file seems unrelated. Also, do you need to update the test cases for this specific change?

lfrenot: Oh, yes I didn't mean to have this file still there

lfrenot: As for updating the test cases, I wanted to do some edits to the generators first

PR #718, opened 2024-10-23
feat: add scripts for evaluation of bv performance

luisacicolini:
`compare-leansat-vs-bitwuzla` sets the width of bitvector in tested files to different values and run the tests `reps` times.


SSA/Projects/InstCombine/HackersDelight/ch2_2AdditionAndLogicalOps.lean
```diff
@@ -10,183 +10,183 @@ namespace AdditionCombinedWithLogicalOperations
...
+variable {x y z : BitVec 64}
 
 theorem neg_eq_not_add_one :
     -x = ~~~ x + 1 := by
   /-
   Bitwuzla proved the goal after 0.000000ms
   LeanSAT proved the goal after 2.000000ms: rewriting 0.000000ms, bitblasting 0.000000ms, SAT solving 0.000000ms, LRAT processing 0.000000ms
   -/
-  bv_compare'
+  sorry
```
tobiasgrosser: Why are these changes needed? Can we drop them?

SSA/Projects/InstCombine/TacticAuto.lean
```diff
@@ -185,7 +185,7 @@ macro "alive_auto": tactic =>
 macro "bv_compare'": tactic =>
   `(tactic|
       (
-        -- bv_compare "/usr/local/bin/bitwuzla"
-        bv_decide
+        -- bv_compare "/usr/local/bin/bitwuzla""""""
+        bv_decide -- test
```
tobiasgrosser: These changes are unrelated.

SSA/Projects/InstCombine/tests/proofs/g2004h02h23hShiftShiftOverflow_proof.lean
```diff
@@ -7,4 +7,3 @@ section g2004h02h23hShiftShiftOverflow_proof
 theorem test_thm (x : BitVec 32) : (x.sshiftRight 17).sshiftRight 17 = x.sshiftRight 31 := sorry
 
 theorem test2_thm (x : BitVec 32) : x <<< 34 = 0#32 := sorry
-
```
tobiasgrosser: This change is unrelated.

xdsl.nix
```diff
@@ -1,46 +0,0 @@
...
-  # do not run tests
-  doCheck = false;
-
-  meta = with lib; {
-    description = "Audio Activity Detection tool that can process online data as well as audio files";
-    mainProgram = "xdsl";
-    homepage = "https://github.com/amsehili/xdsl/";
-    changelog = "https://github.com/amsehili/xdsl/blob/v${version}/CHANGELOG";
-    license = licenses.mit;
-    maintainers = [ ];
```
tobiasgrosser: Why are you dropping nix?

tobiasgrosser (approved):
Otherwise, this LGTM.

SSA/Projects/InstCombine/TacticAuto.lean
luisacicolini: I added the `-- test` becausa to make the tests fully automatic I am `sed`ding `bv_decide` to un/comment it and un/comment the line above. However, the file contains another `bv_decide` that we don't want to replace, so I added the `-- test` to distinguish the two. The `"""""` are a mistake instead

SSA/Projects/InstCombine/HackersDelight/ch2_2AdditionAndLogicalOps.lean
luisacicolini: Yes, atm the scripts replace `sorry` and `bv_compare'` when the tests are being ran, and `sorry` is put back after the tests have been ran to "restore" them to an original state. But if we're happy with always having `bv_compare'` and just change the `Leanwuzla` import that also works

SSA/Projects/InstCombine/TacticAuto.lean
tobiasgrosser: Then maybe explain using something like `-- replace this with bv_compare to evaluate performance`. This is a helpful comment and you can still match on it.

SSA/Projects/InstCombine/HackersDelight/ch2_2AdditionAndLogicalOps.lean
tobiasgrosser: I think we are.

luisacicolini: Actually, AFAIU leaving `bv_compare'` is preventing CI from passing

luisacicolini:
Good to go if CI passes now, I'll then open a new PR for the proper github action

PR #726, opened 2024-10-26
feat: add `(msb, getMsbD)_twoPow`

luisacicolini:


SSA/Projects/InstCombine/ForLean.lean
```diff
@@ -494,6 +494,23 @@ theorem msb_abs {w : Nat} {x : BitVec w} :
...
   simp [BitVec.abs]
 
+@[simp]
+theorem getMsbD_twoPow {i j w: Nat} :
+    (twoPow w i).getMsbD j = (decide (i < w) && decide (j = w - i - 1)) := by
+  simp only [getMsbD_eq_getLsbD, getLsbD_twoPow]
+  by_cases h₀ : i < w <;> by_cases h₁ : j < w <;>
+  simp [h₀, h₁]
+  omega
+  omega
```
tobiasgrosser:
```suggestion
  simp [h₀, h₁] <;> omega
```

tobiasgrosser: If there are multiple cases, we want to either use `·` to enumerate them of handle all of them with `<;>`. Given that you already started with `<;>` we probably should use it for `omega`, too.

tobiasgrosser (approved):
Very nice.

PR #728, opened 2024-10-26
feat: plots and stats for llvm tests

luisacicolini:
wanted to check the diff nicely after fixing the conflicts


luisacicolini:
before merging the PR I would like to merge the scripts for hacker's delight and llvm tests so to only have one and make things a bit tidier :)

luisacicolini:
I feel like this is now ready to be merged. I shall run some more tests today and potentially update the plots etc, but will open another PR for that. CC @tobiasgrosser @bollu

tobiasgrosser:
Cool. Then let's merge this.

PR #731, opened 2024-10-27
chore: fix encoding issues

tobiasgrosser:
When rebuilding the test cases I run into encoding issues that lead to the error:

  codec can't decode byte 0xe2 in position

This change fixes this issue.


lfrenot:
Looks good to me

lfrenot (approved):

PR #738, opened 2024-10-29
feat: Adding boolean notation to llvm constant declaration

lfrenot:
The goal is to make the minimal definition in [SSA/Projects/InstCombine/scripts/files/minimal.lean](https://github.com/opencompl/lean-mlir/compare/instcombine-bool?expand=1#diff-ae27b88181224ab18e88f86dab646614a0ac054f918a9ad392df4c0251af6ae9) able to be interpreted


tobiasgrosser:
This has been -- for now -- superseded with https://github.com/opencompl/lean-mlir/pull/739. We can revise this if we need support at the generic-syntax level.

PR #783, opened 2024-11-04
chore: update data collection script for hacker delight results

luisacicolini:


tobiasgrosser:
Do we want to push all the `results`?

luisacicolini:
can I remove the plotting scripts from the repo? I would like to only keep them in the paper repo so we have one version only and avoid confusion.

PR #808, opened 2024-11-08
feat: simproc to convert bitvector literals into twoPow.

bollu:
This lets us simplify expressions like `x / (8#64)` into `x / twoPow 3`, which then becomes `x >>> 3`.
```lean
/--
We need the `hk: k < w`, since otherwise the `twoPow` function will wrap around and become a smaller division
while the shift will stay a large shift.
I tried rephrasing this as `(x >>> BitVec.ofNat w k)`, but this doesn't have the right effect either:
See that `twoPow w k` wraps around when `k ≥ w`, while `BitVec.ofNat w k` wraps around when `k ≥ 2^w`.
I think we can make these match by setting `(BitVec.ofNat w (k % w))`, but that seems even crazier
than adding a side condition. Will discuss and see what we want.
-/
theorem div_twoPow_eq (x : BitVec w) (k : Nat) (hk : k < w) : x / (twoPow w k) = x >>> k := by 
  apply BitVec.eq_of_toNat_eq
  have : 2^k < 2^w := Nat.pow_lt_pow_of_lt (by decide) hk
  simp [Nat.shiftRight_eq_div_pow, Nat.mod_eq_of_lt this]
```


bollu:
This was upstreamed.

PR #849, opened 2024-11-14
chore: add TACNAME to get the name of the theorem

bollu:
@luisacicolini we add the ability to get the name of the theorem that is being proven, and store this in the `filename` column.


tobiasgrosser:
This is great. Let's merge it.

PR #866, opened 2024-11-21
chore: update mathlib nightly 2024-11-21

luisacicolini:
CC: @ineol some theorems (e.g. `Array.mem_push`) exist already, with the same name and a slightly different declaration in `Array/Lemmas`, and this is currently breaking the build. For now I commented the theorems in `ForLean`, but that obviously did not solve the problem. I'm happy to rename the theorems you wrote with a `'` if that's okay for you, but you might want to take a look at the existing theorems and maybe see if they can be useful?


ineol:
Thank you for the heads-up, I'll take a look tomorrow morning and use the standard lemmas

ineol:
Done!

luisacicolini:
thank you so much @ineol :)

PR #871, opened 2024-11-30
feat: circuit based automata with support for predicates

bollu:
We add a `Predicate` along with a `Term` which models bitvector predicates.

- [x] implement BV predicates
- [x] reduce to `decideAllZeroes`
- [x] implement reflection for predicates (`eq`, `neq`).
- [x] implement new meta-code to convert to negation normal form.
- [x] implement coverage for the other predicates (`<u`, `<=u`, `<s`, `<=s`)


tobiasgrosser:
Wow, any examples of inequalities already in the test cases?

bollu:
No, because that needs meta-code to hook it up, but I'll do it over the next day or two :)

tobiasgrosser:
OK.

SSA/Experimental/Bits/Fast/Reflect.lean
```diff
@@ -130,57 +139,197 @@ TODO(@bollu): We also assume that the goals are in negation normal form, and if
...
+---
+warning: declaration uses 'sorry'
+-/
+#guard_msgs in example : ∀ (w : Nat) (a b : BitVec w), a = b := by
+  intros w a b
+  bv_reflect
+  sorry
+
+/-- error: expcted width to be a free variable, found '10' (bitwidth with width '10' is 'b') -/
+#guard_msgs in example : ∀ (a b : BitVec 10), a = b := by
```
tobiasgrosser:
```suggestion
/-- error: expected width to be a free variable, found '10' (bitwidth with width '10' is 'b') -/
#guard_msgs in example : ∀ (a b : BitVec 10), a = b := by
```

bollu: @tobiasgrosser Thanks for the patch. Unfortunately, I cannot accept the suggestion, since it will break the build: For these changes, it would help a bit more if the patch could be made against the part of the codebase that emits this error. Thanks for pointing the typo out, though, I appreciate it!

tobiasgrosser: I don't get this. I just fixed the typo? Why does this break things?

SSA/Experimental/Bits/Fast/Tactic.lean
```diff
@@ -1,491 +0,0 @@
...
-  apply congrFun
-  native_decide
-  ))
-
-/-!
-# Test Cases
--/
-
-def alive_1 {w : ℕ} (x x_1 x_2 : BitVec w) : (x_2 &&& x_1 ^^^ x_1) + 1#w + x = x - (x_2 ||| ~~~x_1) := by
-  bv_automata
```
tobiasgrosser, line 342: These test cases might be worth reviving whenever your tactic is powerful enough.

SSA/Experimental/Bits/Fast/Reflect.lean
```diff
@@ -0,0 +1,671 @@
...
+
+/-- Can solve implicitly quantified expressions by directly invoking bv_automata3. -/
+theorem eq2 (w : Nat) (a : BitVec w) : a = a := by
+  bv_automata3
+#print eq1
+
+
+open NNF in
+
+example (w : Nat) (a b : BitVec w) : a + b = b + a := by
```
bollu, line 597: @tobiasgrosser I've left examples here, and I'm still adding to them at the moment. Feel invited to add new ones, which will help me flag down what isn't working right now!

bollu:
Core support is now complete. We can add the following incrementally:
- [ ] Constants other than `0`, `1`, `-1` by supporting arbitrary `BitVec.ofNat w n` declarations
- [ ] `shl bv k` where `k` is a compile-time known constant.
- [ ] `lshr bv k` where `k` is a compile-time known constant.
- I *don't think* that `ashr bv k` is possible, since the arithmetic shifting is an "irregular" operation across different bitwidths. I wonder if reading the p-adic decision procedure will give me some insights here.

bollu:
The next step seems to be to actually verify the new algorithm.

SSA/Experimental/Bits/Fast/Tactic.lean
bollu: Done, I've revived them now.

tobiasgrosser (approved):
Let's merge, and then iterate.

PR #885, opened 2024-12-12
feat: Add user-facing tactic bail out based on circuit complexity

bollu:
To implement this, we measure the total size of the circuit that is built for the FSM. This automatically includes the state space of the FSM, since there is one circuit to compute each bit of the FSM state, and is thus the natural measure of complexity.

The user-facing API is:

```lean
theorem add_eq_xor_add_mul_and' (x y : BitVec w) :
    x + y = (x ^^^ y) + (x &&& y) + (x &&& y) := by
  bv_automata_circuit
  bv_automata_circuit (config := { circuitSizeThreshold := 100 } )
```

Furthermore, in implementing this, we cleanup our implementation of reflection. We were previously working directly with raw `Expr`s at the tactic level. Now, we switch to parsing the current goal state (which is a raw `Expr`) into a well-typed and well-denoted term of type `Predicate`, so that we can `Predicate.toFSM.circuitComplexity` easily. Finally, we convert the predicate back into an expression with `Predicate.quote`. 

This process not also helps clean up the code, but made me realize a programming error in the implementation of `BitVec.ofNat` in the corner case where one calls `BitVec.ofNat w <symbolic-natural>`. The previous version, when building raw expressions, would try to reflect this into a `Term`, but such a `Term` would not be a closed term (i.e. it would depend on a free variable, the `<symbolic-natural>`, and would in fact cause a Lean error). This has now been fixed, where we now correctly abstract the entire `BitVec.ofNat w <symbolic-natural>` into an uninterpreted expression.

### DIFF
--- a/SSA/Experimental/Bits/Fast/Reflect.lean
+++ b/SSA/Experimental/Bits/Fast/Reflect.lean
@@ -47,6 +47,49 @@ def BitStream.denote (s : BitStream) (w : Nat) : BitVec w := s.toBitVec w
   simp [denote, toBitVec]
   sorry
 
+open Lean in
+def mkBoolLit (b : Bool) : Expr :=
+  match b with
+  | true => mkConst ``true
+  | false => mkConst ``false
+
+open Lean in
+def Term.quote (t : _root_.Term) : Expr :=
+  match t with
+  | ofNat n => mkApp (mkConst ``Term.ofNat) (mkNatLit n)
+  | var n => mkApp (mkConst ``Term.var) (mkNatLit n)
+  | zero => mkConst ``Term.zero
+  | one => mkConst ``Term.one
+  | negOne => mkConst ``Term.negOne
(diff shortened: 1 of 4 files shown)


SSA/Experimental/Bits/Fast/Reflect.lean
```diff
@@ -316,62 +392,63 @@ and furthermore, it will reflect all terms as variables.
...
-      return {bvToIxMap := map, e := e}
-    | _ =>
-      let e := mkApp (mkConst ``Term.ofNat) nExpr
-      return { bvToIxMap := map, e := e }
+      return {bvToIxMap := map, e := Term.one }
+    | .some n =>
+      return { bvToIxMap := map, e := Term.ofNat n }
+    | none =>
+      logWarning "expected concrete BitVec.ofNat, found symbol '{n}', creating free variable"
+      reflectAtomUnchecked map w e
```
bollu, line 418: Bug that was discovered by cleanup.

PR #893, opened 2024-12-13
feat: Adapt the generic circuit based decision procedure for fixed-width

bollu:
Write a special variant of the generic width decision procedure that only works for a single fixed width. The key idea is to use the automata we have to 'accelerate', and simulate upto the fixed bitwidth (e.g. `64`) and retrieve the result of the automata at that specific width.

The cost is that each step, we consider all combinations of input bitwidths, which can make this somewhat to dramatically expensive. However, note that the entire decision procedure was **not meant** to deal with fixed width anyway, so it's pretty remarkable that we can do this at all...


bollu:
While this didn't actually help, writing the code helped clarify my thinking. @tobiasgrosser , I'd appreciate a merge here, so I can move on to proving the correctness of the core algorithm from tomorrow ^_^

tobiasgrosser:
Sure, feel free to merge.

PR #898, opened 2024-12-17
feat: Dockerfile for bv-evaluation scripts

alexkeizer:
This adds a dockerfile that has the right dependencies to run the experiment scripts.

Mostly taken from the paper repo, except here I use the actual files from the checkout, rather than downloading a new clone of the repo, hence I moved it to the lean-mlir repository.

I also added a `run-script.sh` wrapper, that you can use as `./run-script.sh collect-data-llvm-symbolic.py` to run the specified script inside a docker container (assuming the image was previously built, e.g., with `make docker-image`).

There is still some work left; primarily making sure whatever directories we use to output data to in the scripts are actually mounted to the host fs.


bv-evaluation/collect-data-alive.py
```diff
@@ -4,7 +4,7 @@
 import os 
 import re
 
-paper_directory = 'for-paper/'
+paper_directory = ''
```
alexkeizer, line 7: @luisacicolini can you remind me why we removed this `for_paper/`? I think this change broke the CI

luisacicolini: i dont remember exactly, happy to take a look

PR #925, opened 2024-12-29
feat: Proofs for Bits/Fast/FiniteStateMachine, Bits/Fast/Reflect

bollu:
- [x] Prove that the FSMs correctly simulate the bitstream operations
- [x] Prove that the bitstream operations match the denotations
- [x] Put both proofs together to show an end-to-end proof.

### DIFF
--- a/SSA/Experimental/Bits/Fast/Decide.lean
+++ b/SSA/Experimental/Bits/Fast/Decide.lean
@@ -13,34 +13,41 @@ instance (p : Predicate) :
 We can decide that the evaluation of a predicate is forever false,
 which is to say, that it is true for all bitwidths.
 -/
-instance (p : Predicate) :
+instance DecidableForallPredicateEvalEqFalse (p : Predicate) :
     Decidable (∀ (n : ℕ) (x : List BitStream) , p.eval x n = false) :=
   decidable_of_iff
     (decideIfZeros (predicateEvalEqFSM p).toFSM) $ by
-  rw [decideIfZeros_correct]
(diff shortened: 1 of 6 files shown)


SSA/Experimental/Bits/Fast/Decide.lean
```diff
@@ -23,154 +13,41 @@ instance (p : Predicate) :
...
--- #guard_msgs in #eval! decide (not (xor x y)) (and x y - (or x y)).decr
-
-/-- info: true -/
-#guard_msgs in #eval! decide (not (xor x y)) (and x y + not (or x y))
-
-/-- info: true -/
-#guard_msgs in #eval! decide (or x y) (and x (not y) + y)
-
-/-- info: true -/
-#guard_msgs in #eval! decide (and x y) (or (not x) y - not x)
```
tobiasgrosser, line 174: Why do you drop the test cases?

bollu: The tests have been ported to `Bits/Fast/Reflect.lean`, they are not dropped.

This tests the old decision procedure, which tries to verify `t1.eval = t2.eval`. I drop support for the old version, so I can focus on getting the new version sorry free. Maintaining both adds a bit too much friction to the proof process. I'm pretty close to an end-to-end, sorry free proof.

Note that our current framework subsumes this, as we can simulate `t1.eval = t2.eval` with `Predicate.eval (Predicate.eq t1 t2)`.

PR #949, opened 2025-01-07
feat: add `comb` dialect

luisacicolini:
This PR adds the [`comb` dialect](https://circt.llvm.org/docs/Dialects/Comb/#combadd-circtcombaddop). 
`truthTable` operation and wildcard/cases SV semantics are missing, as they will require more infrastructure (and are not that used AFAIU)


SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,560 @@
...
+| divu (w : Nat)
+| extract (w : Nat) (n : Nat) -- I tried to avoid the two args but could not think of a better solution
+| icmp (w : Nat) (n : Nat)
+| mods (w : Nat)
+| modu (w : Nat)
+| mul (w : Nat)
+| mux (w : Nat)
+| or (w : Nat)
+| parity (w : Nat)
+| replicate (w : Nat) (n : Nat)
```
alexkeizer: I think having the second argument here is fine; I presume it's needed because the output type of `replicate w n` is a bitvector of length `w * n`, and we don't allow dependent types in our model.

That does mean that we can only replicate by a static/constant amount. The [Circt docs](https://circt.llvm.org/docs/Dialects/Comb/#combreplicate-circtcombreplicateop) indeed describe `comb.replicate` as "Concatenate the operand a constant number of times", and specify only a single operand.

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,560 @@
...
+| shl (w : Nat)
+| shrs (w : Nat)
+| shru (w : Nat)
+| sub (w : Nat)
+| xor (w : Nat)
+deriving Inhabited, DecidableEq, Repr
+
+inductive Ty
+| bv (w : Nat) : Ty -- A bitvector of width `Ty2`.
+| nat (n : Nat): Ty -- A very unfair trick
```
alexkeizer: This `nat n` feels a bit dubious

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,560 @@
...
+| bool : Ty
+deriving Inhabited, DecidableEq, Repr
+
+open TyDenote (toType) in
+instance instCombTyDenote : TyDenote Ty where
+toType := fun
+| Ty.bv w => BitVec w
+| Ty.nat _ => Nat
```
alexkeizer: Especially considering we completely ignore the argument to the `nat` type in our denotation.

I'd expect either a single `nat` type, or that `nat n` gets denoted as `{ m : Nat // m = n }` at which point our type has only a single inhabitant and is rather redundant as argument type.

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,560 @@
...
+  | .divu w => [Ty.bv w, Ty.bv w]
+  | .extract w n => [Ty.bv w, Ty.nat n]
+  | .icmp w n => [Ty.bv w, Ty.bv w, Ty.nat n]
+  | .mods w => [Ty.bv w, Ty.bv w]
+  | .modu w => [Ty.bv w, Ty.bv w]
+  | .mul w => [Ty.bv w, Ty.bv w]
+  | .mux w => [Ty.bv w, Ty.bv w, Ty.bool]
+  | .or w => [Ty.bv w, Ty.bv w]
+  | .parity w => [Ty.bv w]
+  | .replicate w n => [Ty.bv w, Ty.nat n]
```
alexkeizer:
```suggestion
  | .replicate w n => [Ty.bv w]
```

Concretely, `n` should be an annotation of the operation, *not* an operand.

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,560 @@
...
+    | .mods _, arg, _ => Comb.mods (arg.getN 0 (by simp [DialectSignature.sig, signature])) (arg.getN 1 (by simp [DialectSignature.sig, signature]))
+    | .modu _, arg, _ => Comb.modu (arg.getN 0 (by simp [DialectSignature.sig, signature])) (arg.getN 1 (by simp [DialectSignature.sig, signature]))
+    | .mul _, arg, _ => Comb.mul (arg.getN 0 (by simp [DialectSignature.sig, signature])) (arg.getN 1 (by simp [DialectSignature.sig, signature]))
+    | .mux _, arg, _ => Comb.mux (arg.getN 0 (by simp [DialectSignature.sig, signature])) (arg.getN 1 (by simp [DialectSignature.sig, signature])) (arg.getN 2 (by simp [DialectSignature.sig, signature]))
+    | .or _, arg, _ => Comb.or (arg.getN 0 (by simp [DialectSignature.sig, signature])) (arg.getN 1 (by simp [DialectSignature.sig, signature]))
+    | .parity _, arg, _ => Comb.parity (arg.getN 0 (by simp [DialectSignature.sig, signature]))
+    | .replicate _ n, arg, _ => Comb.replicate (arg.getN 0 (by simp [DialectSignature.sig, signature])) n
```
alexkeizer: This is effectively how you model `replicate` already; you've ignored the dynamic argument in favour of the `n` annotation

alexkeizer (commented):
I have some comments about the way you model `replicate`

alexkeizer:
To make, say, `add`, variadic you could add another parameter to the operation, as in:
```lean
| add (w : Nat) (arity : Nat)
```

Which you then pick up in the signature to specify we want `arity` operands
```lean
| add w arity => List.replicate arity (Ty.bv w)
```
(You might want to specify `arity+1` instead, so you don't have to deal with the corner case of having 0 arguments).

For the operation semantics, you could fold over the list of argument using the current binary implementations

alexkeizer:
>  more in general management of bitvectors' width: in the semantics I currently treat it as an additional argument of the operations (when the bitvectors involved need to have the same width), but I'm not sure this is the right approach

This is exactly what we do in LLVM dialect, and so far it's worked out great there, so it feels like the way to go here also!

alexkeizer:
I'm not sure quite what you mean by `sv` behaviour or truth table operations, but feel free to ask if you have more specific questions about them!

luisacicolini:
Thanks for the comments @alexkeizer! I'll start to fix the variadic implementation and the replicate.

`sv` is systemverilog-specific behavior, whose semantics requires a more complex model (I am not even sure we want to model that quite yet: you can find it in the wildcard and case in/equality sections [here](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=8299595)).

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -120,25 +120,25 @@ section Dialect
 -/
 
 inductive Op
-| add (w : Nat)
-| and (w : Nat)
-| concat (w₁ : Nat) (w₂ : Nat)
+| add (w : Nat) (n : Nat) -- n is the number of arguments
```
alexkeizer:
```suggestion
/-- add `n+1` bitvectors of length `w` together -/
| add (w : Nat) (n : Nat)
```

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -120,25 +120,25 @@ section Dialect
...
+| and (w : Nat) (n : Nat)
+| concat (w : List Nat) -- len(w) = #args, wi is the width of the i-th arg
```
alexkeizer:
```suggestion
/-- Concatenate `w.length` bitvectors, such that the `i`-th argument is of length `w[i]` -/
| concat (w : List Nat)
```

SSA/Projects/CIRCT/comb/comb.lean
alexkeizer: It's nicer to document this information in the docstring, so it also shows up on hover!

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -163,77 +163,84 @@ open TyDenote (toType)
 -- arg type CONF
 @[simp, reducible]
 def Op.sig : Op  → List Ty
-  | .add w => [Ty.bv w, Ty.bv w]
-  | .and w => [Ty.bv w, Ty.bv w]
-  | .concat w₁ w₂ => [Ty.bv w₁, Ty.bv w₂]
+  | .add w n => List.replicate (n + 1) (Ty.bv w)
+  | .and w n => List.replicate (n + 1) (Ty.bv w)
+  | .concat w  => w.map (fun w => Ty.bv w)
```
alexkeizer:
```suggestion
  | .concat w => w.map (fun w => Ty.bv w)
```

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -163,77 +163,84 @@ open TyDenote (toType)
...
 
 instance : DialectSignature Comb := ⟨Op.signature⟩
 
+def HVector.toList : HVector f (List.replicate n a) → List (f a) :=
+  sorry
+
+def Comb.variadicAdd (l : List (BitVec w)) : List (BitVec w) → BitVec w :=
+  List.foldr BitVec.add _ _
+  sorry
+
```
alexkeizer: Let me know if you need help filling in these sorries! Especially the first one might be a bit tricky

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,539 @@
+import SSA.Core.MLIRSyntax.EDSL
+import SSA.Projects.CIRCT.Stream.Stream
+import SSA.Projects.CIRCT.Stream.WeakBisim
+
+
+namespace Comb
```
alexkeizer: This could use a doc-comment explaining that this file defines the `comb` dialect, possibly with a link to the circt reference

alexkeizer: Also, it seems the file is called `comb/comb.lean`, but we generally capitalize file-names, so it should be `Comb/Comb.lean`

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,539 @@
...
+
+
+def add {w : Nat} (l : List (BitVec w)) : BitVec w :=
```
alexkeizer:
```suggestion

/-- Sum a list of bitvectors; representing a variadic `add` operation -/
def add {w : Nat} (l : List (BitVec w)) : BitVec w :=
```
spurious whitespace and missing docstring

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,539 @@
...
+  List.foldr BitVec.and (0#w) l
+
+/-- Concatenate a list of bitvecs `xs`, where the length of bitvec xs[i] is given by
+  element ls[i] in a list of nat `ls`-/
+def concat {ls : List Nat} (xs : HVector BitVec ls) : BitVec (List.sum ls) :=
+  match (xs) with
+  | (.nil) => 0#0
+  | (.cons x xs) =>
+     let xsSum := concat xs
+     x ++ xsSum
```
alexkeizer:
```suggestion
def concat {ls : List Nat} : HVector BitVec ls → BitVec (List.sum ls)
  | .nil => 0#0
  | .cons x xs =>
     let xsSum := concat xs
     x ++ xsSum
```
No need to `match` on one of the function arguments, there is a shorter syntax for that

alexkeizer: You might have to match on `ls` as well, to make the pattern matcher happy. To do that, simply put `ls` after the `:`, and then add underscores before each match pattern (i.e., `| _, .nil => ... | _, .cons x xs => ...`)

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,539 @@
...
+  BitVec.sdiv x y
+
+def divu (x y : BitVec w) : BitVec w :=
+  BitVec.udiv x y
+
+def extract (x : BitVec w) (lb : Nat) : BitVec (w - lb) :=
+  BitVec.truncate (w - lb) (BitVec.ushiftRight x lb)
+
+def icmp (x y : BitVec w) (op : Nat) : Bool :=
+  match op with
```
alexkeizer: Defining `op` as a `Nat` is a big no-no: you should use an enum/inductive type for this. See [how the LLVM dialect](https://github.com/opencompl/lean-mlir/blob/4ffef329226f4f22004389b30f695ed8dc206301/SSA/Projects/InstCombine/LLVM/Semantics.lean#L416) does this for inspiration

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,539 @@
...
+  | 3 => BitVec.sle x y
+  | 4 => x.toInt > y.toInt
+  | 5 => x.toInt ≥  y.toInt
+  | 6 => BitVec.ult x y
+  | 7 => BitVec.ule x y
+  | 8 => x.toNat > y.toNat
+  | 9 => x.toNat > y.toNat
+  /- ceq/cne = SV case in/equality ("===")
+    tests 4-state logical in/equality (0, 1, x, z)
+    TODO: model the concept.
```
alexkeizer: Rather than say `TODO: model the concept` I'd add a comment here to clarify that we currently only model the 0 and 1 states, not the full 4-valued logic. It's not clear what 'the concept' in this note refers to.

If you really want the todo, I'd add it in the TyDenote class (which presumably denotes into `bitvec w`) as a "TODO: model `x` and `z` states for the full 4-valued logic" or some such

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,539 @@
...
+  -/
+  | 10 => x == y
+  | 11 => x == y
+  /- weq/wne = SV wild card inequality ("==?")
+    also based on 4-state logical in/equality,
+    with x and z values acting as wild cards.
+    TODO: model the concept.
+  -/
```
alexkeizer: See preceding comment

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,539 @@
...
+
+end Comb
+
+namespace MLIR2Comb
+
+section Dialect
+
+inductive Op
+| add (w : Nat) -- n is the number of arguments
+| and (w : Nat)
```
alexkeizer: Is this right? This seems to be a simple binary `add`, rather than the variadic one

SSA/Projects/CIRCT/comb/comb.lean
```diff
@@ -0,0 +1,539 @@
...
+
+def mul {w : Nat} (l : List (BitVec w)) : BitVec w :=
+  List.foldr BitVec.mul (1#w) l
+
+/- according to the docs the input of mux
+  can be any type. I'll stick to bitvecs for now. -/
+def mux (x y : BitVec w) (cond : Bool) : BitVec w :=
+  match cond with
+  | true => x
+  | false => y
```
alexkeizer: Why limit to bitvecs? It's not hard to at least define the semantics generically
```suggestion
def mux (x y : α) (cond : Bool) : α :=
  bif cond then x else y
```

ymherklotz (approved):
This looks good to me :) Having `n` replicate functions is probably the right way to do it too.

luisacicolini:
@alexkeizer @ymherklotz thanks for the review. I've added the `enum`for the `icmp` op, will add one more example and then I believe we should be finally good to go :)

luisacicolini:
closing in favour of #1112

PR #958, opened 2025-01-13
feat: add overflow semantics according to smtlib and map it to llvm semantics

luisacicolini:
will update `Semantics.lean` once the overflow ops are merged into lean


luisacicolini:
I'm in the process of proving the equivalence between smt and llvm semantics, will merge once that happens

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -8,6 +8,41 @@ import Mathlib.Tactic.Tauto
...
 
+/-- Overflow predicate for 2's complement unary minus.
+
+  SMT-Lib name: `bvnego`.
+-/
+
+def not_overflow {w : Nat} (x : BitVec w) : Bool := x.toInt == - (2 ^ (w - 1))
+
+/-- Overflow predicate for unsigned addition modulo 2^m.
+
```
tobiasgrosser:
```suggestion
/-- Overflow predicate for unsigned addition modulo 2^m.
```

You are using both `m` and `w`, choose one. also below.

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -93,13 +128,21 @@ structure NoWrapFlags where
...
     none
-  else if flags.nuw ∧ (x' + y' < x' ∨ x' + y' < y') then
+  else if flags.nuw ∧ uadd_overflow x' y' then
     none
   else
     add? x' y'
 
+  def sadd_overflow_eq {w : Nat} (x y : BitVec w) :
+      (x.msb = y.msb ∧ (x + y).msb ≠ x.msb) = (sadd_overflow x y) := by
+    sorry
```
tobiasgrosser: Can you set w to 32 and see if bv_decide proves it?

tobiasgrosser: We do not necessarily need a symbolic proof.

luisacicolini: I am getting: 
```lean
The prover found a potentially spurious counterexample:
    - It abstracted the following unsupported expressions as opaque variables: [BitVec.ofBool (sadd_overflow x y)]
    Consider the following assignment:
    x = 0xffffffff#32
    y = 0xffffffff#32
    BitVec.ofBool (sadd_overflow x y) = 0x1#1
```

bollu: Try `simp[sadd_overflow]` first so that `bv_decide` can see the defn. As it says:

>  It abstracted the following unsupported expressions as opaque variables: [`BitVec.ofBool (sadd_overflow x y)`]

Therefore, it doesn't know how to interpret the meaning of `sadd_overflow`.

tobiasgrosser: This won't help. The definitions are in function of `toNat` or `toInt`. We need a proof to sth that BVDecide can reason about.

bollu: OK, then we need to write a proof :)

tobiasgrosser: Luisa posted the relevant statements on zulip. I hope the proof will follow soon.

SSA/Projects/InstCombine/ForLean.lean
```diff
@@ -3,6 +3,257 @@ import SSA.Projects.InstCombine.LLVM.Semantics
...
+    (x.bmod y) < 0 ↔ (y + 1) / 2 ≤ x := by
+  simp [Int.bmod_def]
+  rw [Int.emod_eq_of_lt (by omega) (by omega)]
+  split
+  · omega
+  · omega
+
+@[simp]
+theorem toInt_add_mod_elim {x y : BitVec w} :
+    (x.toInt + y.toInt) % 2 ^ w = x.toInt + y.toInt := by
```
alexkeizer: This theorem is not provable!

Consider 
```lean
#eval
  let w := 4
  let x := 8#w
  let y := 8#w
  (x.toInt + y.toInt) % 2 ^ w = x.toInt + y.toInt
```

Which makes sense, `_ % 2 ^ w` will shift the int to the range `0 .. 2 ^w`, so if the result of the addition is negative the modulus cannot be a no-op.

SSA/Projects/InstCombine/ForLean.lean
```diff
@@ -3,6 +3,257 @@ import SSA.Projects.InstCombine.LLVM.Semantics
...
+  clear ltx lty
+  --rw [Int.add_comm, Nat.two_pow_pred_mul_two] at h
+  omega
+
+theorem uadd_overflow_eq {w : Nat} (x y : BitVec w) :
+    uadd_overflow x y = BitVec.carry w x y false := by
+  simp only [uadd_overflow, BitVec.carry]
+  by_cases h : 2 ^ w ≤ x.toNat + y.toNat <;> simp [h]
+
+theorem aaaa : (2 ^ w + 1) / 2 = 2 ^ (w - 1) := by
```
alexkeizer:
```suggestion
theorem aaaa : (2 ^ w + 1) / 2 = 2 ^ w := by
```

That should be easier to prove

tobiasgrosser: I don't think the new theorem is true, no?

SSA/Projects/InstCombine/ForLean.lean
tobiasgrosser: Removed.

SSA/Projects/InstCombine/ForLean.lean
alexkeizer: Ah, I'd read it as `2 ^ (w + 1)`, in which case it would've been true; ignore my suggestion!

tobiasgrosser:
This is now upstream.

PR #963, opened 2025-01-15
chore: update to nightly-2025-01-14

tobiasgrosser:


tobiasgrosser:
@luisacicolini, can you fix the DC example?

luisacicolini:
I sorried and commented the offending code since the proof was complex and I don't want to block the update. Will be back to fix the sorries asap, should be good to merge in the meantime.

PR #966, opened 2025-01-15
feat: fix eval script to kill the children of the lean process

ineol:
At least on MacOS, the `runner.py` script was leaving lots of busy orphan processes


bollu:
Sure, let's pull it in!

PR #970, opened 2025-01-17
chore: turn precompilation on

ineol:


tobiasgrosser:
Can we merge this?

ineol:
I think we should check that the build still works / and that the additional compile time is worth it. CI has been at it already for 1h and it's nowhere done

ineol:
I think I know it's taking so long. It's loading one dynamic library for each module in Mathlib we depend on and I think there are so many because I imported `Mathlib.Tactic`. I'll remove that import and see how much it helps.

tobiasgrosser:
Nice. This seems to improve things.

PR #976, opened 2025-01-20
feat: make AutoStructs faster

ineol:
This PR improves the performance of the automata based decision procedure for bitvectors: it solves all the 2500 tests of dataset2.

It is achieved by:
- using HashSets in the determinization procedure
- constructing the NFA for the terms using NFA constructions instead of FSM
- (related) minimize the NFA at each step of the interpretation of terms

Note: I still need to prove the correctness of these changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ test-semantics/*.llvm
 *.olean
 lakefile.olean
 .lake/
+profile.json
--- a/SSA/Experimental/Bits/AutoStructs/Basic.lean
+++ b/SSA/Experimental/Bits/AutoStructs/Basic.lean
@@ -215,8 +215,7 @@ def RawCNFA.createSink (m : RawCNFA A) : State × RawCNFA A :=
   (s, m)
 
 def RawCNFA.transSet (m : RawCNFA A) (ss : Std.HashSet State) (a : A) : Std.HashSet State :=
-  ss.fold (init := ∅) fun ss' s =>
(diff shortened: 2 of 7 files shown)


bollu:
@ineol do you know how much each step wins you? I could very much imagine that one of the steps is quite beneficial for the model-checking based approach as well!

bollu:
Very cool, BTW :D Brings me hope that the model checker can be made much faster by stealing your insights ;)

ineol:
> @ineol do you know how much each step wins you? I could very much imagine that one of the steps is quite beneficial for the model-checking based approach as well!

I was testing on the 500 first examples of dataset2. 
1. Baseline (with precompilation) was 428/500
2. With hashset in the determinization was 475/500
3. When I switched to using NFA ops to build NFA for the terms instead of evaluating the FSM it went down to 473/500
4. When I minimized in the intermediate results I got 100%

In step 3 the size of the NFA for one of the test case (the one if FastCopy/Profile.lean) was around 18k, and double in step 2 I believe. To test I minimized once at the end and I got 28 states. I'd like to find out what is the maximal size  of the intermediate NFAs, as it's well known that the intermediate steps create much bigger NFAs than the result .

So I think the big win is keeping the size in check by minimizing before it blows up. I guess the equivalent would be to (use SAT based techniques to) simplify the boolean expressions as we're building the FSM?

ineol:
So I just checked, and on large examples the intermediate increase to around 100 states.

bollu:
LGTM since this is in `Experimental`!

PR #984, opened 2025-01-26
feat: Try to build smaller circuit for bv_automata_circuit

bollu:
I believe the old circuit was establishing a much stronger inductive invariant than necessary, of the form $$S(0) \land S(1) \dots \land S(k) \implies S(0) \land S(1) \dots \land S(k) \land S(k + 1)$$.

We now establish the invariant $$S(0) \land S(1) \dots \land S(k) \implies S(k + 1)$$ which is of course sufficient. This should build a much smaller SAT formula than previously.


bollu:
This showed no performance improvement, abandoning for other ideas, but the simplification will be adapted in a subsequent PR

PR #989, opened 2025-01-28
feat: add correct k-induction algorithm.

bollu:
Insight: recall that correctness proof increased on thenumber of boolean
outputs where it produced true. The new algorithm ensures that this
invariant is maintained. In the kth step, it shows that either
$\texttt{safe}(s) \implies \texttt{safe}(s + 1) \lor (\texttt{safe}(s) \land \texttt{safe}(s+1) \implies \texttt{safe}(s+2)) \lor (safe(s) \land safe(s+1) \land \dots \land safe(s + k) \implies safe(s+k+1)$ ,
as well as $\texttt{safe}(s0) \land \texttt{safe}(s0 + 1) \land \dots  \texttt{safe}(s0 + k)$.

The fact that we keep the previous inductive invariants around
makes the solver only try to proceed in the cases where depth $k$
was insufficient, thereby ensuring that we always incraese the
number of states we are proving to be \texttt{safe}, or finding a cex thereof.
Thus, termination is guaranteed.

Also recall that the space of boolean functions has a trivial upper
bound, which this process will eventually reach (thanks @hargoniX).


alexkeizer:
Mysterious indeed, you've made me curious what this PR is actually about :rofl:

PR #992, opened 2025-01-31
chore: update nightly-testing-2025-01-30

luisacicolini:
Sadly, a few things broke. I will work on them over the day and push frequently to avoid overlaps, any help would be great, especially @ineol if you could take a look! Quite a lot of errors are raised in `ForMathlib` and it might be a while before I manage to fix that. 

I will start by fixing `ForLean`.


ineol:
I'll take a look today

ineol:
Should all be fixed!

luisacicolini:
that's perfect, thank you so much @ineol :)

PR #996, opened 2025-02-04
chore: Instcombine tailor to automata

bollu:
Change instcombine scripts to only produce problems that are within the presburger fragment


bollu:
Superceded by https://github.com/opencompl/lean-mlir/pull/997

PR #1023, opened 2025-03-11
feat: alternative, more robust EDSL elaborator 

alexkeizer:
This PR implements `elabIntoCom'` as an alternative to `elabIntoCom` which follows the following strategy:
1) Evaluate the AST object into a meta-time object, 
2) Then, *run* the AST -> Com parsing code at meta-time
3) Turn the meta-time Com object into an expression via `ToExpr`

Most of the changed lines of this PR are to actually implement the `Com` to `Expr` transformation.
Replacing kernel reduction with metatime computation should be much more reliable, as it removes the need for `unseal`ing specific irreducible definitions. The transformation is made generic over a `DialectMetaMorhpism`, which just bundles `Op -> Expr` and `Ty -> Expr` transformations. This will be used in, for instance, the LLVM elaborator to also instantiate metavariables at meta-time.

See also the `comb-dialect-edsl2` branch, which demonstrated the new elaborator against #949, or #1024 which uses the new elaborator for the LLVM dialect.

### DIFF
--- a/SSA/Core/HVector.lean
+++ b/SSA/Core/HVector.lean
@@ -3,7 +3,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 -/
 import Batteries.Tactic.Basic
 import Mathlib.Tactic.TypeStar
-
+import Qq
 
 /-- An heterogeneous vector -/
 inductive HVector {α : Type*} (f : α → Type*) : List α → Type _
@@ -12,7 +12,7 @@ inductive HVector {α : Type*} (f : α → Type*) : List α → Type _
 
 namespace HVector
 
-variable {A B : α → Type*} {as : List α}
+variable {α : Type u} {A B : α → Type*} {as : List α}
 
 /-
   # Definitions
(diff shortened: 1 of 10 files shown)


bollu:
@alexkeizer what happened to the PR where you were planning on upstreaming ToExpr to lean?

alexkeizer:
> @alexkeizer what happened to the PR where you were planning on upstreaming ToExpr to lean?

It got merged! `ToExpr` now lives in core

alexkeizer:
I just merged in a few more changes and fixes I made while working on #1024; this should now be ready for review!

(Disregard the failing CI, it seems the evaluation job is automatically failing/hanging because there is no runner available)

tobiasgrosser:
@bollu, can you give this a pass?

SSA/Core/HVector.lean
```diff
@@ -134,4 +141,40 @@ macro_rules
...
 
 infixl:50 "::ₕ" => HVector.cons
+
+/-
+  # ToExpr
+-/
+section ToExpr
+open Lean Qq
+
+class ToExpr {α : Type u} (A : α → Type v) [∀ a, ToExpr (A a)] where
```
bollu:
```suggestion
class ToExprPi {α : Type u} (A : α → Type v) [∀ a, ToExpr (A a)] where
```

cause it's a name for a pi-type?

bollu:
LGTM to me @alexkeizer , also 'cause your choice of calling it morphism makes sense :) There is indeed a category of dialects with morphisms `D -> E` being the same as `DialectMetaMorphism D E`.

luisacicolini:
thank you! I have applied @bollu's suggestion, will make sure everything builds correctly and then merge

PR #1024, opened 2025-03-11
feat: move LLVM dialect to new elaborator

alexkeizer:
This PR changes the `[llvm ..]` macro to use the new elaborator from #1023, reimplementing the metavariable instantiation in metacode. This has the benefit that the elaborator now gives the fully instantiated program, removing the need for (brittle) automation like `simp_alive_meta`. Anecdotally, elaboration of `AliveAutoGenerated.lean` now also feels a decent bit faster.


SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
```diff
@@ -239,14 +239,15 @@ private def pretty_test_exact :=
...
-  simp_alive_meta
   rfl
 
 example : pretty_test_generic = prettier_test_generic    := rfl
 
 /-! ## antiquotations test -/
 
+/-
+TODO: antiquotations don't work with the new elaborator ;(
+
```
tobiasgrosser, line 252: What precisely do we loose here? Can we still express bitwidth generic rewrites?

SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
```diff
@@ -239,14 +239,15 @@ private def pretty_test_exact :=
   }]
 
 example : pretty_test = prettier_test_generic 32 := by
-  unfold pretty_test prettier_test_generic
-  simp_alive_meta
```
tobiasgrosser, line 243: Very cool. I would like a demo, maybe in our 1:1 today?

SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
alexkeizer: > Can we still express bitwidth generic rewrites?

Yes, we have a deep embedding of meta-variables that allow us to express these generic rewrites.
This deep embedding works as intended in the new elaborator.
We lose the ability to piggy-back on Lean's syntax-antiquotations to construct "arbitrary" parts of the MLIR AST (in "s because in practice we could only use such anti-quotations in particular spots or risk horribly breaking the elaborator).

In practice, we only used this ability to insert arbitrary constant values, as in the example. We could still recover this ability as part of the new elaborator, but we'd have to put in the effort to enable our deep embedding to be used for constant values as well.

In general, it means that dialect implementations that want to depend on Lean variables for particular components of their programs will have to put in the effort to make such a deep embedding (as we did by having `MetaLLVM` and `LLVM` dialects). The benefit being that once the effort has been put in, there is a very clear distinction between what is robustly supported and what is not supported at all.

tobiasgrosser: This sounds great to me.

SSA/Projects/InstCombine/LLVM/PrettyEDSL.lean
```diff
@@ -239,14 +239,15 @@ private def pretty_test_exact :=
...
   rfl
 
 example : pretty_test_generic = prettier_test_generic    := rfl
 
 /-! ## antiquotations test -/
 
+/-
+TODO: antiquotations don't work with the new elaborator ;(
```
alexkeizer:
```suggestion
TODO: Constant value antiquotations don't work with the new elaborator.
            To properly support these test-cases, we should include constant values
            into the existing deeply-embedded metavariable instantiation
```

PR #1036, opened 2025-03-16
feat: do not precompile in CI

tobiasgrosser:


tobiasgrosser:
Precompilation is slow. This is unfortunate. Let's close this for now. Potentially reconsider if we figure out which tests run by default.

PR #1062, opened 2025-03-17
chore: downgrade toolchain to nightly-2025-03-15, to match Mathlib

alexkeizer:
Currently, `lake exe cache get` complains about the mismatched toolchains and refuses to work, hence let's downgrade to match Mathlib.

### DIFF
--- a/lean-toolchain
+++ b/lean-toolchain
@@ -1,1 +1,1 @@
-leanprover/lean4:nightly-2025-03-16
+leanprover/lean4:nightly-2025-03-15


tobiasgrosser:
The mathlib cache anyway does not work as we have `precompileModules = true` set.

tobiasgrosser:
You likely will need to revert more to get back to 03-15. I would suggest to wait for the next mathlib nightly and we will be back in sync.

alexkeizer:
> The mathlib cache anyway does not work as we have `precompileModules = true` set.

That's unfortunate, but I guess removes the need for this PR

tobiasgrosser:
The use of precompileModules is something work discussion. Alternatively, we can investigate if we can run a CI that builds these precompiledModules-enabled mathlib. This would streamline our CI quite a bit.

  # Privilege required
  put          Run 'mk' then upload linked files missing on the server
  put!         Run 'mk' then upload all linked files
  put-unpacked 'put' only files not already 'pack'ed; intended for CI use
  commit       Write a commit on the server
  commit!      Overwrite a commit on the server
  collect      TODO
  
  Not sure if we can just upload files to the servers.

tobiasgrosser:
Lake exe has some support for this.

tobiasgrosser:
This should now be in sync.

PR #1064, opened 2025-03-18
feat: `def_signature` elaborator for convenient dialect signature definition

alexkeizer:
This PR adds a convenience syntax `def_signature` to define an instance of `DialectSignature` for a dialect with a slightly prettier syntax.

I've also changed the `PaperExamples` dialects to demonstrate this syntax. For example, the example dialect with regions has it's signature defined as follows:
```lean
def_signature for SimpleReg
  | .const _    => () → .int
  | .add        => (.int, .int) → .int
  | .iterate _  => { (.int) → .int } → (.int) -[.pure]-> .int
```

Read as: "iterate" has one region, of type `(int) -> int`, and one regular argument, of type `int`, and returns an `int`, without performing any side-effects (i.e., it is pure). Note that the purity annotation on the last line is redundant: the default assumption when using regular arrows (`->`) is that the operations is pure. I just used the other arrow to showcase the syntax.

As alluded to by a TODO in the code, I've thought about automatically opening up relevant namespaces, so that we could write `const` and `ty` instead of `.const` and `.ty`. However, on second thought I'm not sure that's worth the effort, as the dot-notation is already pretty easy to use, and it's not immediately obvious which namespace to open if the `Ty` is itself a `def` (as is the case for `LLVM.Ty`).

### DIFF
--- a/SSA/Core/Framework/Macro.lean
+++ b/SSA/Core/Framework/Macro.lean
@@ -4,7 +4,25 @@ Released under Apache 2.0 license as described in the file LICENSE.
 import SSA.Core.Framework
 import SSA.Core.Framework.Trace
 
-/-! ## `def_signature` macro -/
+/-! ## `def_signature` elaborator
+This file defines `def_signature` elaborator that makes it easier to define the
+signature of operations in a LeanMLIR dialect.
+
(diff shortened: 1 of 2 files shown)


tobiasgrosser (approved):
This looks beautiful. 🤩

PR #1066, opened 2025-03-20
fix: reduce concurrency of llvm evaluation CI

alexkeizer:
For some reason, running with a large number of workers results in a reduction of instances that get reported as solved. The build logs don't show any particular errors, rather, they seem like they cut short.

I hypothesize that somehow the high concurrency is running against some resource limit (memory, perhaps?) and the compile processes are getting killed. Reducing the number of workers used by the evaluation CI seems to bring us back to regular numbers. We also increase the timeout, to account for the fact that CI will now likely take longer.

### DIFF
--- a/.github/workflows/evaluation.yml
+++ b/.github/workflows/evaluation.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Run LLVM
         continue-on-error: true
         run: |
-          (cd bv-evaluation; python3 ./compare-leansat-vs-bitwuzla-llvm.py -j128)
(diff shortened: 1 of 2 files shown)


alexkeizer:
I've opened this PR to double check that reducing the number of workers also results in the original numbers on the latest commit (the previous run with a lower concurrency and good numbers was on the commit that introduced the regression).

tobiasgrosser:
This has been resolved in https://github.com/opencompl/lean-mlir/pull/1067.

PR #1071, opened 2025-03-21
chore: nightly-testing-2025-03-20

tobiasgrosser:


alexkeizer:
I had a look, but it doesn't seem like our meta-code/EDSL elaborator is the problem here. I suspect something changed in the reducability of the rewrite engine, and that's why the `rfl` proof no longer works.

PR #1083, opened 2025-03-25
feat: implement `def_denote` elaborator for pretty dialect semantics definition

alexkeizer:
Expand on the existing `def_denote` elaborator, by
* implementing support for variadic argument list via an escape hatch

Then, implements a `def_denote` elaborator for defining dialect semantics via idiomatic curried functions, which are then transformed into the uncurried form expected by the framework. Additionally, region arguments, which would typically be uncurried functions, are automatically prettified into curried functions.


alexkeizer:
I'm merging this as there's been no complaints and I don't want the PR to get too stale.

@bollu feel free to revert / post-facto review if you find any big problems.

PR #1088, opened 2025-03-27
nightly-testing-2025-03-25

tobiasgrosser:


tobiasgrosser:
@ineol, this seems to break the worklist algorithm. Could you take a look?

ineol:
Should be fixed

PR #1090, opened 2025-03-31
chore: merge the two automata frontends

ineol:
This should also fix the alive examples thanks to the more robust preprocessing.


ineol:
@bollu There are two admits in Predicate.eval_eq_denote because I changed the definition of `Predicate.denote` to treat unsigned and signed comparison the same way.

PR #1110, opened 2025-04-08
chore: re-enable native compilation

ineol:


tobiasgrosser:
Lean 2024-04-16 still yields: 😢 

> info: stderr:
> ibc++abi: terminating due to uncaught exception of type lean::exception: error loading library, libLake_shared.so: cannot open shared object file: No such file or directory
> error: Lean exited with code 134

ineol:
I've been meaning to try on a Linux machine, because it works on my macbook, and it's strange that it tries to load Lake anyways.

ineol:
It's very strange, because it worked on our Linux machine...

tobiasgrosser:
@ineol, is this still something you would like to enable in CI?

tobiasgrosser:
In fact, https://github.com/opencompl/lean-mlir/pull/1415 seems more up-to-date.

PR #1111, opened 2025-04-13
fix: DCE example and add proof that the new example is the result of DCE

salinhkuhn:


alexkeizer:
@Sarah, we usually prefix a PR name with something like `feat:`, `refactor:`, `fix:` or `chore:` (I'm not sure what the complete list is; our convention is a not very well-defined copy of mathlib / core Lean conventions). We also tend to write PR titles/description in the present tense (as in "This PR fixes foo and adds bar"). 

I've renamed your PR to "fix: DCE example and add proof that the new example is the result of DCE".

Otherwise LGTM. The CI failure is not actually related to your change; it's a more general failure that was introduced to the main branch at some point, so I've hit merge for this PR.

PR #1112, opened 2025-04-14
feat: add `comb` dialect

luisacicolini:
This PR adds the [`comb` dialect](https://circt.llvm.org/docs/Dialects/Comb/#combadd-circtcombaddop). 
`truthTable` operation and wildcard/cases SV semantics are missing, as they will require more infrastructure, e.g. establishing and defining 4-value logic. 

I have no idea what's wrong with the `InstCombine` files, git seems to think that there's been some changes but they look identical to me (and I did `git restore` the path, no changes)


SSA/Projects/InstCombine/scripts/test-gen.py
```diff
@@ -239,4 +239,4 @@ def process_file(file):
 if __name__ == "__main__":
     remove()
     with Pool(300) as p:
-        p.map(process_file, os.listdir(directory))
+        p.map(process_file, os.listdir(directory))
```
luisacicolini, line 242:
```suggestion
        p.map(process_file, os.listdir(directory))
```

SSA/Projects/InstCombine/scripts/cfg.py
```diff
@@ -118,4 +118,4 @@ class Msg(Enum):
     E_PTRARG = 9
     
     def is_error(self):
-        return self.value > 2
+        return self.value > 2
```
alexkeizer, line 121: Seemingly unrelated whitespace change. You've got a few more of these in other files also

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -0,0 +1,617 @@
...
+  | .ule => (x ≤ y)
+
+
+/-- Signed modulo -/
+def mods (x y : BitVec w) : BitVec w :=
+  BitVec.smod x y
+
+/-- Unsigned modulo -/
+def modu (x y : BitVec w) : BitVec w :=
+  BitVec.umod x y
```
alexkeizer: Do we need all of these to be separate definitions? I believe you only use them in `def_denote`, right?

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -0,0 +1,617 @@
...
+  | .and _ _ => fun xs => CombOp.and (HVector.replicateToList (f := TyDenote.toType) xs)
+  | .concat _ => fun xs => CombOp.concat xs
+  | .divs _ => fun xs => CombOp.divs xs
+  | .divu _ => fun xs => CombOp.divu xs
+  | .extract _ n => fun xs => CombOp.extract xs n
+  | .icmp p _ => fun xs =>
+      let p' := Option.get! (ofString? p)
+      CombOp.icmp p' xs
+  | .mods _ => fun xs => CombOp.mods xs
+  | .modu _ => fun xs => CombOp.modu xs
```
alexkeizer:
```suggestion
  | .mods _ => BitVec.mods
  | .modu _ => BitVec.modu
```
Could we not just write it like so? Removing the need for all those `CompOp` definitions that just wrap a BitVec operation.

Also, the naming of your argument is a bit misleading. `xs` suggest a list or vector, but I *think* it's just a `(xs : BitVec w)`, right? Then you apply `CombOp.mods xs` implying that `mods` is somehow operating on a list of values, but what's really happening (IIUC) is that you're partially applying the first argument of a binary BitVec def, returning a `BitVec _ -> BitVec _` function, and `def_denote` is happy with that. I'd suggest either going all in on point-free, and just writing `BitVec.mods` (or `CombOp.mods` if you don't take the suggestion) *or* fully explicate all arguments, as in `fun x y => BitVec.mods x y`.

The same applies to all other non-variadic operations.

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -0,0 +1,617 @@
...
+  | .mul _ _ => fun xs => CombOp.mul (HVector.replicateToList (f := TyDenote.toType) xs)
```
alexkeizer, line 106: For the variadic operations, I'd suggest writing a `fold` combinator, that combines `HVector.replicateToList` with `List.foldr`, so that we can then write the variadic operations as:
```
| .add _ _ => CombOp.fold (· + ·) 0#_
| .mul _ _ => CombOp.fold (· * ·) 1#_
```

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -0,0 +1,617 @@
...
+/-- we currently do not support the parsing of heterogeneous vectors -/
+def mkTy : MLIR.AST.MLIRType φ → MLIR.AST.ExceptM Comb Ty
+  | MLIR.AST.MLIRType.undefined s => do
+    match s.splitOn "_" with
+    | ["Bool"] =>
+      return .bool
+    | ["Nat"] =>
+      return .nat
+    | ["BitVec", r] =>
+      return .bv (r.toNat!)
```
alexkeizer: You might want to match on `MLIRType.int` instead, and map that to your `.bv` type (have a look at LLVM, I believe that's what we do there as well).

That way you can use the builtin `i64` type in your programs (which I believe CIRCT uses for comb, right?), and also you won't have to use `String.splitOn` here anymore, you can just directly match on the strings for the other cases.

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -0,0 +1,617 @@
...
+    (args := a)
+    (regArgs := .nil)
+
+def concat {l : List Nat} {Γ : Ctxt _} (ls : Γ.Var (.hList l)): Expr (Comb) Γ .pure (.bv (l.sum)) :=
+  Expr.mk
+    (op := .concat l)
+    (ty_eq := rfl)
+    (eff_le := by constructor)
+    (args := .cons ls <| .nil)
+    (regArgs := .nil)
```
alexkeizer: Here again you've got a huge number of boilerplate definitions. Do we actually need these? I know we have similar constructors in other dialects where we didn't bother writing a parser, but for Comb you did write a parser, so there should be no need for these defs.

Of course, I see you've used them in the definition of said parser, but you should be able to write that much more concisely.

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -0,0 +1,617 @@
...
+| [], h => .nil
+| ⟨ty', var⟩::rest, h =>
+  have hty : ty' = ty := by simp_all
+  have hrest : rest.all (·.1 = ty) := by simp_all
+  .cons (hty ▸ var) (ofList _ rest hrest)
+
+def ofList' {α : Type u} {f : α → Type v} :
+    (xs : List (Σ a, f a)) → HVector f (xs.map Sigma.fst)
+  | ⟨_, x⟩ :: xs => .cons x (ofList' xs)
+  | [] => .nil
```
alexkeizer: This should likely go to the HVector file. Also, do we need both? They seem very similar.
If we do need both, we should have some docstring that points out the difference.

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -0,0 +1,617 @@
...
+  match opStx.name with
+  | op@"Comb.parity" | op@"Comb.concat" =>
+    match opStx.args with
+    | v₁Stx::[] =>
+      let ⟨ty₁, v₁⟩ ← MLIR.AST.TypedSSAVal.mkVal Γ v₁Stx
+      match ty₁, op with
+      | .bv w, "Comb.parity" => return ⟨_, .bool, parity v₁⟩
+      | .hList l, "Comb.concat" => return ⟨_, .bv l.sum, concat v₁⟩
+      | _, _ => throw <| .generic s!"type mismatch"
+    | _ => throw <| .generic s!"expected one operand for `monomial`, found #'{opStx.args.length}' in '{repr opStx.args}'"
```
alexkeizer: Your error mentions `monomial`!

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -0,0 +1,617 @@
...
+              return ⟨_, .bv w, and args.length (heq ▸ argsᵥ)⟩
+            else
+              throw <| .generic s!"Unexpected argument types for '{repr opStx.args}'"
+        | ⟨.bv w, _⟩, "Comb.mul" =>
+            if hall : args.all (·.1 = .bv w) then
+              let argsᵥ := ofList (.bv w) _ hall
+              have heq : args.length - 1 + 1 = args.length := by omega
+              return ⟨_, .bv w, mul args.length (heq ▸ argsᵥ)⟩
+            else
+              throw <| .generic s!"Unexpected argument types for '{repr opStx.args}'"
```
alexkeizer: This code is highly repetitive and doing more-or-less the same thing in each branch, modulo the specific opcode emitted, there should be some function we can factor out of this to reduce the duplication.

SSA/Projects/InstCombine/scripts/cfg.py
luisacicolini, line 121:
```suggestion
        return self.value > 2
```

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: that is correct - I maintained the definitions separate to mirror exactly the operators of `Comb`.

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: ah yes I had not thought of that! Indeed those operations are not variadic. I've updated that for the cases where bitvec operations exist and left the remaining ones unchanged (modulo the naming suggestion which I've applied). In some cases the corresponding bitvec operation does not exist, e.g., `BitVec.shiftLeft` takes a bitvec and a nat, while `comb.shl` takes two bitvecs, so I could not use the `BitVec` API. (The same happened for the other ops that have not been updated.)

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: [`comb`](https://circt.llvm.org/docs/Dialects/Comb/#combshl-circtcombshlop) actually uses bitvecs for almost everything afaict, so it would be nice to maintain that I guess?

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: ah no I forgot to delete one - I just wanted to keep both the implementations you and Leo proposed! will delete one.

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: fixed

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: done

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: Do you think this would be better than having the `fold` in the definition of each variadic op? I thought it would be better to have each op in the denotation referring to its `CombOp` semantics

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: could you please make an example? I don't understand how these can be avoided

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: of course, I will extract some functions depending on the number of operands I guess

SSA/Projects/CIRCT/Comb/Comb.lean
ymherklotz: What does this mean for the denotation of concrete programs? With the unfolded version does it mean we get unfolded definitions in the final denotation, i.e will the fold also completely disappear?

If not, I think I agree with Luisa that wrapping them into CombOp definitions might be nice to then use further rewrites to unfold them, but I'm guessing that the current denotation should already unfold everything correctly, in which case the intermediate definition is maybe unnecessary.

SSA/Projects/CIRCT/Comb/Comb.lean
ymherklotz: Im guessing that the definitions can just be completely inlined, because they are only used in the parser and not needed again. And in addition to that the main thing that changes in each definition is the constructor and number of arguments, so it might be possible to make this much more concise using some helper functions.

ymherklotz:
I just wanted to say that to me the definition looks quite good already, so I was wondering if it should already be merged before trying to make definitions more concise, as that will maybe be a bit more work, and isn't critical for the semantics?

luisacicolini:
@ymherklotz I agree - as we use the dialect we'll likely change a few things anyways, @alexkeizer let me know if you agree!

SSA/Projects/CIRCT/Comb/Comb.lean
alexkeizer: Yes, "signless integer bitvector" in the documentation means `i$x`, for some `$x`. See, e.g., the [canonicalization](https://github.com/llvm/circt/blob/main/test/Dialect/Comb/canonicalization.mlir) test file which has the following example program:
```
hw.module @idempotentDeduped1(in %arg0 : i7, in %arg1 : i7, out resAnd: i7, out resOr: i7) {
  %0 = comb.and %arg0    : i7
  %1 = comb.and %0, %arg1 : i7
  %2 = comb.or %arg0    : i7
  %3 = comb.or %0, %arg1 : i7
  hw.output %1, %3 : i7, i7
}
```

LeanMLIR's generic MLIR AST parser special-cases the `i...` types, since they're builtin, and parses them into the `MLIRType.int` constructor (which has extra parameters for the signed-ness and width). I guess we ought to improve the documentation of that type to emphasize it means "int" in the fixed-width `i$x` sense rather than true integers.

alexkeizer:
Yeah, that makes sense! In general, I think we settled on a policy that stuff that doesn't affect the core framework (and doesn't break CI) is fine to merge without review / explicit approval.

SSA/Projects/CIRCT/Comb/Comb.lean
alexkeizer: Hmm, that's a good point. What do you mean by "final denotation" exactly?

If you mean "the result of simplifying with `simp_peephole`": by default it should unfold to exactly the rhs of `def_denote` applied to a concrete HVector of arguments. The intention being that dialect implementors then add further, dialect-specific rewrites as needed. For example, we could add a rewrite like
```
CombOp.fold (· + ·) 0#_ (x ::h xs) = x + CombOp.fold  (· + ·) 0#_ xs
```
plus the analogue for an empty vector, and it should then unfold the variadic operations as desired.

Speaking off, I've been slightly tweaking how that tactic works in #1127, so that adding dialect-specific rewrites becomes much easier. Currently, you'd have to wrap `simp_peephole` with a dialect-specific `simp_comb_peephole` that passes in the extra rewrites; after #1127 gets merged you'd only have to tag your dialect-specific rewrite with `@[simp_denote]` and it gets picked up by plain `simp_peephole` automatically.

SSA/Projects/CIRCT/Comb/Comb.lean
alexkeizer: Fair point. I personally dislike having huge files with a lot of boilerplate, hence my suggestion to eliminate them. Still, I'll concede that having a clear API boundary has benefits as well.

In that case, I'd suggest putting these wrappers in a separate file (akin to how we have LLVM/Semantics.lean), as I feel that's a bit more maintainable. In a similar vein, I'd have personally put the parsing code in a separate file, to keep a cleaner organization. This is partly based on personal opinion, though, so as always feel free to ignore if you disagree.

luisacicolini: happy to split it into different files, sounds very reasonable.

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: Okay I'll refactor the parser to extract some functions based on the number of argumens and inline this as well

luisacicolini:
I'll split the semantics and parser into different files (eta: tomorrow morning), inline the boilerplate in the parser and then merge.

luisacicolini:
opened #1152 in case more refactoring is needed. will merge this to continue with the work.

PR #1126, opened 2025-04-17
chore: move `DC` to new semantics

luisacicolini:
This PR updates the `DC` dialect to use the new `def_signature` and `def_denote`. I also went through the existing proofs and commented the broken ones (likely broken due to the last months' lean updates since we're not running those in CI: I will investigate them and open a new PR to fix those proofs over the next weeks)

There is one `sorry` in the definition of `pair`: I thought about it some time and haven't found a way to get that right.


luisacicolini:
As per the discussions in #1112 and #1128 I removed the `pair` operation and updated the names in the denotation

luisacicolini:
Since I plan on eventually implementing `pair`, I created an issue (#1148) but will in the meantime merge this PR to continue with the work.

PR #1127, opened 2025-04-17
refactor: enhance `simp_peephole` so that we won't need dialect-specific wrappers like `simp_alive_peephole`, make `simp_peephole at _` behave like `simp at`

alexkeizer:
This PR refactors the `simp_peephole` tactic in a few mayor ways:
* We register a `simp_denote` simpset that contains all the simp-lemmas `simp_peephole` used to use, but is easily extensible. This means we don't need the `simp_peephole [ts,...]` syntax to pass in extra rewrites anymore, as those rewrites ought to be added to the `simp_denote` simpset anyway.
    * We hook into the `def_denote` elaborator to make it add instances of `DialectDenote` added in this way to the `simp_denote` simpset automatically.
* We replace the `generalize_or_fail` hack we used in `simp_peephole at V` with a more robust simproc (`elimValuation`) that rewrites a universal quantifier `∀ (V : Valuation [t₁, ..., tₙ])` where the context of the valuation is a concrete list of types, into a sequence of quantifiers `∀ (x0 : ⟦t₁⟧) (x1 : ⟦t₂⟧) ... (xn : ⟦tₙ⟧)`
    * Any uses of `simp_peephole at V` should thus be replaced by `revert V; simp_peephole`
* Furthermore, we add a simple preprocessing step to `simp_peephole`, which can apply `funext` if the goal is a function equality, so that the aforementioned quantifier rewrite can apply automatically.
* Finally, now that `simp_peephole at _` is available, we change it's meaning to be in line with `simp at _`, i.e., allowing users to simplify semantics in hypotheses as well, not just in the goal.

Unfortunately, the new way of eliminating the valuation introduces the new variables in a different order: the new simproc adds new variables in the same order as they appear in the context, whereas the old way adds them in an unclear order, likely dependent on the order the variable appears in the program. The new order seems nicer, so this is an acceptable break and I regenerated AliveStatements accordingly.

With these changes, `simp_peephole` can do everything `simp_alive_peephole` can, so I've deprecated the latter (and it's aliaw `simp_alive_ssa`). I've thought about doing a global find/replace, but worry that might introduce too much noise in the diff; maybe we get rid of these tactics in a follow-up PR at some point.


bollu:
> maybe we get rid of these tactics in a follow-up PR at some point

Please do! This is a major quality-of-life change, thanks a lot Alex!

SSA/Core/ErasedContext.lean
```diff
@@ -714,6 +714,21 @@ instance : ToExpr (Var Γ ty) where
...
+-/
+def Valuation.mkOfElems (Ty instTyDenote : Expr) (Γ : Array Expr) (xs : Array Expr) : Expr :=
+  let nil := (
+    mkApp (.const ``List.nil [0]) Ty,
+    mkApp2 (mkConst ``Ctxt.Valuation.nil) Ty instTyDenote
+  )
+  Prod.snd <| (Γ.zip xs).foldl (init := nil) fun (Γ, V) (ty, x) => (
+      mkApp3 (.const ``List.cons [0]) Ty ty Γ,
+      mkApp6 (mkConst ``Ctxt.Valuation.snoc) Ty instTyDenote Γ ty V x
+  )
```
bollu: I personally now prefer `for` loops, but I'm happy to leave the `foldl` around.

SSA/Core/Tactic.lean
```diff
@@ -15,115 +16,124 @@ open Ctxt (Var Valuation DerivedCtxt)
...
-        The parentheses seems to enable `simp` to unify typeclass arguments as well, and thus the
-          simp-lemma applies even for non-standard instances.
-
-        One caveat: `simp [(X)]` only works if `X` is a lemma, *not* if `X` is a definition to be
-        unfolded. Thus, we replace `HVector.denote` with its equation lemmas
-          `(HVector.denote_cons)` and `(HVector.denote_nil)` -/
-        (HVector.denote_cons), (HVector.denote_nil),
-        $ts,*]
-    ))
+  withLocalDeclsDND declInfo <| fun xs => do
```
bollu, line 99: Wow, interesting naming convention, `D` for default, `ND` for non-dependent? :D

SSA/Core/Tactic.lean
```diff
@@ -15,115 +16,124 @@ open Ctxt (Var Valuation DerivedCtxt)
...
+    let V := Valuation.mkOfElems Ty instTyDenote Γelems xs
+    let newType ← mkForallFVars xs (body.instantiate1 V)
+    let proof :=
+      let mp  :=
+        ←withLocalDeclD .anonymous e <| fun eProof => do
+          mkLambdaFVars #[eProof] <|← mkLambdaFVars xs <| mkApp eProof V
```
bollu:
```suggestion
          mkLambdaFVars (#[eProof] ++  xs) <| mkApp eProof V
```

I guess you opted against this because you didn't want to pay the price of reconstructing thearray because of `++`?

bollu (approved):
LGTM, lovely cleanup <3

SSA/Core/Tactic.lean
alexkeizer: Yup, that is how it is meant to read, not my favourite naming choice either :sweat_smile:

SSA/Core/ErasedContext.lean
alexkeizer: No, you're right, the for loop is cleaner here

SSA/Core/Tactic.lean
alexkeizer: That was indeed the reason, although looking at it now I can't imagine the performance difference is really a big deal. I'll go with your suggestion

PR #1128, opened 2025-04-17
chore: move `Handshake` to new semantics

luisacicolini:
This PR updates the `Handshake` dialect to use the new `def_signature` and `def_denote`. The examples still work, except the ones that I previously fixed in #1126. I also complete the signature of some operations that were missing.
There is one `sorry` in the denotation of `source`, which I could not fix.


alexkeizer:
`source` is a new operation that was not previously modeled, right? In that case, I'd suggest just commenting it out of the `Op` definition; having a sorry in the `def_denote` semantics seems like it'd cause problems.

Otherwise, LGTM

luisacicolini:
@alexkeizer thanks! the semantics `source` existed already, but it was not denoted. Happy to remove it anyways, does not seem to be particularly used.

luisacicolini:
removed `source` and improved the denotation as per the suggestions in #1112

alexkeizer:
If we have the semantics, then what was the issue that needed a `sorry` in the denotation? Was there a type mismatch?

luisacicolini:
@alexkeizer the signature is: 
`.source t => () → (Ty.stream t)`
the denotation I tried were: 
* `.source t => fun t => HandshakeOp.source t` -> yielded `type mismatch HandshakeOp.source t has type Stream ℕ : Type but is expected to have type Option ⟦t✝⟧ : Type`
* `.source t => HandshakeOp.source t` -> yielded `type mismatch HandshakeOp.source t argument t has type Ty2 : Type but is expected to have type ⟦t⟧ : Type`

and a bunch of other combinations that yielded pretty much the same thing.
I can't tell what's going on'how to fix these from here.

luisacicolini:
I need to merge this to continue working on mixing `comb` with `DC` and `Handshake`. Opened an issue (#1147) to fix `source` over the next weeks, without blocking this PR.

PR #1140, opened 2025-04-21
wip: prototype zext/sext extensions for automata/k-induction

bollu:
This PR prototypes the reflection we would need to do for multiple widths, and also proves correct the on-paper reduction of zext and sext for finite automata.


bollu:
Superceded by #1444

PR #1142, opened 2025-04-21
WIP: feat: typeclass driven refinement notation, with initial dialect-generic refinement support in the framework

alexkeizer:
This PR adds `HRefinement` and `Refinement` typeclasses to drive the `⊑` relation notation, in the same way core Lean uses `HAdd`/`Add` for `+`.

Then, we build on this with a `DialectRefinement` typeclass that we expect dialect-implementors to implement. From this second class, we then derive a few `HRefinement` typeclasses for the various things we might want to compare (e.g., pure values, monadic values, or even `Com` and `Expr`s).

Finally, we remove the existing `⊑` notation that was hardcoded for LLVM, and replace it with instances of this more generic machinery.

------

Still WIP, in particular the last item (migrating the existing notation) is not completed yet


SSA/Core/Framework/Refinement.lean
```diff
@@ -0,0 +1,278 @@
...
+`y` have compatible types. In fact, it would be somewhat simpler if we just bundle
+in the compatibility assumption. If we wanted to state some generic property in
+the current phrasing, we'd have to write:
+```
+∀ {t u} (h : IsTypeCompatible t u) (x : ⟦t⟧) (y : ⟦u⟧), x ⊑ y
+```
+In the bundled version, we could shorten that to simply:
+```
+∀ {t u} (x : ⟦t⟧) (y : ⟦u⟧), x ⊑ y
+```
```
bollu, line 127: Yes, I agree with this rationale. Furthermore, it upgrades smoothly to a refinement of monads, where `m ⊑ n` iff for all `x  ⊑ y`, we have that `m x ⊑ n y`.

SSA/Core/Util/Poison.lean
```diff
@@ -0,0 +1,132 @@
...
+theorem isRefinedBy_poison : a? ⊑ (@poison β) ↔ a? = poison := by
+  cases a?
+  · simp
+  · simp only [not_value_isRefinedBy_poison, false_iff]; rintro ⟨⟩
+
+section PreOrder
+variable {α : Type} [HRefinement α α]
+
+/--
+Refinement on poison values is reflexive if refinement of the underlying values is reflexive.
```
bollu, line 159:
```suggestion
Refinement on poison values is reflexive if refinement of the underlying values is reflexive.
```

I believe this is iff?

alexkeizer: I believe so, too, but we only prove the one direction in this lemma, as that's the direction we generally need

alexkeizer:
superseded by #1165

PR #1154, opened 2025-04-28
feat: mixed `DC` and `Comb` semantics

luisacicolini:
This PR adds a mixed `DC` and `Comb` dialect that allows the parsing of MLIR regions with both the dialects. In particular, I included two functions that pop/push values from/to streams. While this is clearly a hacky solution not consistent with the semantics, it at least got the example runnning. In subsequent PRs I will work on the functor to better manage mixed regions, such that we don't explicitly pop/push values from/to streams.

I also removed the `DCExample` file , as it is in the `DC` directory already.


SSA/Projects/CIRCT/DCxComb/DCxComb.lean
```diff
@@ -0,0 +1,588 @@
...
+  | .shrs _ => BitVec.sshiftRight'
+  | .shru _ => fun x y => CIRCTStream.DCxCombOp.shru x y
+  | .sub _ => BitVec.sub
+  | .xor _ _ => fun xs => CIRCTStream.DCxCombOp.xor (HVector.replicateToList (f := TyDenote.toType) xs)
+
+
+end Dialect
+
+
+def mkTy : MLIR.AST.MLIRType φ → MLIR.AST.ExceptM DCxComb DCxComb.Ty
```
alexkeizer, line 176:
```suggestion
def mkTy : MLIR.AST.MLIRType 0 → MLIR.AST.ExceptM DCxComb DCxComb.Ty
```

The `φ` gives a static bound on the number of mvars. By setting it zero you don't have to check the mvar branch below!

luisacicolini:
I am closing this PR since I am refactoring the comb dialect (#1237) and the semantics of the mixed dialect are incorrect.

PR #1157, opened 2025-04-30
feat: HVector.foldld dependent fold over hvector

alexkeizer:
This PR adds `HVector.foldld`, which is a fold from the left, where the type of the accumulator may depend on the type of the elements already processed. 

That is, it simultaneously folds over the list of types `as` and the elements in an `HVector _ as`.


alexkeizer:
@luisacicolini this is the funny fold you need for `concat`. I've enabled auto-merge so you can pull it in once CI is happy.

PR #1158, opened 2025-04-30
refactor: make LLVM dialect into a (semireducible) def instead of a (reducible) abbrev

alexkeizer:
This PR changes the `LLVM` dialect to be a def rather than an abbrev. As per https://github.com/leanprover/lean4/issues/7984, having reducible definitions in the type of instances causes problems:
> As with simp, there are some unavoidable indexing issues when using reducible patterns. Allowing reducible instances was never intended and should probably be denied.

Thus, we should move away from defining dialects as reducible abbrevs. As a result, instances on `InstCombine.Ty` will no longer be found when synthesizing an instance for `LLVM.Ty`. Thus, we refactor such instances to reinforce that `InstCombine.Ty` is an internal name, and `LLVM.Ty` is the preferred spelling (idem for `Op`).

For further motivation, consider the following MWE:
```lean
abbrev MweDialect : Dialect where
  Op := Empty
  Ty := Unit

def nat : MweDialect.Ty := ()

instance : Monad MweDialect.m := inferInstanceAs (Monad Id)
instance : TyDenote MweDialect.Ty where
  toType
  | () => Nat

instance : DialectHRefinement MweDialect MweDialect where
  IsTypeCompatible := Eq
  IsRefinedBy := @fun
    | (), (), _, x, y => x = y

instance : Fact (IsTypeCompatible nat nat) where
  out := rfl

-- The `#check` works, the instance exists
#check (instRefinementPure : HRefinement ⟦nat⟧ ⟦nat⟧)
-- Yet, it is unable to be synthesized
#synth HRefinement ⟦nat⟧ ⟦nat⟧
```

When we turn `MweDialect` from an `abbrev` into a `def`, the `#synth` line
starts working again!

### DIFF
--- a/SSA/MLIRNatural.lean
+++ b/SSA/MLIRNatural.lean
@@ -6,6 +6,7 @@ import SSA.Projects.InstCombine.LLVM.CLITests
 import Cli
 
 open Lean
+open InstCombine.LLVM.Ty (bitvec)
 
 -- Note that this application of 4 here doesn't really do anything, parameters not supported for now
 @[reducible]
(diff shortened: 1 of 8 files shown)


alexkeizer:
@bollu this is a pretty straightforward refactor, but since it touches quite a bit it'd be nice to get your OK before I hit merge!

PR #1160, opened 2025-04-30
fix: import Tactic.SimpSet in Framework.Macro

alexkeizer:
We use the `simp_denote` simpset in the elaborator-generated code, so it's important we ensure it's (transitively) imported


tobiasgrosser (approved):
This looks great to me.

PR #1164, opened 2025-05-02
refactor: make `EffectKind.toMonad` semireducible

alexkeizer:
This PR removes the `@[reducible]` annotation for EffectKind.toMonad. This definition often shows up in typeclass instances, so it shouldn't be reducible, to avoid the problems reported in https://github.com/leanprover/lean4/issues/7984


alexkeizer:
I'm hitting merge myself, as I believe this to be uncontroversial, given the comments on the referenced issue!

PR #1165, opened 2025-05-02
feat: typeclass driven refinement notation, with initial dialect-generic refinement support in the framework

alexkeizer:
This PR adds HRefinement and Refinement typeclasses to drive the ⊑ relation notation, in the same way core Lean uses HAdd/Add classes for +. Then, we build on this with a DialectRefinement typeclass that we expect dialect-implementors to implement. From this second class, we then derive a few HRefinement typeclasses for the various things we might want to compare (e.g., pure values, monadic values, or even Com and Exprs).

Furthermore, we replace the hardcoded ⊑ notation in the InstCombine project with an instance of this generic notation.

### DIFF
--- a/SSA/Core/EffectKind.lean
+++ b/SSA/Core/EffectKind.lean
@@ -234,3 +234,16 @@ def liftEffect_compose {e1 e2 e3 : EffectKind} {α : Type} [Pure m]
     (h13 : e1 ≤ e3 := le_trans h12 h23) :
     ((liftEffect (α := α) h23) ∘ (liftEffect h12)) = liftEffect (m := m) h13 := by
   cases e1 <;> cases e2 <;> cases e3 <;> (solve | rfl | contradiction)
+
+/-!
+## `toMonad` coercion
+-/
+
+/--
+Coerce a value of type `eff.toMonad m α` into a monadic value `m α`, by applying
+either `pure` or the identity, depending on the effect `eff`.
+
+NOTE: This is simply `liftEffect` with the second effect fixed to be impure.
+-/
+abbrev coe_toMonad [Pure m] {eff : EffectKind} : eff.toMonad m α → m α :=
+  liftEffect (le_impure eff)
--- a/SSA/Core/EffectKind.lean
+++ b/SSA/Core/EffectKind.lean
@@ -245,5 +245,6 @@ either `pure` or the identity, depending on the effect `eff`.
 
 NOTE: This is simply `liftEffect` with the second effect fixed to be impure.
 -/
-abbrev coe_toMonad [Pure m] {eff : EffectKind} : eff.toMonad m α → m α :=
+@[simp]
+def coe_toMonad [Pure m] {eff : EffectKind} : eff.toMonad m α → m α :=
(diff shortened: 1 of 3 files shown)


SSA/Core/EffectKind.lean
```diff
@@ -234,3 +234,16 @@ def liftEffect_compose {e1 e2 e3 : EffectKind} {α : Type} [Pure m]
...
+-/
+
+/--
+Coerce a value of type `eff.toMonad m α` into a monadic value `m α`, by applying
+either `pure` or the identity, depending on the effect `eff`.
+
+NOTE: This is simply `liftEffect` with the second effect fixed to be impure.
+-/
+abbrev coe_toMonad [Pure m] {eff : EffectKind} : eff.toMonad m α → m α :=
+  liftEffect (le_impure eff)
```
bollu: I am now generally a bit way of `abbrev`s: Why must this be an `abbrev`?

SSA/Core/Framework/Refinement.lean
```diff
@@ -0,0 +1,159 @@
...
+end DialectHRefinement
+
+/--
+A lawful homogenous (i.e., within a single dialect) refinement instance is one
+where refinement is reflexive and transitive (i.e., it is a preorder).
+-/
+class LawfulDialectRefinement (d : Dialect) [TyDenote d.Ty] [DialectHRefinement d d] where
+  isRefinedBy_rfl : ∀ {t : d.Ty} (x : d.m ⟦t⟧), x ⊑ x
+  isRefinedBy_trans : ∀ {t u v : d.Ty} (x : d.m ⟦t⟧) (y : d.m ⟦u⟧) (z : d.m ⟦v⟧),
+    x ⊑ y → y ⊑ z → x ⊑ z
```
bollu: If an honestly "lawful" dialect refinement better be monotone, then perhaps we rename this to `PreorderDialectRefinement`, and the other notion as `MonotoneDialectRefinement` (defined in the appropriate fle), and then define `Lawful = Preorder + Monotone`?


```suggestion
class PreorderDialectRefinement (d : Dialect) [TyDenote d.Ty] [DialectHRefinement d d] where
  isRefinedBy_rfl : ∀ {t : d.Ty} (x : d.m ⟦t⟧), x ⊑ x
  isRefinedBy_trans : ∀ {t u v : d.Ty} (x : d.m ⟦t⟧) (y : d.m ⟦u⟧) (z : d.m ⟦v⟧),
    x ⊑ y → y ⊑ z → x ⊑ z
```

bollu (approved):
LGTM, with the comment about renaming `Lawful` to `Preorder`, so that we can define monotonicity in the next PR as `Monotone`, and use this to define `Lawful = Preorder + Monotone`.

SSA/Core/EffectKind.lean
alexkeizer: That's a very sensible wariness! Here I didn't want to duplicate all the simp lemmas we already have for `liftEffect`, but making it a `@[simp] def` has the same effect, with less pitfalls (I hope?), so I'll go with that

PR #1178, opened 2025-05-05
refactor: replaces uses of PoisonOr.Refinement syntax with generic HRefinement syntax

### LINKED ISSUE #1176: Syntax in aliveStatements changed
When running `cd SSA/Projects/InstCombine; ./update_alive_statements.py` our syntax has changed.

```diff
 theorem bv_AddSub_1156 :
-    ∀ (e : LLVM.IntW w), LLVM.add e e ⊑ LLVM.shl e (LLVM.const? w 1) := by
+    ∀ (e : LLVM.IntW w), PoisonOr.IsRefinedBy (LLVM.add e e) (LLVM.shl e (LLVM.const? w 1)) := by
   simp_alive_undef
   simp_alive_ops
   simp_alive_case_bash
@@ -39,7 +41,7 @@ theorem bv_AddSub_1156 :
   all_goals sorry
```

alexkeizer:
This PR migrates any uses of the PoisonOr-specific notation to the generic HRefinement notation, allowing us to remove the former and promote the latter to a global notation (rather than a scoped notation).

To facilitate this move, the PR also adds a simproc to the `simp_denote` simpset that normalizes the refinement notation instances derived from `DialectHRefinement` into the base `DialectHRefinement` instance.


tobiasgrosser:
Nice. This improves the situation in https://github.com/opencompl/lean-mlir/issues/1176. It would be great if you can get this reviewed with @bollu and then merged. I can then update `AliveStatements.lean` and start re-enabling our larger llm tests.

SSA/Core/EffectKind.lean
```diff
@@ -15,8 +17,8 @@ def toMonad (e : EffectKind) (m : Type → Type) : Type → Type :=
 
 section Lemmas
 
-@[simp] theorem toMonad_pure    : pure.toMonad m = Id := rfl
-@[simp] theorem toMonad_impure  : impure.toMonad m = m := rfl
+@[simp, simp_denote] theorem toMonad_pure   : pure.toMonad m = Id := rfl
+@[simp, simp_denote] theorem toMonad_impure : impure.toMonad m = m := rfl
```
bollu, line 21: Can we think whether we need both `simp` as well as `simp_denote`?

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -21,6 +21,12 @@ namespace IntW
 
 instance : Inhabited (IntW w) := by unfold IntW; infer_instance
 
+instance : Refinement (LLVM.IntW w) := inferInstanceAs (Refinement <| PoisonOr _)
+@[simp, simp_llvm]
+theorem isRefinedBy_iff (x y : LLVM.IntW w) :
+    x ⊑ y ↔ @HRefinement.IsRefinedBy (PoisonOr _) (PoisonOr _) _ x y := by
+  rfl
```
bollu, line 37: It's not clear to me why this is the "good" direction of the simplification? In general, the typeclass tower is unclear to me!

bollu (changes requested):
I think we're close! I request having tests for the simproc as we do for `BitVec` simprocs, as I really find myself nervous with complex simprocs we're writing.

I also find the rewrite directions confusing, so maybe a chat on/off-line about this would help :)

SSA/Core/EffectKind.lean
alexkeizer: So this definitely has to be `simp_denote`, for the automation to work.

Besides that, I feel it makes sense as a global `simp` lemma. When writing a proof by hand, I'd expect `pure.toMonad m` to simplify, so, to me, it'd be surprising if `simp` (without extra prompting) didn't do so. Especially since not all `simp_denote` lemmas necessarily make sense in a manual proof context, so I wouldn't want to have to rely on `simp [simp_denote]`.

SSA/Projects/InstCombine/LLVM/Semantics.lean
alexkeizer: I added a docstring that hopefully clarifies:
```
/--
`isRefinedBy_iff` rewrites refinement of `LLVM.IntW` values into refinement
of `PoisonOr _` values, effectively unfolding the `LLVM.IntW` definition (and
the Refinement instance).

By making this a simp-lemma, we ensure all `PoisonOr` simp-lemmas apply without
having to duplicate that API for `LLVM.IntW`.
-/
```

I'd agree that normally this rewrite might not want to a `simp` rewrite, as it effectively means that the simp-normal form for `LLVM.IntW` becomes `PoisonOr`. However, given that we were originally OK with the former being an abbrev, in this case the rewrite seems fine to me. Certainly, duplicating all PoisonOr simp-lemmas does not seem worth the maintenance hassle to me.

alexkeizer:
> I also find the rewrite directions confusing, so maybe a chat on/off-line about this would help :)

Yeah, there is a lot of instances going around. I tried to clarify this a bit by writing the following documentation:
````
/-!
## Normal Forms

In a concrete dialect, we expect to have a kind of waterfall of refinement
instances. Take, for example, the LLVM dialect:
* Firstly, there is the `PoisonOr _` type, which has a refinement relation.
* Then, there is the `LLVM.IntW w` type, which is a (semireducible!) alias for
  `PoisonOr (BitVec w)`. Because it is not reducible, it has its own refinement
  instance.
* Using the latter, there is the `DialectHRefinement` instance for the LLVM
  dialect.
* This `DialectHRefinement` instance implies refinement instances for both
  monadic value `LLVM.m ⟦t⟧` and pure values `⟦t⟧`.

Recall, however, that LLVM is a *pure* dialect (that is, `LLVM.m` is `Id`).
Suppose we had a `MemLLVM` dialect, which also captures memory side effects,
then we'd expect the following refinement instances:
* We still start with `PoisonOr _` and `LLVM.IntW _`, and their refinement.
* Then, we expect some `EffectM` monad, which will have a refinement instance
  for `EffectM α`, with `α` being some generic type with a refinement instance.
* Using both of the above, we then define the `DialectHRefinement` instance,
  which implies the same refinement instances as before.

Note that in either case, the semantics content is defined at the top of the
list, with the instances after `DialectHRefinement` simply being ways to express
this refinement for alternative spellings of the same type.

Thus, although a rewrite like `toMonad_pure_IsRefinedBy_toMonad_pure_iff`
*in general* seems like a good idea (the rhs is a simpler expression, after all),
for a concrete dialect this rewrite is actually hiding the semantic content by
going *down* this list of instances.

Luckily, we have a dedicated simpset intended for concrete rewrites, `simp_denote`.
Thus, we add set of lemmas to this simpset which rewrite the derived instances
in terms of the original `DialectHRefinement` instance.

We expect dialect implementors to then add their dialect-specific simplemmas to
`simp_denote`, to go all the way up the list of instances.
-/
````

Basically, there is one instance that actually has semantic content (the PoisonOr instance) with a tower of instances on top that simply re-export that instance verbatim for other (def-eq) spellings of that same type. 

Since these other instances don't actually add anything, to me, it makes sense to just normalize all these instances up the chain, even if normally you wouldn't want to do that (since normally that might explode your expression with all the slight alterations that happen at each level).

Is there anything concrete there that you feel needs more clarification? Let's have a chat when you're available.

alexkeizer:
Since @bollu is away and I think his concerns have been addressed I'm hitting merge (feel free to post-merge call me out on it if there are any other issues!)

tobiasgrosser:
Thank you!

PR #1181, opened 2025-05-06
feature: RiscV dialect base and semantics,

salinhkuhn:
This PR contains a Risc-V dialect for a 64-bit address space. We support the base ISA, M and B extensions as well as part of the different Z extensions. The semantics faithfully implement the semantics specified in Sail. We have partially removed the toInt's and toNat's in the semantics definition, part of this transformation is still wip and marked as such.


bollu:
Jesus, that's a chonky PR! 

I will hit "approve", but I would like us to think about the following questions:

1) How do we know that this faithfully implements Sail? Is it possible, in a future PR, to directly reuse Sail's definitions? (adding sail as a dependency, of course). 
2) Is it possible to auto-generate this ourselves? I guess this is the same as the above question: Can we somehow directly hook into sail, instead of having to trust this implementation?

tobiasgrosser:
I will do a review in a bit.

salinhkuhn:
Regarding the sail part, I have proofs for every translation step in Lean, hence an end-to-end proof that the provided semantics are correct even when removing toInt and toNats. In my own project I had sail as a dependency. I didn't included the file containing the proof because I assumed that would be too much. Autogenerating the definitions from Sail would be nice but the original sail def contain toInts and toNats which is impractical if we want to use bv_decide to prove the rewrites correct.

+ I didn't include the pretty printer for the moment, there is one but I will wait until this is approved.

bollu:
Right. I feel it would be worth it to add Sail as a dependency, and the proofs of equivalence as a follow up PR @salinhkuhn. Let's see what @alexkeizer  and @tobiasgrosser  think.

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,336 @@
...
+def ZBS_RTYPE_pure64_RISCV_BEXT (rs2_val : BitVec 64) (rs1_val : BitVec 64)  :=
+  BitVec.setWidth 64
+    (match
+      BitVec.and rs1_val ( (BitVec.setWidth 64 1#1) <<< (BitVec.extractLsb 5 0 rs2_val)) !=
+        0#64 with
+    | true => 1#1
+    | false => 0#1)
+
+ def ZBS_RTYPE_pure64_BINV (rs2_val : BitVec 64) (rs1_val : BitVec 64) :=
+  BitVec.xor rs1_val ((BitVec.zeroExtend 64 1#1) <<< (BitVec.extractLsb 5 0 rs2_val))
```
tobiasgrosser:
```suggestion
def ZBS_RTYPE_pure64_BINV (rs2_val : BitVec 64) (rs1_val : BitVec 64) :=
    BitVec.xor rs1_val ((BitVec.zeroExtend 64 1#1) <<< (BitVec.extractLsb 5 0 rs2_val))
```

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,336 @@
...
+  The semantic defintions do not contain toInt and toNat when possible. For this
+  we rewrote the Sail semantics into pure bit vector only operations including a proof
+  of correctness.
+  This rewrite should allow to gain in automation using the bv_decide.
+-/
+
+namespace RV64Semantics
+
+def ADDIW_pure64 (imm : BitVec 12) (rs1_val : BitVec 64) :  BitVec 64 :=
+     BitVec.signExtend 64 (BitVec.setWidth 32 (BitVec.add (BitVec.signExtend 64 imm) rs1_val))
```
tobiasgrosser:
```suggestion
def ADDIW_pure64 (imm : BitVec 12) (rs1_val : BitVec 64) :  BitVec 64 :=
  BitVec.signExtend 64 (BitVec.setWidth 32 (BitVec.add (BitVec.signExtend 64 imm) rs1_val))
```

The indentation throughout this file is inconsistent. Lean indents definitions by two whitespaces. Please make sure you follow this through this file.

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,336 @@
...
+
+-- loads immediate into upper 20 bits and then fills the rest up with 0 and returns it as the result
+def UTYPE_pure64_lui (imm : BitVec 20) (pc : BitVec 64) : BitVec 64 :=
+     BitVec.signExtend 64 (imm ++ (0x000 : (BitVec 12)))
```
tobiasgrosser:
```suggestion
def UTYPE_pure64_lui (imm : BitVec 20) : BitVec 64 :=
     BitVec.signExtend 64 (imm ++ (0x000 : (BitVec 12)))
```

The 2nd argument is unused. Should we drop it? If it must be there, document in the comment why it is there and name it `_` to avoid the warning.

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,336 @@
...
+
+def UTYPE_pure64_AUIPC (imm : BitVec 20) (pc : BitVec 64)  : BitVec 64 :=
+    BitVec.add (BitVec.signExtend 64 (BitVec.append imm (0x000 : (BitVec 12)))) pc
+
+def SHIFTIWOP_pure64_RISCV_SLLIW (shamt : BitVec 5) (rs1_val : BitVec 64) : BitVec 64 :=
+    BitVec.signExtend 64 ( (BitVec.extractLsb' 0 32 rs1_val) <<< (shamt))
+
```
tobiasgrosser:
```suggestion
    BitVec.signExtend 64 ( (BitVec.extractLsb' 0 32 rs1_val) <<< shamt)

```

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,336 @@
...
+-- wip: still needs to be rewritten into bitvec only operations
+def SHIFTIWOP_pure64_RISCV_SRLIW (shamt : BitVec 5) (rs1_val : BitVec 64) : BitVec 64 :=
+    BitVec.signExtend 64 ( (BitVec.extractLsb' 0 32 rs1_val) >>> (shamt))
+
```
tobiasgrosser:
```suggestion
    BitVec.signExtend 64 ( (BitVec.extractLsb' 0 32 rs1_val) >>> shamt)
```

Check for unnecessary parenthesis throughout the file.

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,336 @@
...
+def UTYPE_pure64_lui (imm : BitVec 20) (pc : BitVec 64) : BitVec 64 :=
+     BitVec.signExtend 64 (imm ++ (0x000 : (BitVec 12)))
+
+def UTYPE_pure64_AUIPC (imm : BitVec 20) (pc : BitVec 64)  : BitVec 64 :=
+    BitVec.add (BitVec.signExtend 64 (BitVec.append imm (0x000 : (BitVec 12)))) pc
+
+def SHIFTIWOP_pure64_RISCV_SLLIW (shamt : BitVec 5) (rs1_val : BitVec 64) : BitVec 64 :=
+    BitVec.signExtend 64 ( (BitVec.extractLsb' 0 32 rs1_val) <<< (shamt))
+
+-- wip: still needs to be rewritten into bitvec only operations
```
tobiasgrosser: Is this comment still up-to-date? To me it seems as if this is a BitVec only function, no?

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,336 @@
...
+      BitVec.signExtend 64 (BitVec.sub rs1_val32 rs2_val32)
+
+def RTYPEW_pure64_RISCV_SLLW (rs2_val : BitVec 64) (rs1_val : BitVec 64) : BitVec 64 :=
+    let rs1_val32 := BitVec.extractLsb' 0 32 rs1_val;
+    let rs2_val32 := BitVec.extractLsb' 0 32 rs2_val;
+    let shamt := BitVec.extractLsb' 0 5 rs2_val32;
+    BitVec.signExtend 64 (rs1_val32 <<< shamt)
+
+def RTYPEW_pure64_RISCV_SRLW (rs2_val : BitVec 64) (rs1_val : BitVec 64)  : BitVec 64 :=
+    let rs1_val32 := BitVec.extractLsb' 0 32 rs1_val;
```
tobiasgrosser:
```suggestion
def RTYPEW_pure64_RISCV_SRLW (rs2_val : BitVec 64) (rs1_val : BitVec 64) : BitVec 64 :=
    let rs1_val32 := BitVec.extractLsb' 0 32 rs1_val;
```

There should be a single whitespace, instead of a sequence of two whitespaces.

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,336 @@
...
+    let rs2_val32 := BitVec.extractLsb' 0 32 rs2_val;
+    let shamt := BitVec.extractLsb' 0 5 rs2_val32;
+    BitVec.signExtend 64 (rs1_val32 >>> shamt)
+
+def RTYPEW_pure64_RISCV_SRAW (rs2_val : (BitVec 64)) (rs1_val : (BitVec 64)) :=
+  BitVec.signExtend 64
```
tobiasgrosser:
```suggestion
def RTYPEW_pure64_RISCV_SRAW (rs2_val : BitVec 64) (rs1_val : BitVec 64) :=
  BitVec.signExtend 64
```

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,336 @@
...
+      (BitVec.sshiftRight' (BitVec.extractLsb 31 0 rs1_val) (BitVec.extractLsb 4 0 rs2_val))
+
+def RTYPE_pure64_RISCV_ADD (rs2_val : BitVec 64) (rs1_val : BitVec 64) :BitVec 64 :=
+      BitVec.add rs1_val rs2_val
```
tobiasgrosser:
```suggestion
def RTYPE_pure64_RISCV_ADD (rs2_val : BitVec 64) (rs1_val : BitVec 64) : BitVec 64 :=
      BitVec.add rs1_val rs2_val
```

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,336 @@
...
+
+def RTYPE_pure64_RISCV_SLTU (rs2_val : BitVec 64) (rs1_val : BitVec 64) : BitVec 64 :=
+    let b := BitVec.ult rs1_val rs2_val;
+      BitVec.setWidth 64 (BitVec.ofBool b)
+
+def RTYPE_pure64_RISCV_AND (rs2_val : BitVec 64) (rs1_val : BitVec 64) : BitVec 64 :=
+      BitVec.and rs2_val rs1_val
+
+def RTYPE_pure64_RISCV_OR(rs2_val : BitVec 64) (rs1_val : BitVec 64) : BitVec 64 :=
+      BitVec.or rs2_val rs1_val
```
tobiasgrosser:
```suggestion
def RTYPE_pure64_RISCV_OR(rs2_val : BitVec 64) (rs1_val : BitVec 64) : BitVec 64 :=
      BitVec.or rs2_val rs1_val
```
```suggestion
def RTYPE_pure64_RISCV_OR (rs2_val : BitVec 64) (rs1_val : BitVec 64) : BitVec 64 :=
      BitVec.or rs2_val rs1_val
```

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,390 @@
...
+
+/--
+## Dialect operation definitions
+
+Specifing the signature of each `RISCV64` operation. `Sig` refers to the input types
+for each operation as a list of types.
+
+We mark it as `simp` and `reducible` such that the type checker and elaborator
+always fully unfold a `sig` to its underlying defintion and when the simplifier
+encounters a `sig` it can replace it by its definition.
```
tobiasgrosser:
```suggestion
always fully unfold a `sig` to its underlying definition and when the simplifier
encounters a `sig` it can replace it by its definition.
```

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,390 @@
...
+  Ty := Ty
+
+instance : DialectSignature RV64 := ⟨Op.signature⟩
+
+/--
+## Dialect semantics
+
+We assign semantics defined in `RV64` to each operation. This defines the meaning of each operation in Lean.
+
+When writting RISC-V-like abstract MLIR ISA intutivly we write `op  arg1 arg2`. The `RISCV64` are defined to
```
tobiasgrosser:
```suggestion
When writting RISC-V-like abstract MLIR ISA intuitively we write `op  arg1 arg2`. The `RISCV64` are defined to
```

tobiasgrosser (approved):
Hi @salinhkuhn, this PR looks great already. What a comprehensive model of the RISC-V ISA!

To get it in, I first did a first stylistic pass. Can you address all the comments and double-check for similar formatting changes throughout your PR? I will then do a semantic pass and feel this can be merged latest tomorrow.

SSA/Projects/RISCV64/Semantics.lean
salinhkuhn: no, you are right. I've rewritten it, comment is useless

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,384 @@
+
+import SSA.Projects.RISCV64.Semantics
```
tobiasgrosser:
```suggestion
import SSA.Projects.RISCV64.Semantics
```

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,384 @@
...
+
+open RV64Semantics
+
+namespace RISCV64
+/-! ## The `RISCV64` dialect -/
+
+/-!
+## Dialect operation definitions
+Modelled operations such that their inputs arguments are up to two register content values. Any other attributes e.g flags
+are encoded as part of the op-code.
```
tobiasgrosser:
```suggestion
Modelled operations such that their input arguments are up to two register content values. Any other attributes , e.g., flags
are encoded as part of the op-code.
```

The first part of the sentence is not clear to me. Also, break the line after 100 characters.

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,384 @@
...
+
+We modell the RV64I base instruction set: https://github.com/riscv/riscv-isa-manual/blob/main/src/rv64.adoc, therefore allow for an 64-bit address space.
+And operations out of selected RISC-V extension:
+    - `M`: standart integer division and multiplication instruction extension
```
tobiasgrosser:
```suggestion
And operations out of selected RISC-V extension:
```

This is not a full sentence

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,384 @@
...
+  | slti (imm : BitVec 12)
+  | sltiu (imm : BitVec 12)
+  | srli (shamt : BitVec 6)
+  | srai (shamt : BitVec 6)
+  | slliw (shamt : BitVec 5)
+  | srliw (shamt : BitVec 5)
+  | sraiw (shamt : BitVec 5)
+  | slt
+  | sltu
+  -- RISC-V `M` extension instructions (multiply& divide)
```
tobiasgrosser:
```suggestion
  -- RISC-V `M` extension instructions (multiply & divide)
```

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,332 @@
...
+  BitVec.add (rs1_val <<< 3#2) rs2_val
+
+/-
+to complete in the future:
+extend with
+ZBB_RYTPE_pure64_RISCV_
+|pack
+|packh
+slli.uw
+ -/
```
tobiasgrosser: Not sure we need this comment?

tobiasgrosser (approved):
LGTM.

(also, please resolve the comments that you addressed. If you can post a brief statement how you resolved each comment, this would help to track that they got resolved)

SSA/Projects/RISCV64/Semantics.lean
salinhkuhn: removed it and listed on my own future to dos

SSA/Projects/RISCV64/Base.lean
salinhkuhn: rewrote it

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,384 @@
...
+  Ty := Ty
+
+instance : DialectSignature RV64 := ⟨Op.signature⟩
+
+/--
+## Dialect semantics
+We assign semantics defined in `RV64` to each operation. This defines the meaning of each operation in Lean.
+When writting RISC-V-like abstract MLIR ISA intuitively we write `op  arg1 arg2`. The `RISCV64` semantics are defined to
+first process the second operand then the first. Therefore we first pass `.get 1` then `.get 0` into the
+functions that define our semantics.
```
tobiasgrosser, line 313: Can you break these lines below 100 characters?

tobiasgrosser:
Feel invited to merge after my last comment has been addressed.

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,388 @@
+import SSA.Projects.RISCV64.Semantics
+import SSA.Core.Framework
+set_option maxHeartbeats 1000000000000000000
```
alexkeizer, line 3: Why do we need this option? If it's for one specific definition, it'd be nicer to scope the option to just that definition by using `set_option .... in`, for example:
```
set_option maxHeartbeats 100000000000 in
def FooThatNeedsABunchOfHeartbeats := ...
```

In either case, it'd also be nice to have a comment to explain why (to the best of your current understanding; no need to spend hours figuring this out) we need that many heartbeats.

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,388 @@
...
+
+namespace RISCV64
+/-! ## The `RISCV64` dialect -/
+
+/-!
+## Dialect operation definitions
+Each instruction operates on up to two registers. Any other attributes not
+passed via registers, e.g., flags and immediates are encoded as part of the op-code.
+
+We modell the RV64I base instruction set: https://github.com/riscv/riscv-isa-manual/blob/main/src/rv64.adoc,
```
alexkeizer, line 15: modell -> model

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,388 @@
...
+therefore allow for an 64-bit address space.
+Additionally, we modell operations out of selected RISC-V ISA extension:
+    - `M`: standart integer division and multiplication instruction extension
```
alexkeizer, line 18: standart -> standard

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,388 @@
...
+  | xor
+  | sll
+  | srl
+  | sra
+  | addw
+  | subw
+  | sllw
+  | srlw
+  | sraw
+  -- fence
```
alexkeizer, line 48: I guess this comment means we *don't* model fence? Maybe write `-- fence missing ` like you did for `slli.uw`

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,388 @@
...
+inductive Ty
+  | bv : Ty
+  deriving DecidableEq, Repr, Inhabited
+
+instance : ToString (Ty) where
+  toString t := repr t |>.pretty
+
+/-!
+Connecting the `bv` type to its underlying Lean type `BitVec 64`. By providing a `TyDenote` instance,
+we define how the `RISCV64` types transalte into actual Lean types.
```
alexkeizer, line 122: transalte -> translate

SSA/Projects/RISCV64/Base.lean
alexkeizer, line 15: So the type of comments we use in Lean is a little bit subtle: as you've noticed the most common one for documentation are `/-! -/` and `/-- -/`.

The former, `/-! ... -/` is used to tell the over-arching story of a code file, or to add structure to that file with headers, like we do when we write `/-! ## Dialect Operations -/`.

The latter, `/-- ... -/`, is also called a docstring, and is used to attach documentation to a specific definition or theorem. These comments are special because they show up when we hover over a usage of that definition/theorem in a different file, so we often write these docstring in a kind of independent style, where the text should make sense in isolation from the rest of the file (since it will appear by itself in these hovers).

This is all to say, that I'd suggest you put the header in a `/-! ... -/` doc comment by itself, and then describe the `Op` type in a `/-- ... -/` docstring. Furthermore, It's very nice to include the links to the documentation as you have! To make it even better, you could turn them into a kind of footnote reference. That way you still have the link, but the text flows a bit nicer.

For example, you might write:
```
/-! ## Dialect Operations -/

/-- 
`Op` models the RV64I base instruction set [1] plus selected RISC-V ISA extensions: `M` for standard integer division and multiplication [2], `B` for bit manipulation instructions (comprises instructions provided by the `Zba`, `Zbb`, and `Zbs` extensions) [3], and `Zicond` for conditional operations [4].

We model RV64 as an SSA IR, meaning that instructions won't specify any registers (instead, they'll later receive SSA values). However, any other attributes (e.g., flags or immediate values) are still encoded as part of the operation.

[1] https://link/to/documation
[2] ...
[3] ...
[4] ...
```

You'll notice I didn't mention the comment about the 64bit address space. It's good to explicitly call that out, as you have, but I feel that comment might be better on the docstring for `Ty`, since it relates more the fact we only encode 64-bit vectors, rather than any specific operation.

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,388 @@
...
+  /- RISC-V `Zicond` conditional operations extension  -/
+  | czero.eqz
+  | czero.nez
+  deriving DecidableEq, Repr
+
+/--
+## Dialect type definitions
+Defining a type system for the `RISCV64` dialect. `bv` represents bit vector.
+-/
+inductive Ty
```
alexkeizer, line 113: Here you correctly give a docstring to the `Ty` inductive. However, we generally don't start docstrings with a header.

Instead, you could write something like
```lean
/-! ## Dialect Types -/

/--
`Ty` models the types of the RV64 dialect.
Recall that RV64I has a fixed 64-bit address space, so we choose to only have 64-bit wide bitvectors.

```

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,388 @@
...
+  | bv : Ty
```
alexkeizer, line 114: Note you can even add docstrings to the variants of an inductive. That way, if you hover over `Ty` you'll get the general docstring, but if you hover over `Ty.bv` you'll get the docstring for that specific variant.

For example, you could write
```
/-- `bv` represents the type of 64-bit wide bitvectors -/
| bv : Ty
```

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -0,0 +1,323 @@
...
+
+-- loads immediate into upper 20 bits and then fills the rest up with 0 and returns it as the result
+def UTYPE_pure64_lui (imm : BitVec 20) : BitVec 64 :=
+  BitVec.signExtend 64 (imm ++ (0x000 : (BitVec 12)))
+
+def UTYPE_pure64_AUIPC (imm : BitVec 20) (pc : BitVec 64) : BitVec 64 :=
+  BitVec.add (BitVec.signExtend 64 (BitVec.append imm (0x000 : BitVec 12))) pc
+
+def SHIFTIWOP_pure64_RISCV_SLLIW (shamt : BitVec 5) (rs1_val : BitVec 64) : BitVec 64 :=
+  BitVec.signExtend 64 ((BitVec.extractLsb' 0 32 rs1_val) <<< shamt)
```
alexkeizer, line 28: I'd suggest picking a consistent normal form. In this case, `BitVec.extractLsb' 0 w _` is the same as `BitVec.setWidth w _`, so ideally you'd only use one of those two forms in your entire semantics. The latter is shorter and simpler, so I'd suggest only using `setWidth`.

Do note: even though `extractLsb' 0 w` is equivalent to `setWidth w`, I don't believe they are definitionally *equal*. Thus, making this change might break some of your proofs. You'd fix those proofs by rewriting along [`setWidth_eq_extractLsb'`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/BitVec/Lemmas.html#BitVec.setWidth_eq_extractLsb'). Since this change will have some impact elsewhere, I'd suggest you make a dedicated PR where you *only* change `extractLsb'` to `setWidth` (and the changes needed to fix proofs that break). That is, to keep it separate from the other changes I'm suggesting. Keeping the PRs separate will make it much easier for us to review and give feedback. It'll also be easier for you to get the more trivial changes out of the way and merged first!

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,388 @@
...
+  | sraiw (shamt : BitVec 5)
+  | slt
+  | sltu
+  -- RISC-V `M` extension instructions (multiply & divide)
+  | mul    -- performs signed multiplication on 64 x 64 bits and returns the lower 64 bits of the result .
+  | mulu   -- performs unsigned multiplication on 64 x 64 bits and returns the lower 64 bits of the result .
+  | mulw
+  | mulh   -- performs signed multiplication on 64 x 64 bits and returns the upper 64 bits of the result.
+  | mulhu  -- performs unsigned multiplication on 64 x 64 bits and returns the upper 64 bits of the result.
+  | mulhsu -- performs multiplication on (rs1 signed) x (rs2 unsigned) and returns the upper 64 bits of the result.
```
alexkeizer, line 64: Nit: I'd suggest adding a `MulFlags` structure similar to what they have in Sail. Then in `Op` you'd simply have a single `mul` instruction that takes a `MulFlags`, and then in the parser you'd add cases for `mul`, `mulu`, `mulw`, `mulh`, etc. and connect them to the right flags.

N.B: when I write `Nit` in a review comment, that means I'm also OK with what you currently have. Either I think my suggestion is cleaner, but admit there's an element of personal preference which isn't worth debating (so it's fine if you disagree), or just want to make you aware of a different option in the design space. Usually, that means if you disagree with such a comment, it's totally fine to just briefly reply saying why you disagree but further ignore the comment.

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,388 @@
...
+  |.remu  =>  [Ty.bv, Ty.bv]
+  |.addiw (imm : BitVec 12) => [Ty.bv]
+  |.lui (imm : BitVec 20) => [Ty.bv]
+  |.auipc (imm : BitVec 20)  => [Ty.bv]
+  |.slliw (shamt : BitVec 5)  => [Ty.bv]
+  |.srliw (shamt : BitVec 5) => [Ty.bv]
+  |.sraiw (shamt : BitVec 5) => [Ty.bv]
+  |.slli (shamt : BitVec 6) => [Ty.bv]
+  |.srli (shamt : BitVec 6) => [Ty.bv]
+  |.srai (shamt : BitVec 6) => [Ty.bv]
```
alexkeizer, line 161: Note that there is no need for the type ascription here, `| .addiw imm => [Ty.bv]` is also perfectly valid.

Also, when you're not actually using the variable on the rhs, it's convention to use the wildcard `_` pattern, as in
`| .addiw _ => [Ty.bv]`. If you prefer having at least some description of the blank value, we'd at least use a name prefixed with `_`, e.g. `| .addiw _imm => [Ty.bv]`, to emphasize the rhs does not depend on `imm`.

Finally: I believe VSCode should've given you a warning about unused variables on this line. Please don't ignore warnings! If you get a warning but aren't sure what it means, just ask on Zulip! In this case, unused variables are relatively harmless, but in other cases warnings give valuable information about things that might be wrong with your code.

SSA/Projects/RISCV64/Base.lean
```diff
@@ -0,0 +1,388 @@
...
+instance : DialectDenote (RV64) where
+  denote
+  |.li imm, _ , _ => BitVec.ofInt 64 imm
+  |.addiw imm, regs, _ => ADDIW_pure64 imm (regs.getN 0 (by simp [DialectSignature.sig, signature]))
+  |.lui imm, regs , _ => UTYPE_pure64_lui imm
+  |.auipc imm, regs, _ => UTYPE_pure64_AUIPC imm (regs.getN 0 (by simp [DialectSignature.sig, signature]))
+  |.slliw shamt, regs, _ => SHIFTIWOP_pure64_RISCV_SLLIW shamt (regs.getN 0 (by simp [DialectSignature.sig, signature]))
+  |.srliw shamt, regs, _ => SHIFTIWOP_pure64_RISCV_SRLIW shamt (regs.getN 0 (by simp [DialectSignature.sig, signature]))
+  |.sraiw shamt, regs, _ => SHIFTIWOP_pure64_RISCV_SRAIW shamt (regs.getN 0 (by simp [DialectSignature.sig, signature]))
+  |.slli shamt, regs, _ => SHIFTIOP_pure64_RISCV_SLLI shamt (regs.getN 0 (by simp [DialectSignature.sig, signature]))
```
alexkeizer, line 325: Nit: you could use the `def_denote` mechanism here, which should be quite a bit cleaner. I believe the Comb dialect on master uses it nowadays, so you could have a look there for an example.

Note: this might also be a bigger change that could well deserve it's own dedicated follow-up PR. In general, smaller more focussed PRs are better!

alexkeizer (commented):
Thanks for polishing and submitting the dialect Sarah! Sorry I couldn't review this yesterday. The core definitions look very nice, I've mostly got some suggestions about documentation! 
You can create follow-up PRs to do the changes I suggest.

SSA/Projects/RISCV64/Base.lean
salinhkuhn: ✓

SSA/Projects/RISCV64/Base.lean
salinhkuhn: I will keep that in mind and perform the change in the same PR where I will adapt the semantics to the bit vec operations free of toNat and toInt.

SSA/Projects/RISCV64/Base.lean
salinhkuhn: ✓, I changed it to _imm / _shamt, I think it is helpful the indicate the argument. I changed it to be consistently the same naming throughout the file and removed the type annotation.

SSA/Projects/RISCV64/Base.lean
salinhkuhn: I did the change ✓. And directly upstreamed it with this PR. I will try to keep any future PR's smaller but atm I have a lot of code to provide...

SSA/Projects/RISCV64/Base.lean
alexkeizer: > I will keep that in mind and perform the change in the same PR where I will adapt the semantics to the bit vec operations free of toNat and toInt.

Please don't! It's much easier to have PRs that have a single purpose, instead of a bigger PR that combine many different changes together

SSA/Projects/RISCV64/Base.lean
alexkeizer: > And directly upstreamed it with this PR

What do you mean by upstreaming it with this PR? If you pushed the change to the branch you used to open this PR, that doesn't work. Only the changes that were in that branch at the time it was merged are actually merged. If you want to create follow up changes, you need to make a new branch with a new PR

PR #1195, opened 2025-05-09
chore: improve documentation for RiscV dialect

salinhkuhn:


SSA/Projects/RISCV64/Base.lean
```diff
@@ -1,26 +1,28 @@
 import SSA.Projects.RISCV64.Semantics
 import SSA.Core.Framework
 set_option maxHeartbeats 1000000000000000000
```
alexkeizer, line 7:
```suggestion

/-
This file has a number of very large inductive types, which seem to cause Lean to run out of heartbeats.
We avoid the issue by increasing the heartbeats. Since this applies to most inductives in this file, we do so globally.
-/
set_option maxHeartbeats 1000000000000000000
```

I've reworded your comment a bit. Try to write documentation as full, grammatically-correct sentences, ideally with some context as to what it is commenting. That is, not `/- needed ...` but `/- We increase the heartbeats because ...`

That said, is this comment about multiple large types in this file true? Is there any other problematically large inductive type besides `Op`? Or is is the case that it's not only defining `Op` that causes heartbeats problems, but also definitions that match on `Op`? If the latter is true, please include that explicitly in the comment!

alexkeizer (changes requested):
Thanks for cleaning this up, it already looks much better!

I think you missed a few small ones, still: 
* please clarify the `-- fence` comment a bit more, and
* please fix the header in the docstring for Op.sig (around line 135). Docstrings shouldn't start with headers, those should go in a `/-! -/` style comment above it!

Once that is done and you clarify the heartbeat comment, this looks good to merge!

SSA/Projects/RISCV64/Base.lean
alexkeizer: BTW: if you accept this suggestion via GitHub, do make sure to remove the old comment!

salinhkuhn:
> Thanks for cleaning this up, it already looks much better!
> 
> I think you missed a few small ones, still:
yes, so I understood from the assignment on zulip that I should first do a PR only addressing the heartbeats so that's why I did not include all changes yet. At what granularity should I do this in the future ? I got told to first fix one comment but I guess then its fine to do multiple "smaller" comments at once ?

salinhkuhn:
e.g I also rewrote the defintions to use def_denote and did not include that here yet.

SSA/Projects/RISCV64/Base.lean
```diff
@@ -1,26 +1,31 @@
 import SSA.Projects.RISCV64.Semantics
 import SSA.Core.Framework
+/- This file has a number of very large inductive types, which seem to cause Lean to run out of heartbeats.
+We avoid the issue by increasing the heartbeats. Since this applies to most inductives in this file, we do so globally.
+Additionally, this file contains definitions that match on these large inductive types,. These also causes Lean to require
+more heartbeats.  -/
```
tobiasgrosser:
```suggestion
Additionally, this file contains definitions that match on these large inductive types. These also causes Lean to require
more heartbeats.  -/
```

alexkeizer:
> yes, so I understood from the assignment on zulip that I should first do a PR only addressing the heartbeats so that's why I did not include all changes yet. At what granularity should I do this in the future ? I got told to first fix one comment but I guess then its fine to do multiple "smaller" comments at once?

Sorry about the confusion there! Especially at first, more granular is better, *so long* as the scope of your PR is clear and well-defined. Since you called this pr `feature: added comment and links` I thought you intended this PR to address all the comments-related review comments, thus I pointed out some of those that you didn't address. Especially since you *did* include some changes to the `Op` docstring as well as the heartbeats.

For example, if you'd named this PR `chore: document heartbeat setting in RiscV dialect`, then it'd be clear that you intend only that change (and then I would've told you to leave the change to the Op docstring for the next PR). This is why it's important to have a good description of your PR, at the very least by giving at a clear name, but it doesn't hurt to write a short sentence or two to describe the change if the PR name alone is not completely self-explanatory.


> e.g I also rewrote the defintions to use def_denote and did not include that here yet.

Yes, that's the right call! Changing the semantics to use `def_denote` is a clearly scoped, well-contained change that deserves it's own PR.

In general, it's a bit of an art to determine what makes a good PR. The appropriate size of a PR is closely related to how easy the changes are to review, but that's of course very hard to judge for somebody that is new to the process. We'll inevitably give more feedback on stuff you might want to include/leave out for more of your PRs, and hopefully in the process you'll build up an intuition for it!

As a broad rule of thumb: try to think a bit about how difficult it is to determine whether a change you propose is correct, and how bad the fall-out is if you've made a mistake. If you're just changing documentation, it's relatively easy to confirm that nothing breaks (since we haven't actually changed any functionality), so it's fine to have a bigger PR where you improve documentation for an entire file. 
If you change the semantics of a dialect, however, that is pretty crucial to correctness of the whole project, so that requires much more scrutiny: changing to `def_denote` definitely deserves a dedicated PR.

alexkeizer (approved):
LGTM, please accept Tobias' suggestion and then feel free to hit the merge button (I think you should be able to do so, but if not let me know and I'll do it)

tobiasgrosser:
@salinhkuhn, this is ready to go from @alexkeizer side. It would be useful if you address his last comment ("accept Tobias' suggestion") and then merge as you read this. This will reduce mental state for us and we can focus on the next steps.

PR #1208, opened 2025-05-12
feat: generic dialect parsing utilites & improve LLVM parser

alexkeizer:
This PR adds a set of utility functions that are helpful when defining a dialect parser, and uses these new utilities to clean up the LLVM dialect parser. These utilities are defined in a new `SSA.Core.MLIRSyntax.Transform.Utilities` file which will have to be imported separately by a dialect implementation.

In particular, there is the `AST.Op.mkExprOf` utility, which takes in a parsed `d.Op` together with an `AST.Op` (remember that the latter corresponds more to a parsed `Expr`, since it includes argument variables also), then parses the argument variables according to the signature of the parsed operation to construct an `Expr` (or throw an error if the arguments aren't correct). And there are some simple utilities to work with attributes, such as `AST.Op.hasAttr` to check if an attribute is present, or `AST.Op.getIntAttr` to fetch an integer attribute value, or throw a value if the attribute is not present or not an integer.

Note, though, that in LLVM we need to parse at least some variables to determine the width of the operation. Thus, we have added some API for parsing arguments more easily as well. To avoid doing double work, the aforementioned `mkExprOf` has an optional argument for these parsed variables. For example, if we have some `opStx : AST.Op` which we've determined is a binary operation, we could write the following:
```lean
let args <- opStx.parseArgs  -- parse arguments into a list of indeterminate length
let args <- args.assumeArity -- assert that there must be exactly two arguments
let <ty, _> := args[0]              -- access the type of the first argument
let op := ...
opStx.mkExprOf (args? := some args) op
```
See the LLVM parser for a more detailed example usage.

Finally, we change the way the LLVM parser to look at the operation name *first*, and assert the corresponding arity after. That is, if it tries to parse an add operation with 3 arguments, before it would complain that "add is not a recognized tertiary operation" which would mean that either "add" is an invalid operation, or that it was simply called with the wrong arity. After this PR, it would throw an error saying that add expects exactly two arguments (because we called `args.assumeArity 2` in that branch of the parser!). This also ensures that an `unsupportedOp` error is thrown iff the operation name was not recognized, which is the intended contract but which was *not* the case before. Thus fixing some potential problems in compound dialect parsers built on top of LLVM.


alexkeizer:
@luisacicolini @salinhkuhn Please have a look at this PR and how I cleaned up the LLVM dialect parser here. I feel the utility functions I added here could be very useful for the Comb and RiscV parsers as well!

SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -12,244 +13,123 @@ open InstCombine (LLVM MetaLLVM MOp Width)
...
-  mkExpr := mkExpr
+  let args ← opStx.parseArgs Γ
+
+  /- `binW` is the (lazily computed) width, assuming a binary operation -/
+  let binW : AST.ReaderM (MetaLLVM φ) (Width φ) := do
+  -- ^^ NOTE: `binW` is bound with := rather than ← on purpose, to ensure laziness
+    let args ← args.assumeArity 2
+    return getVarWidth args[0]
+
+  /- `unW` is the (lazily computed) width, assuming a unary operation -/
```
tobiasgrosser:
```suggestion
  /- `unW` is the (lazily computed) width, assuming an unary operation -/
```

SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -12,244 +13,123 @@ open InstCombine (LLVM MetaLLVM MOp Width)
...
+  let unW := do
+    let args ← args.assumeArity 1
+    return getVarWidth args[0]
+
+  let mkExprOf := opStx.mkExprOf (args? := args) Γ
+  match opStx.name with
+    -- Tertiary Operations
```
tobiasgrosser:
```suggestion
    -- Ternary Operations
```

tobiasgrosser (approved):
Wow, this is so much nicer. I am excited to see this merged.

PR #1212, opened 2025-05-13
feat: hybrid RiscV plus LLVM dialect [1/3]

salinhkuhn:
This PR introduces the core definitions for the Hybrid Dialect, which merges functionality from both the RV64 and InstCombine dialects. It includes:
-Definitions of types and operands.
-Implementations of typeclass instances for the dialect.
-functions to assist with dialect parsing, specifically for simultaneously mapping values and types.
The full parsing infrastructure will be included in a follow-up PR.

### DIFF
--- a/SSA/Core/HVector.lean
+++ b/SSA/Core/HVector.lean
@@ -222,5 +222,38 @@ instance [Lean.ToExpr α] [∀ a, Lean.ToExpr (A a)] [HVector.ToExprPi A]
 
 end ToExprPi
 
+/-!
+  ## Folding and mapping over HVectors, monadic and non-monadic.
+-/
+section mapHVec
+def _root_.HVector.foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
+    ∀ {l : List α}, (init : B) → (as : HVector A l) → m B
+  | [],   b, .nil       => return b
+  | t::_, b, .cons a as => do foldlM f (← f t b a) as
+
+/-! Simultaneous map on the type and value level of an HVector. -/
+def _root_.HVector.ubermap {A : α → Type} {B : β → Type}
+    {l : List α}
+    (F : α → β)
+    (f : {a : α} → (v : A a) → B (F a) )
+    (as : HVector A l) : (HVector B (F <$> l)) :=
+  match l, as with
+  | [], .nil => .nil
+  | _t :: _ts, .cons a as => HVector.cons (f a) (HVector.ubermap F f as)
+
+/-!
+Simultaneous map on the type and value level of an HVector while
+performing monadic effects for value translation.-/
+def _root_.HVector.ubermapM [Monad m] {α : Type 0} {A : α → Type} {B : β → Type}
+    {l : List α}
+    {F : α → β}
+    (f : (a : α) → (v : A a) → m (B (F a)) )
+    (as : HVector A l) : m (HVector B (F <$> l)) :=
+  match l, as with
+  | [], .nil => return .nil
+  | t :: _ts, .cons a as => do return HVector.cons (← f t a) (← HVector.ubermapM f as)
+
+end mapHVec
+
 
 end HVector
--- a/SSA/Core/HVector.lean
+++ b/SSA/Core/HVector.lean
@@ -70,6 +70,24 @@ def foldld {β : Type*} (B : β → Type*) (fType : β → α → β)
   | [], .nil, _, init         => init
   | _::_, .cons a as, _, init => foldld B fType fElem as (fElem init a)
 
+
+def foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
+    ∀ {l : List α}, (init : B) → (as : HVector A l) → m B
+  | [],   b, .nil       => return b
+  | t::_, b, .cons a as => do foldlM f (← f t b a) as
+
+/--
+Simultaneous map on the type and value level of an HVector while
+performing monadic effects for value translation.-/
+def mapM' [Monad m] {α : Type 0} {A : α → Type} {B : β → Type}
+    {l : List α}
+    {F : α → β}
+    (f : (a : α) → (v : A a) → m (B (F a)) )
+    (as : HVector A l) : m (HVector B (F <$> l)) :=
+  match l, as with
+  | [], .nil => return .nil
+  | t :: _ts, .cons a as => do return HVector.cons (← f t a) (← HVector.mapM' f as)
+
 def get {as} : HVector A as → (i : Fin as.length) → A (as.get i)
   | .nil, i => i.elim0
   | .cons x  _, ⟨0,   _⟩  => x
@@ -221,30 +239,5 @@ instance [Lean.ToExpr α] [∀ a, Lean.ToExpr (A a)] [HVector.ToExprPi A]
     toExpr }
 
 end ToExprPi
-
-/-!
-  ## Folding and mapping over HVectors, monadic and non-monadic.
--/
-section mapHVec
-
-def foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
-    ∀ {l : List α}, (init : B) → (as : HVector A l) → m B
-  | [],   b, .nil       => return b
-  | t::_, b, .cons a as => do foldlM f (← f t b a) as
-
-/-!
-Simultaneous map on the type and value level of an HVector while
-performing monadic effects for value translation.-/
-def mapM' [Monad m] {α : Type 0} {A : α → Type} {B : β → Type}
-    {l : List α}
-    {F : α → β}
-    (f : (a : α) → (v : A a) → m (B (F a)) )
-    (as : HVector A l) : m (HVector B (F <$> l)) :=
-  match l, as with
-  | [], .nil => return .nil
-  | t :: _ts, .cons a as => do return HVector.cons (← f t a) (← HVector.mapM' f as)
-
-end mapHVec
-
 
 end HVector
--- a/SSA/Core/HVector.lean
+++ b/SSA/Core/HVector.lean
@@ -226,13 +226,13 @@ end ToExprPi
   ## Folding and mapping over HVectors, monadic and non-monadic.
 -/
 section mapHVec
-def _root_.HVector.foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
+def foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
     ∀ {l : List α}, (init : B) → (as : HVector A l) → m B
   | [],   b, .nil       => return b
   | t::_, b, .cons a as => do foldlM f (← f t b a) as
 
 /-! Simultaneous map on the type and value level of an HVector. -/
-def _root_.HVector.ubermap {A : α → Type} {B : β → Type}
+def ubermap {A : α → Type} {B : β → Type}
     {l : List α}
     (F : α → β)
     (f : {a : α} → (v : A a) → B (F a) )
@@ -244,7 +244,7 @@ def _root_.HVector.ubermap {A : α → Type} {B : β → Type}
 /-!
 Simultaneous map on the type and value level of an HVector while
 performing monadic effects for value translation.-/
-def _root_.HVector.ubermapM [Monad m] {α : Type 0} {A : α → Type} {B : β → Type}
+def ubermapM [Monad m] {α : Type 0} {A : α → Type} {B : β → Type}
     {l : List α}
     {F : α → β}
     (f : (a : α) → (v : A a) → m (B (F a)) )
--- a/SSA/Core/HVector.lean
+++ b/SSA/Core/HVector.lean
@@ -232,27 +232,17 @@ def foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
   | [],   b, .nil       => return b
   | t::_, b, .cons a as => do foldlM f (← f t b a) as
 
-/-! Simultaneous map on the type and value level of an HVector. -/
-def ubermap {A : α → Type} {B : β → Type}
-    {l : List α}
-    (F : α → β)
-    (f : {a : α} → (v : A a) → B (F a) )
-    (as : HVector A l) : (HVector B (F <$> l)) :=
-  match l, as with
-  | [], .nil => .nil
-  | _t :: _ts, .cons a as => HVector.cons (f a) (HVector.ubermap F f as)
-
 /-!
 Simultaneous map on the type and value level of an HVector while
 performing monadic effects for value translation.-/
-def ubermapM [Monad m] {α : Type 0} {A : α → Type} {B : β → Type}
+def mapM' [Monad m] {α : Type 0} {A : α → Type} {B : β → Type}
     {l : List α}
     {F : α → β}
     (f : (a : α) → (v : A a) → m (B (F a)) )
     (as : HVector A l) : m (HVector B (F <$> l)) :=
   match l, as with
   | [], .nil => return .nil
-  | t :: _ts, .cons a as => do return HVector.cons (← f t a) (← HVector.ubermapM f as)
+  | t :: _ts, .cons a as => do return HVector.cons (← f t a) (← HVector.mapM' f as)
 
 end mapHVec
 
--- a/SSA/Core/HVector.lean
+++ b/SSA/Core/HVector.lean
@@ -70,7 +70,6 @@ def foldld {β : Type*} (B : β → Type*) (fType : β → α → β)
   | [], .nil, _, init         => init
   | _::_, .cons a as, _, init => foldld B fType fElem as (fElem init a)
 
-
 def foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
     ∀ {l : List α}, (init : B) → (as : HVector A l) → m B
   | [],   b, .nil       => return b
--- a/SSA/Core/HVector.lean
+++ b/SSA/Core/HVector.lean
@@ -226,6 +226,7 @@ end ToExprPi
   ## Folding and mapping over HVectors, monadic and non-monadic.
 -/
 section mapHVec
+
 def foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
     ∀ {l : List α}, (init : B) → (as : HVector A l) → m B
   | [],   b, .nil       => return b
--- a/SSA/Projects/InstCombine/Base.lean
+++ b/SSA/Projects/InstCombine/Base.lean
@@ -134,7 +134,6 @@ instance : DecidableEq LLVM.Ty :=
 instance : DecidableEq LLVM.Op :=
     inferInstanceAs <| DecidableEq (InstCombine.MOp 0)
 
-
 @[deprecated "Use `LLVM.Op` instead" (since:="2025-04-30")] abbrev Op := LLVM.Op
 @[deprecated "Use `LLVM.Ty` instead" (since:="2025-04-30")] abbrev Ty := LLVM.Ty
 
--- a/SSA/Projects/InstCombine/Base.lean
+++ b/SSA/Projects/InstCombine/Base.lean
@@ -126,6 +126,15 @@ def LLVM : Dialect where
   Op := MOp 0
   Ty := MTy 0
 
+/-- Defining an instance for LLVM.Ty from InstCombine.Ty instance.-/
+instance : DecidableEq LLVM.Ty :=
+  inferInstanceAs <| DecidableEq (InstCombine.MTy _)
+
+/-- Defining an instance for LLVM.Op from InstCombine.Op instance. -/
+instance : DecidableEq LLVM.Op :=
+    inferInstanceAs <| DecidableEq (InstCombine.MOp 0)
+
+
 @[deprecated "Use `LLVM.Op` instead" (since:="2025-04-30")] abbrev Op := LLVM.Op
 @[deprecated "Use `LLVM.Ty` instead" (since:="2025-04-30")] abbrev Ty := LLVM.Ty
 
--- a/SSA/Projects/LLVMAndRiscv.lean
+++ b/SSA/Projects/LLVMAndRiscv.lean
@@ -16,8 +16,8 @@ open InstCombine(LLVM)
 namespace LLVMRiscV
 
 
-/-! 
-# Hybrid dialect 
+/-!
+# Hybrid dialect
 
 This file contains a hybrid dialect combining
 SSA.Projects.RISCV64 and SSA.Projects.InstCombine.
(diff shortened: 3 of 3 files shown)


SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,164 @@
...
+    (as : HVector A l) : (HVector B (F <$> l)) :=
+  match l, as with
+  | [], .nil => .nil
+  | _t :: _ts, .cons a as => HVector.cons (f a) (HVector.ubermap F f as)
+
+/-!
+Simultaneous map on the type and value level of an HVector while
+performing monadic effects for value translation.-/
+@[simp_denote]
+def _root_.HVector.ubermapM [Monad m] {A : α → Type} {B : β → Type}
```
tobiasgrosser: Should this be in the HVector file?

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,164 @@
...
+  Signature.outTy (f <$> s) = f s.outTy := rfl
+
+def _root_.HVector.foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
+    ∀ {l : List α}, (init : B) → (as : HVector A l) → m B
+  | [],   b, .nil       => return b
+  | t::_, b, .cons a as => do foldlM f (← f t b a) as
+
+/-! Simultaneous map on the type and value level of an HVector. -/
+@[simp_denote]
+def _root_.HVector.ubermap {A : α → Type} {B : β → Type}
```
tobiasgrosser: Should this be in the HVector file?

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,164 @@
...
+    | .riscv someRiscVTy => someRiscVTy
+    | _  => .bv
+  ) Γ
+
+/-- Projection of `outTy` commutes with `Signature.map`. -/
+@[simp, simp_denote]
+theorem outTy_map_signature_eq {s : Signature α} {f : α → β} :
+  Signature.outTy (f <$> s) = f s.outTy := rfl
+
+def _root_.HVector.foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
```
tobiasgrosser: Should this be in the HVector file?

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: That makes sense. Perhaps @alexkeizer can share his thoughts on whether he'd like to have this available outside of my project.

tobiasgrosser: He is likely to agree. Let's move this to not add latency.

salinhkuhn: I just tried moving it to the HVector file. However, since I tagged the lemma to be part of the simp_denote set so that simp_peephole can apply it, I’m not sure how to move it correctly. E.g moving it to HVector.lean  would require HVector to import, for example, SSA.Core.Tactic, which leads to cyclic imports. Since the simp_denote attribute is newer, @alexkeizer could you tell me how to relocate those maps correctly ?

tobiasgrosser: Use `attribute` as visible in https://github.com/opencompl/lean-mlir/blob/main/SSA/Projects/InstCombine/ForLean.lean#L385, to attach the attribute locally in your file.

salinhkuhn: thank you. Done.

SSA/Core/HVector.lean
```diff
@@ -222,5 +222,38 @@ instance [Lean.ToExpr α] [∀ a, Lean.ToExpr (A a)] [HVector.ToExprPi A]
 
 end ToExprPi
 
+/-!
+  ## Folding and mapping over HVectors, monadic and non-monadic.
+-/
+section mapHVec
+def _root_.HVector.foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
```
tobiasgrosser: Do we need the `_root_.HVector` prefix.

salinhkuhn: no, not needed

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+import SSA.Projects.RISCV64.Semantics
+import SSA.Projects.RISCV64.PrettyEDSL
+import SSA.Core.HVector
+
+open InstCombine(LLVM)
+namespace LLVMRiscV
+
+
+/-! # Hybrid dialect -/
+/-
```
alexkeizer:
```suggestion
/-! 
# Hybrid dialect 

```

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+we insert unrealized_conversion_cast like MLIR does during lowering.
+see: section (UnrealizedConversionCastOp)
+https://mlir.llvm.org/docs/Dialects/Builtin/#builtinunrealized_conversion_cast-unrealizedconversioncastop
+ -/
+
+/-- Defining an instance for LLVM.Ty from InstCombine.Ty instance.-/
+instance : DecidableEq LLVM.Ty :=
+  fun a b => by
+    simp only [LLVM] at a b
+    exact (inferInstanceAs (DecidableEq (InstCombine.MTy 0)) a b)
```
alexkeizer:
```suggestion
instance : DecidableEq LLVM.Ty :=
  inferInstanceAs <| DecidableEq (InstCombine.MTy _)
```

This ought to work as well, same for the other instances that just wrap the InstCombine.M{Op,Ty} instance.

That said, it's nice to define an instance close to the type it applies to, so could you move all instance for `LLVM.Ty` and `LLVM.Op` to the InstCombine/Base file?

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+
+/-- Defining an instance for LLVM.Op from InstCombine.Op instance. -/
+instance : DecidableEq LLVM.Op :=
+  fun a b => by
+    simp only [LLVM] at a b
+    exact (inferInstanceAs (DecidableEq (InstCombine.MOp 0)) a b)
+
+inductive Ty where
+  | llvm : (LLVM.Ty) -> Ty
+  | riscv : (Dialect.Ty RISCV64.RV64) -> Ty
```
alexkeizer:
```suggestion
  | riscv : RISCV64.RV64.Ty -> Ty
```

Even though the dot notation is maybe slightly mis-leading, it does look nicer and we do use it in other files, so it's best to be consistent.

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+  deriving DecidableEq, Repr
+
+inductive Op where
+  | llvm : (Dialect.Op LLVM) -> Op
+  | riscv : (Dialect.Op RISCV64.RV64) -> Op
```
alexkeizer:
```suggestion
  | llvm : LLVM.Op -> Op
  | riscv : RISCV64.RV64.Op -> Op
```

Could you check this everywhere?

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+    exact (inferInstanceAs (DecidableEq (InstCombine.MTy 0)) a b)
+
+/-- Defining an instance for LLVM.Op from InstCombine.Op instance. -/
+instance : DecidableEq LLVM.Op :=
+  fun a b => by
+    simp only [LLVM] at a b
+    exact (inferInstanceAs (DecidableEq (InstCombine.MOp 0)) a b)
+
+inductive Ty where
+  | llvm : (LLVM.Ty) -> Ty
```
alexkeizer:
```suggestion
  | llvm : LLVM.Ty -> Ty
```
no need for these parentheses

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+  | riscv : (Dialect.Ty RISCV64.RV64) -> Ty
+  deriving DecidableEq, Repr
+
+inductive Op where
+  | llvm : (Dialect.Op LLVM) -> Op
+  | riscv : (Dialect.Op RISCV64.RV64) -> Op
+  | builtin.unrealized_conversion_cast.riscvToLLVM : Op
+  | builtin.unrealized_conversion_cast.LLVMToriscv : Op
```
alexkeizer:
```suggestion
  /-- Cast a RiscV value to an LLVM value -/
  | castRiscv : Op
  /-- Cast an LLVM value to a RiscV value -/
  | castLLVM : Op
```

Let's try to keep the operation names a bit short. Note that the way we need constructors of this inductive type is not strictly related to the way MLIR names those operations, so we don't have to repeat the whole `builtin.unrealized_conversion_cast` namespace.

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+We wrap `BitVec 64`in `Option (BitVec 64)` -/
+def builtin.unrealized_conversion_cast.riscvToLLVM (toCast : BitVec 64 ) : Option (BitVec 64 ) := some toCast
+
+/--
+Semantics of an unrealized conversion cast from LLVM to RISC-V.
+This cast attempts to lower an `(Option (BitVec 64))` to a concrete `(BitVec 64)`.
+If the value is `some`, we extract the underlying `BitVec`.
+If it is `none` (e.g., an LLVM poison value), we default to the zero `BitVec`.
+-/
+def builtin.unrealized_conversion_cast.LLVMToriscv (toCast : Option (BitVec 64)) : BitVec 64 := toCast.getD 0#64
```
alexkeizer: Same here, the dots here add a bunch of namespaces on the Lean side we don't really need, could you rename it into something a bit shorter without dots?

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+  | .builtin.unrealized_conversion_cast.riscvToLLVM =>
+      {sig := [Ty.riscv .bv], outTy := Ty.llvm (.bitvec 64), regSig := []}
+  | .builtin.unrealized_conversion_cast.LLVMToriscv =>
+      {sig := [Ty.llvm (.bitvec 64)], outTy := (Ty.riscv .bv), regSig := []}
+
+instance : ToString (Dialect.Ty LLVMPlusRiscV)  where
+  toString t := repr t |>.pretty
+
+@[simp_denote]
+def llvmArgsFromHybrid : {tys : List LLVM.Ty} → HVector TyDenote.toType (tys.map LLVMRiscV.Ty.llvm) → HVector TyDenote.toType tys
```
alexkeizer: This line is too long!

We generally don't want lines that are more than 100 characters wide, so if a line is longer than that please insert newlines. Note that you can configure vscode to show you a light grey line at 100 characters as a guideline!

I believe you might have some more too-long lines throughout the file, please check this.

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+  | [], .nil => .nil
+  | _ :: _, .cons x xs => .cons x (llvmArgsFromHybrid xs)
+
+@[simp_denote]
+def riscvArgsFromHybrid : {tys : List RISCV64.RV64.Ty} → HVector TyDenote.toType (tys.map LLVMRiscV.Ty.riscv) → HVector TyDenote.toType tys
```
alexkeizer: this line is also too long, for example

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+
+@[simp, reducible]
+instance : DialectDenote (LLVMPlusRiscV) where
+  denote
+  | .llvm (llvmOp), args , .nil => DialectDenote.denote llvmOp (llvmArgsFromHybrid args) .nil
+  | .riscv (riscvOp), args , .nil => DialectDenote.denote riscvOp (riscvArgsFromHybrid args) .nil
+  | .builtin.unrealized_conversion_cast.riscvToLLVM, elemToCast, _ =>
+    let toCast : (BitVec 64) := (elemToCast.getN 0 (by simp [DialectSignature.sig, signature])) -- adapting to the newly introduced PoisonOr wrapper.
+    let casted : Option (BitVec 64) := builtin.unrealized_conversion_cast.riscvToLLVM toCast
+    PoisonOr.ofOption casted
```
alexkeizer: Please have the cast semantics function return a `PoisonOr`, then you don't have to use `ofOption` here

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,144 @@
...
+inductive Op where
+  | llvm : (Dialect.Op LLVM) -> Op
+  | riscv : (Dialect.Op RISCV64.RV64) -> Op
+  | builtin.unrealized_conversion_cast.riscvToLLVM : Op
+  | builtin.unrealized_conversion_cast.LLVMToriscv : Op
+  deriving DecidableEq, Repr
+
+/-- Semantics of an unrealized conversion cast from RISC-V 64 to LLVM.
+We wrap `BitVec 64`in `Option (BitVec 64)` -/
+def builtin.unrealized_conversion_cast.riscvToLLVM (toCast : BitVec 64 ) : Option (BitVec 64 ) := some toCast
```
alexkeizer:
```suggestion
def builtin.unrealized_conversion_cast.riscvToLLVM (toCast : BitVec 64) : PoisonOr (BitVec 64) := .value toCast
```

Please check your whitespace throughout the file!

SSA/Projects/LLVMAndRiscv.lean
alexkeizer:
```suggestion
def builtin.unrealized_conversion_cast.LLVMToriscv (toCast : PoisonOr (BitVec 64)) : BitVec 64 := 
  toCast.toOption.getD 0#64
```

SSA/Core/HVector.lean
```diff
@@ -222,5 +222,39 @@ instance [Lean.ToExpr α] [∀ a, Lean.ToExpr (A a)] [HVector.ToExprPi A]
...
+def foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
+    ∀ {l : List α}, (init : B) → (as : HVector A l) → m B
+  | [],   b, .nil       => return b
+  | t::_, b, .cons a as => do foldlM f (← f t b a) as
+
+/-! Simultaneous map on the type and value level of an HVector. -/
+def ubermap {A : α → Type} {B : β → Type}
+    {l : List α}
+    (F : α → β)
+    (f : {a : α} → (v : A a) → B (F a) )
```
alexkeizer, line 239:
```suggestion
    (f : {a : α} → (v : A a) → B (F a))
```

SSA/Core/HVector.lean
```diff
@@ -222,5 +222,39 @@ instance [Lean.ToExpr α] [∀ a, Lean.ToExpr (A a)] [HVector.ToExprPi A]
...
+    (as : HVector A l) : (HVector B (F <$> l)) :=
+  match l, as with
+  | [], .nil => .nil
+  | _t :: _ts, .cons a as => HVector.cons (f a) (HVector.ubermap F f as)
```
alexkeizer: This seems to be the same as the existing `HVector.map'`, is it?

If it is, please use that one, if it isn't add an explanation to the docstring why `ubermap` is different from `map'`

SSA/Core/HVector.lean
```diff
@@ -222,5 +222,39 @@ instance [Lean.ToExpr α] [∀ a, Lean.ToExpr (A a)] [HVector.ToExprPi A]
...
+
+/-!
+Simultaneous map on the type and value level of an HVector while
+performing monadic effects for value translation.-/
+def ubermapM [Monad m] {α : Type 0} {A : α → Type} {B : β → Type}
```
alexkeizer: I don't think we have a `mapM'`, but if `ubermap` is indeed the same as `map'`, then we should name this `mapM'` to match.

SSA/Projects/LLVMAndRiscv.lean
alexkeizer: (By the way, for this kind of comment, where I ask you to check the whole file, it's nice to add a comment saying you indeed checked the whole thing before hitting resolve. That way I can see you noticed that comment also and didn't just mistakenly only fix this one occurence)

alexkeizer:
@salinhkuhn thanks for polishing and submitting the PR! I left some comments, please have a look at them and let me know when you've addressed everything and the PR is ready for a second round of review

SSA/Projects/LLVMAndRiscv.lean
tobiasgrosser: Perfect. I f you addressed a comment, feel invited to click 'resolve conversation'. (I do this now for the ones that are addressed already).

tobiasgrosser:
@salinhkuhn, it seems that on this PR @alexkeizer comments still need to be addressed, meaning the ball is in your court.  Please address those and when done, pass the ball back to us. 🎾

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: Done. Let me know if the new location works for you.

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: done

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: Should be fix throughout the file now.

SSA/Core/HVector.lean
salinhkuhn: it is the same. I changed it now.

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: done

salinhkuhn:
@alexkeizer @tobiasgrosser Thx, I adressed your comments. Review round 2 can start.

SSA/Core/HVector.lean
```diff
@@ -222,5 +222,29 @@ instance [Lean.ToExpr α] [∀ a, Lean.ToExpr (A a)] [HVector.ToExprPi A]
...
+
+def foldlM {B : Type*} [Monad m] (f : ∀ (a : α), B → A a → m B) :
+    ∀ {l : List α}, (init : B) → (as : HVector A l) → m B
+  | [],   b, .nil       => return b
+  | t::_, b, .cons a as => do foldlM f (← f t b a) as
+
+/-!
+Simultaneous map on the type and value level of an HVector while
+performing monadic effects for value translation.-/
+def mapM' [Monad m] {α : Type 0} {A : α → Type} {B : β → Type}
```
alexkeizer:
```suggestion
/--
Simultaneous map on the type and value level of an HVector while
performing monadic effects for value translation.
-/
def mapM' [Monad m] {α : Type 0} {A : α → Type} {B : β → Type}
```
Remember: `/-- ... -/` is for definitions (so-called docstrings)!

SSA/Projects/LLVMAndRiscv.lean
alexkeizer: @salinhkuhn it seems you didn't respond to this yet. If you don't want to adopt my suggestion, do leave a response here to explain why

alexkeizer (approved):
Thanks, it looks almost perfect. There seems to be one comment still open from the previous review that you haven't responded to, can you have a look at that.

The place you put the HVector lemmas is OK, but even better would be to define the monadic foldlM & mapM' close to their non-monadic versions, for example by putting them right after `foldld` on line 72. That way, all maps/folds are defined in one section of the file, making it easier to find them.

Could you fix these two things, and then we can merge?

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: adapted the change

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: done

salinhkuhn:
> Thanks, it looks almost perfect. There seems to be one comment still open from the previous review that you haven't responded to, can you have a look at that.
> 
> The place you put the HVector lemmas is OK, but even better would be to define the monadic foldlM & mapM' close to their non-monadic versions, for example by putting them right after `foldld` on line 72. That way, all maps/folds are defined in one section of the file, making it easier to find them.
> 
> Could you fix these two things, and then we can merge?

fixed it, thank you for the review Alex.
I moved the foldM and mapM' and shortened the name of the casts and casts semantics. @alexkeizer I will merge as soon as you give me the OK.

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,139 @@
...
+import SSA.Core.Util
+import SSA.Core.Util.ConcreteOrMVar
+import SSA.Projects.InstCombine.ForStd
+import SSA.Projects.InstCombine.LLVM.Semantics
+import SSA.Projects.InstCombine.Tactic
+import SSA.Projects.RISCV64.Syntax
+import SSA.Projects.RISCV64.Base
+import SSA.Projects.RISCV64.Semantics
+import SSA.Projects.RISCV64.PrettyEDSL
+import SSA.Core.HVector
```
tobiasgrosser: @salinhkuhn, can you double-check that this list of imports is minimal. Given that the file is relatively short and this list is long, I suspect not all may be needed.

salinhkuhn: Done. It's true they're not needed for this part of the file yet. They'll be added in the next PRs as soon as necessary.

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -0,0 +1,129 @@
+import SSA.Projects.InstCombine.Base
+import SSA.Projects.RISCV64.Base
+import SSA.Core.HVector
```
tobiasgrosser:
```suggestion
import SSA.Projects.InstCombine.Base
import SSA.Projects.RISCV64.Base
```

tobiasgrosser: It seems you missed one.

tobiasgrosser (approved):
Assuming my last comment is merged, this is good to go. Let's see if @alexkeizer aproves as well.

SSA/Core/HVector.lean
```diff
@@ -70,6 +70,24 @@ def foldld {β : Type*} (B : β → Type*) (fType : β → α → β)
   | [], .nil, _, init         => init
   | _::_, .cons a as, _, init => foldld B fType fElem as (fElem init a)
 
+
```
alexkeizer:
```suggestion
```

SSA/Projects/InstCombine/Base.lean
```diff
@@ -126,6 +126,15 @@ def LLVM : Dialect where
...
 
+/-- Defining an instance for LLVM.Ty from InstCombine.Ty instance.-/
+instance : DecidableEq LLVM.Ty :=
+  inferInstanceAs <| DecidableEq (InstCombine.MTy _)
+
+/-- Defining an instance for LLVM.Op from InstCombine.Op instance. -/
+instance : DecidableEq LLVM.Op :=
+    inferInstanceAs <| DecidableEq (InstCombine.MOp 0)
+
+
```
alexkeizer:
```suggestion
```

alexkeizer (approved):
LGTM! (Stands for Looks Good To Me, i.e., feel free to merge!)

PR #1215, opened 2025-05-14
feat: implement toString instances for Com&  Expr to enable opt tool to pretty print

salinhkuhn:
This PR implements toString functions for the Com and Expr types, enabling the opt tool to print the bb0.mlir file to the command line in a more readable format.
It also enhances the opt tool with a --verbose flag, which uses the Repr instances to print detailed output. If the flag is not provided, the output defaults to the toString representations.
Currently names are not preserved and the return statement is printed as a "return" not yet in the dialect generic syntax.
E.g the following input: 
```
  ^bb0(%arg0: i32):
    %0 = "llvm.mlir.constant"() {value = 8 : i32} : () -> i32
    %1 = "llvm.mlir.constant"() {value = 31 : i32} : () -> i32
    %2 = "llvm.or"(%0, %1) <{isDisjoint}> : (i32, i32) -> i32
    %3 = "llvm.sdiv"(%0, %2) <{isExact}> : (i32, i32) -> i32
    "llvm.return"(%3) : (i32) -> ()
```
will be printed to the command line as follows by invoking lake exe opt:
```
^entry(%0 : i32):
   %1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
   %2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
   %3 = "llvm.or disjoint" (%1, %2) : (i32, i32) → (i32)
   %4 = "llvm.sdiv exact" (%1, %3) : (i32, i32) → (i32)
  return %4 : (i32) → ()
```


SSA/Projects/InstCombine/Base.lean
```diff
@@ -386,6 +404,7 @@ namespace LLVM.Ty
 def width : LLVM.Ty → Nat
   | bitvec w => w
 
+
```
tobiasgrosser: Unrelated.

tobiasgrosser: Unrelated.

tobiasgrosser:
The CI fails with:

> info: Except.ok ⟨EffectKind.pure, ⟨i32, "llvm.and" (%3, %1) : (i32, i32) → (i32)⟩⟩
> info: /home/runner/work/lean-mlir/lean-mlir/SSA/Projects/InstCombine/Test.lean:179:15: Except.ok ⟨EffectKind.pure, ⟨i32, "llvm.add" (%4, %3) : (i32, i32) → (i32)⟩⟩
> error: /home/runner/work/lean-mlir/lean-mlir/SSA/Projects/InstCombine/Test.lean:179:0: ❌️ Docstring on `#guard_msgs` does not match generated message:

Also, this PR warns that the base branch is not up-to-date with the latest changes.

SSA/Core/MLIRSyntax/GenericParser.lean
```diff
@@ -350,9 +350,8 @@ info: MLIR.AST.MLIRType.undefined "shape.value"
 #guard_msgs in #eval tyUserIdent
 
 private def tyi32NoGap : MLIRTy := [mlir_type| i32]
-/--
-info: MLIR.AST.MLIRType.int (MLIR.AST.Signedness.Signless) (ConcreteOrMVar.concrete 32)
--/
+
+/-- info: MLIR.AST.MLIRType.int (MLIR.AST.Signedness.Signless) i32 -/
```
tobiasgrosser: Is this change needed?

tobiasgrosser:
Can we have somehow a complete guard_msg test case that shows how these changes work. Similarly, it would be nice if the commit message would show the before/after of the `opt` output for our `bb` test case.

SSA/Core/MLIRSyntax/GenericParser.lean
salinhkuhn: yes, else it complains. It's because now we print i32 and not  (ConcreteOrMVar.concrete 32).

salinhkuhn:
> The CI fails with:
> 
> > info: Except.ok ⟨EffectKind.pure, ⟨i32, "llvm.and" (%3, %1) : (i32, i32) → (i32)⟩⟩
> > info: /home/runner/work/lean-mlir/lean-mlir/SSA/Projects/InstCombine/Test.lean:179:15: Except.ok ⟨EffectKind.pure, ⟨i32, "llvm.add" (%4, %3) : (i32, i32) → (i32)⟩⟩
> > error: /home/runner/work/lean-mlir/lean-mlir/SSA/Projects/InstCombine/Test.lean:179:0: ❌️ Docstring on `#guard_msgs` does not match generated message:
> 
> Also, this PR warns that the base branch is not up-to-date with the latest changes.

I just checked that and they fail because they expect the printing to be different. They require the output to be more verbose e.g : 

```Except.ok ⟨EffectKind.pure, ⟨i32, InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 8 : () → (i32)⟩⟩```


 but with the new printing we get 

```Except.ok ⟨EffectKind.pure, ⟨i32, "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)⟩```

**Comment by salinhkuhn** [comment]
> Can we have somehow a complete guard_msg test case that shows how these changes work. Similarly, it would be nice if the commit message would show the before/after of the `opt` output for our `bb` test case.

I added before and after to the commit message now. I also pushed again and added a testcase to the beginning of the Test.lean.

**Comment by alexkeizer** [comment]
Firstly, `Repr` has the implicit contract that it's output is expected to be parse-able Lean code (resulting in the same object). It is thus not a good idea to use Repr for our specialized pretty-printing purposes. However, `ToString` has no such contract, and is totally fair game. Would it be possible to use only `ToString` for pretty printing, leaving the `Repr` instance as is? That way, there is not strictly a need for a new typeclass (yet): we already have the distinction between Repr and ToString.

That said, for LLVM, specifically, I'm curious how attributes are handled. Could you include an example program that has those, for example, how does a program with an `llvm.add %2, %0 overflow<nuw> : i32` instruction get pretty-printed? It's fine if they are ignored for now (after all, we are primarily interested in RiscV, which shouldn't have attributes), but, if that is the case, make sure you document this very explicitly (both in the PR description and in the code).

File: SSA/Core/MLIRSyntax/GenericParser.lean (hunk 2 above)
Comment (alexkeizer) [reply SSA/Core/MLIRSyntax/GenericParser.lean]: Should this not print either `i32` or `[mlir_type| i32]` (rather than MLIR.AST.MLIRType.int (MLIR.AST.Signedness.Signless) ... full-out), since the fact that it is a singless int is already implied by `i32`?

File: SSA/Core/Util/ConcreteOrMVar.lean [hunk 3]
```diff
@@ -12,12 +12,28 @@ import Mathlib.Data.Vector.Basic
...
+  format x :=
+    match x with
+    | .concrete a => f!"i{a}"
+    | .mvar i => f!"mvar({i})"
+
+instance : ToString (ConcreteOrMVar Nat φ) where
+  toString x :=
+    match x with
+    | .concrete a => s!"i{a}"
+    | .mvar i => s!"mvar({i})"
```
alexkeizer: Do we need all these three instances? Could we leave the Repr instance as the derived one, and only use `ToString` for pretty-printing?

alexkeizer: Note also that there is `ConcreteOrMVar` is not only used for signless-integer types. They could be used as the width of an unsigned int, or even as the value of a (non-negative) constant.

For example, IIUC, the type `u32` would be pretty-printed as:
`MLIR.AST.MLIRType.int (MLIR.AST.Signedness.Unsigned) i32`
which should be `u32` or `[mlir_type| u32]` instead!

I believe you should do the pretty printing at the `MLIRType` level instead, since that's where you can actually inspect the signedness and determine if you should prefix it with an `i` or `u` or `si` depending on said signedness.

SSA/Projects/InstCombine/Base.lean
```diff
@@ -99,7 +99,13 @@ def reprWithoutFlags (op : MOp.BinaryOp) (prec : Nat) : Format :=
     | .udiv ⟨false⟩        => "udiv"
     | .udiv ⟨true⟩         => "udiv exact"
   Repr.addAppParen (Format.group (Format.nest
-    (if prec >= max_prec then 1 else 2) f!"InstCombine.MOp.BinaryOp.{op}"))
+    (if prec >= max_prec then 1 else 2) f!"llvm.{op}"))
+/- Changed this to use the "llvm" prefix for better pretty-printing.
+This affects every use of `Repr` for `MOp.BinaryOp`, which may not be desirable.
+TODO: Consider whether this change is acceptable globally.
+An alternative would be to introduce a separate `Serialize` typeclass that is same as `Repr` but uses the "llvm.{op}" format,
+while keeping `Repr` more verbose "InstCombine.MOp.BinaryOp.{op}". -/
```
alexkeizer: As mentioned on Zulip: for `Repr` there is the expectation that the result is valid Lean code, so this is indeed not acceptable. However, `ToString` has no such expectation, so if you define a `ToString` instance like this and use that for pretty printing, that is totally fine!

SSA/Projects/InstCombine/Base.lean
```diff
@@ -115,6 +121,15 @@ inductive MOp (φ : Nat) : Type
...
 
+instance : Repr (MOp 0) where
+   reprPrec op _p :=
+     match op with
+     | .unary _w op => f!"\"{repr  op}\""
+     | .binary _w op => f!"\"{repr  op}\""
+     | .select  _w => repr "select"
+     | .icmp  _pred _w => repr "icmp"
+     | .const w val => f!"\"llvm.mlir.constant\" \{ value = {val} : {w} }"
+
```
alexkeizer: Here, too, could we make this the `ToString` instance rather than `Repr`?

alexkeizer (commented):
I think we should move away from `Repr` for pretty-printing, as we're breaking some of the implicit `Repr` contract with these instances! In future we could add a new DialectSerialize typeclass, but for now ToString should suffice, so let's try that first.

SSA/Core/Framework.lean
```diff
@@ -267,9 +267,48 @@ def Lets.repr (prec : Nat) : Lets d eff Γ t → Format
 instance : Repr (Expr d Γ eff t) := ⟨flip Expr.repr⟩
 instance : Repr (Com d Γ eff t) := ⟨flip Com.repr⟩
 instance : Repr (Lets d Γ eff t) := ⟨flip Lets.repr⟩
-
```
tobiasgrosser, line 270: unrelated

SSA/Core/MLIRSyntax/GenericParser.lean
```diff
@@ -350,9 +350,8 @@ info: MLIR.AST.MLIRType.undefined "shape.value"
 #guard_msgs in #eval tyUserIdent
 
 private def tyi32NoGap : MLIRTy := [mlir_type| i32]
-/--
-info: MLIR.AST.MLIRType.int (MLIR.AST.Signedness.Signless) (ConcreteOrMVar.concrete 32)
--/
+
+/-- info: MLIR.AST.MLIRType.int (MLIR.AST.Signedness.Signless) (ConcreteOrMVar.concrete 32) -/
```
tobiasgrosser: I think you can avoid this diff by keeping the original formatting.

SSA/Core/Util/ConcreteOrMVar.lean
```diff
@@ -5,19 +5,34 @@ import Mathlib.Data.Fin.Basic
 import Mathlib.Data.Vector.Basic
 -- import Mathlib.Tactic
 
-
```
tobiasgrosser: unrelated.

salinhkuhn:
So I implemented part of the toString functions for Expr and Com. When I run 
lake exe opt on bb0.mlir

I get the following output: 
```
{ 
^entry(%0 : i32):
%1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
%2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
%3 = "llvm.ashr" (%0, %2) : (i32, i32) → (i32)
%4 = "llvm.and" (%3, %1) : (i32, i32) → (i32)
%5 = "llvm.add" (%4, %3) : (i32, i32) → (i32)
return %5 : (i32) → ()
 }
 ```
 if we pass the --verbose flag, it should print using the Repr and output :
 ```
Flag `--verbose` was set.
⟨EffectKind.pure, ⟨i32, {
    ^entry(%0 : i32):
      %1 = InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 8 : () → (i32)
      %2 = InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 31 : () → (i32)
      %3 = InstCombine.MOp.binary
        (ConcreteOrMVar.concrete 32)
        (InstCombine.MOp.BinaryOp.ashr) (%0, %2) : (i32, i32) → (i32)
      %4 = InstCombine.MOp.binary
        (ConcreteOrMVar.concrete 32)
        (InstCombine.MOp.BinaryOp.and) (%3, %1) : (i32, i32) → (i32)
      %5 = InstCombine.MOp.binary
        (ConcreteOrMVar.concrete 32)
        (InstCombine.MOp.BinaryOp.add) (%4, %3) : (i32, i32) → (i32)
      return %5 : (i32) → ()
  }
```
Attributes especially flags are not working yet. But it would be nice to get a quick opinion if that's a path to follow further.

SSA/Projects/InstCombine/Test.lean
```diff
@@ -108,6 +119,7 @@ info: Except.ok ⟨EffectKind.pure, ⟨i32, InstCombine.MOp.const (ConcreteOrMVa
 -/
 #guard_msgs in #eval mkExpr    (Γn 2) op1    ["0", "arg0"]
 
+#eval mkExpr    (Γn 3) op2    ["1", "0", "arg0"]
```
tobiasgrosser: These changes are not needed, no?

salinhkuhn: No, I shall remove them. It was to figure out why the CI failed on the previous push.

tobiasgrosser:
> So I implemented part of the toString functions for Expr and Com. When I run lake exe opt on bb0.mlir
> 
> I get the following output:
> 
> ```
> { 
> ^entry(%0 : i32):
> %1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
> %2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
> %3 = "llvm.ashr" (%0, %2) : (i32, i32) → (i32)
> %4 = "llvm.and" (%3, %1) : (i32, i32) → (i32)
> %5 = "llvm.add" (%4, %3) : (i32, i32) → (i32)
> return %5 : (i32) → ()
>  }
> ```
> 
> if we pass the --verbose flag, it should print using the Repr and output :
> ```

Nice. I left some comments. Otherwise, this LGTM, but let's see what @alexkeizer says.

You could also indent the `%***` lines by a bit, making the overall code look nicer.

SSA/Core/Util/ConcreteOrMVar.lean
salinhkuhn: I got stuck there. If we could look at that in the call later that would be helpful. But I will look into it again now

salinhkuhn: > Do we need all these three instances? Could we leave the Repr instance as the derived one, and only use `ToString` for pretty-printing?

I removed the explicit Repr instance and toFormat instance. Therefore only the toString instance is left.

SSA/Core/Framework.lean
```diff
@@ -267,9 +267,48 @@ def Lets.repr (prec : Nat) : Lets d eff Γ t → Format
...
 end Repr
 
+/- # ToString instances for Com and Expr  -/
+section ToString
+variable {d} [DialectSignature d] [Repr d.Op] [Repr d.Ty][ToString d.Ty][ToString d.Op]
+
+/-- Format a list of formal arguments as `(%0 : t₀, %1 : t₁, ... %n : tₙ)` -/
+partial def formatFormalArgListTupleStr [ToString Ty] (ts : List Ty) : String :=
+  let args := (List.range ts.length).zip ts |>.map
+    (fun (i, t) => s!"%{i} : {toString t}") -- assumes toString function for types. -- tried using it with naming but then difficult to track.
```
alexkeizer: I'm not sure what this comment is saying, could you explain a bit more (doesn't have to be as a comment, feel free to just reply to this comment)?

Also, do keep the max line width in mind: that applies to comments, too!

SSA/Core/Framework.lean
```diff
@@ -267,9 +267,48 @@ def Lets.repr (prec : Nat) : Lets d eff Γ t → Format
...
+  "(" ++ String.intercalate ", " args ++ ")"
+
+  partial def Expr.toString1 [ToString d.Op] : Expr d Γ eff t → String
```
alexkeizer: Why is this called `toString1`, and not just `toString`?
```suggestion
  partial def Expr.toString [ToString d.Op] : Expr d Γ eff t → String
```

SSA/Core/Framework.lean
```diff
@@ -267,9 +267,48 @@ def Lets.repr (prec : Nat) : Lets d eff Γ t → Format
...
+    | Expr.mk (op : d.Op) _ _ args regArgs =>
+        let outTy : d.Ty  := DialectSignature.outTy op
```
```suggestion
        let outTy : d.Ty := DialectSignature.outTy op
```
mind your whitespace (check this thoroughly, please!)

SSA/Core/Framework.lean
```diff
@@ -267,9 +267,48 @@ def Lets.repr (prec : Nat) : Lets d eff Γ t → Format
...
+        let argTys := DialectSignature.sig op
+
+      s!"{toString op}{formatArgTuple args} : {formatTypeTuple argTys} → ({toString outTy})"
+
+partial def comToStringAux : Com d Γ eff t → String
```
alexkeizer: Why not:
```suggestion
partial def Com.toStringAux : Com d Γ eff t → String
```
That would be a bit more consistent. You could even go a bit more descriptive and name it `Com.toStringBody` to emphasize it's formatting just the body of the basic block.

SSA/Core/Framework.lean
```diff
@@ -267,9 +267,48 @@ def Lets.repr (prec : Nat) : Lets d eff Γ t → Format
...
+instance : ToString (Com d Γ eff t)  where toString := Com.toString
+instance : ToString (Expr d Γ eff t) where toString := Expr.toString1
+
+end ToString
+
+
+
+
+
+
```
```suggestion
```
why so many newlines?

SSA/Core/Util/ConcreteOrMVar.lean
```diff
@@ -5,19 +5,34 @@ import Mathlib.Data.Fin.Basic
...
 inductive ConcreteOrMVar (α : Type u) (φ : Nat)
   | concrete (a : α)
   | mvar (i : Fin φ)
   deriving DecidableEq, Repr, Inhabited, Lean.ToExpr
+/-
+instance : Repr (ConcreteOrMVar Nat φ) where
+  reprPrec x _prec :=
+    match x with
+    | .concrete a => f!"i{a}"
+    | .mvar i => f!"mvar({i})" -/
```
alexkeizer: Could you remove this instance, rather than leaving it commented out?
```suggestion
```

SSA/Projects/InstCombine/Base.lean
```diff
@@ -100,6 +100,12 @@ def reprWithoutFlags (op : MOp.BinaryOp) (prec : Nat) : Format :=
     | .udiv ⟨true⟩         => "udiv exact"
   Repr.addAppParen (Format.group (Format.nest
     (if prec >= max_prec then 1 else 2) f!"InstCombine.MOp.BinaryOp.{op}"))
+/- Changed this to use the "llvm" prefix for better pretty-printing.
+This affects every use of `Repr` for `MOp.BinaryOp`, which may not be desirable.
+TODO: Consider whether this change is acceptable globally.
+An alternative would be to introduce a separate `Serialize` typeclass that is same as `Repr` but uses the "llvm.{op}" format,
+while keeping `Repr` more verbose "InstCombine.MOp.BinaryOp.{op}". -/
```
alexkeizer: Is this comment still up-to-date and accurate? If not, we should remove it

SSA/Projects/InstCombine/Base.lean
```diff
@@ -115,6 +121,45 @@ inductive MOp (φ : Nat) : Type
...
+  s!"llvm.{op}"
+
+instance : ToString (MOp.BinaryOp) where
+  toString := toStringWithoutFlags
+
+
+instance : ToString (MOp 0) where
+   toString  op :=
+     match op with
+     | .unary _w op => s!"\"{repr op}\""
```
alexkeizer: Should this not be `toString` rather than `repr`?
```suggestion
     | .unary _w op => s!"\"{toString op}\""
```

alexkeizer (commented):
It already looks good, although it seems you still have some old comments lying around that need cleaning up.

Also, could you add docstrings to your definitions to explain what they do? We haven't really been great at documenting everything previously, but for new PRs we should strive to have a docstring for each new definition.

Finally, try to format your PR title as a proper sentence, so put spaces between words rather than dots and dashes, and use imperative style (read [mathlib's PR style guide](https://leanprover-community.github.io/contribute/commit.html) for more detail; although we aren't always as strict as the mathlib guideline, it'd be good to build a habit of following what they prescribe in that document)

SSA/Core/Framework.lean
salinhkuhn: I experimented with different ways of naming variables—specifically, I tried  naming free variables with arg0 to argn, and then assigning new local variables with purely numeric names. However, this approach caused issues: when one of the free variables needed to be printed, it ended up being assigned a numeric name as well, which led to confusion and unexpected behavior. Overall, I explored various naming strategies, but they didn't work out as intended

SSA/Core/Framework.lean
salinhkuhn: Whenever I rename the function to just toString, the calls to {toString op} and {toString outTy} result in errors. Lean tries to resolve toString as the one defined for Expr, so when I pass something of type d.Op or d.Ty, it complains because it expects an Expr. However, when I change the function name to something else, like toString1, the issue disappears

salinhkuhn: resolved now.

SSA/Core/Util/ConcreteOrMVar.lean
salinhkuhn: done.

SSA/Projects/InstCombine/Base.lean
salinhkuhn: Done. I use the Repr instance for the unary operations when defining their toString. Since the Repr instance is derived and doesn't include the InstCombine prefix, the output is already clean and readable, making it suitable to reuse directly for toString.

SSA/Projects/InstCombine/Base.lean
salinhkuhn: removed it.

salinhkuhn:
To visualize what this PR does: 
The input file with the following content, will be printed to the command line by running lake exe opt "file-path". 

```
  ^bb0(%arg0: i32):
    %0 = "llvm.mlir.constant"() {value = 8 : i32} : () -> i32
    %1 = "llvm.mlir.constant"() {value = 31 : i32} : () -> i32
    %2 = "llvm.ashr"(%arg0, %1) : (i32, i32) -> i32
    %3 = "llvm.and"(%2, %0) : (i32, i32) -> i32
    %4 = "llvm.add"(%3, %2) : (i32, i32) -> i32
    "llvm.return"(%4) : (i32) -> ()
```
The output will be the following:
```
^entry(%0 : i32):
  %1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
  %2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
  %3 = "llvm.ashr" (%0, %2) : (i32, i32) → (i32)
  %4 = "llvm.and" (%3, %1) : (i32, i32) → (i32)
  %5 = "llvm.add" (%4, %3) : (i32, i32) → (i32)
 return %5 : (i32) → ()
```
Programs containing flags e.g 
```
  ^bb0(%arg0: i32):
    %0 = "llvm.mlir.constant"() {value = 8 : i32} : () -> i32
    %1 = "llvm.mlir.constant"() {value = 31 : i32} : () -> i32
    %2 = "llvm.or"(%0, %1) <{isDisjoint}> : (i32, i32) -> i32
    %3 = "llvm.sdiv"(%0, %2) <{isExact}> : (i32, i32) -> i32
    "llvm.return"(%3) : (i32) -> ()
```
will be displayed as follows :
```
^entry(%0 : i32):
   %1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
   %2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
   %3 = "llvm.or disjoint" (%1, %2) : (i32, i32) → (i32)
   %4 = "llvm.sdiv exact" (%1, %3) : (i32, i32) → (i32)
  return %4 : (i32) → ()
```
Additionally, one can pass the --verbose flag to get more detailed output, including the effectKind, which will be printed as follows (for the first example):
```
Flag `--verbose` was set.
⟨EffectKind.pure, ⟨i32, {
    ^entry(%0 : i32):
      %1 = InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 8 : () → (i32)
      %2 = InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 31 : () → (i32)
      %3 = InstCombine.MOp.binary
        (ConcreteOrMVar.concrete 32)
        (InstCombine.MOp.BinaryOp.ashr) (%0, %2) : (i32, i32) → (i32)
      %4 = InstCombine.MOp.binary
        (ConcreteOrMVar.concrete 32)
        (InstCombine.MOp.BinaryOp.and) (%3, %1) : (i32, i32) → (i32)
      %5 = InstCombine.MOp.binary
        (ConcreteOrMVar.concrete 32)
        (InstCombine.MOp.BinaryOp.add) (%4, %3) : (i32, i32) → (i32)
      return %5 : (i32) → ()
  }⟩⟩
```

alexkeizer:
> Finally, try to format your PR title as a proper sentence, so put spaces between words rather than dots and dashes, and use imperative style (read [mathlib's PR style guide](https://leanprover-community.github.io/contribute/commit.html) for more detail; although we aren't always as strict as the mathlib guideline, it'd be good to build a habit of following what they prescribe in that document)

@salinhkuhn did you have a look at the PR style guide I linked in my previous review?

alexkeizer:
> To visualize what this PR does: The input file with the following content, will be printed to the command line by running lake exe opt "file-path".
> 
> ```
>   ^bb0(%arg0: i32):
>     %0 = "llvm.mlir.constant"() {value = 8 : i32} : () -> i32
>     %1 = "llvm.mlir.constant"() {value = 31 : i32} : () -> i32
>     %2 = "llvm.ashr"(%arg0, %1) : (i32, i32) -> i32
>     %3 = "llvm.and"(%2, %0) : (i32, i32) -> i32
>     %4 = "llvm.add"(%3, %2) : (i32, i32) -> i32
>     "llvm.return"(%4) : (i32) -> ()
> ```
> 
> The output will be the following:
> 
> ```
> ^entry(%0 : i32):
>   %1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
>   %2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
>   %3 = "llvm.ashr" (%0, %2) : (i32, i32) → (i32)
>   %4 = "llvm.and" (%3, %1) : (i32, i32) → (i32)
>   %5 = "llvm.add" (%4, %3) : (i32, i32) → (i32)
>  return %5 : (i32) → ()
> ```
> 
> Programs containing flags e.g
> 
> ```
>   ^bb0(%arg0: i32):
>     %0 = "llvm.mlir.constant"() {value = 8 : i32} : () -> i32
>     %1 = "llvm.mlir.constant"() {value = 31 : i32} : () -> i32
>     %2 = "llvm.or"(%0, %1) <{isDisjoint}> : (i32, i32) -> i32
>     %3 = "llvm.sdiv"(%0, %2) <{isExact}> : (i32, i32) -> i32
>     "llvm.return"(%3) : (i32) -> ()
> ```
> 
> will be displayed as follows :
> 
> ```
> ^entry(%0 : i32):
>    %1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
>    %2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
>    %3 = "llvm.or disjoint" (%1, %2) : (i32, i32) → (i32)
>    %4 = "llvm.sdiv exact" (%1, %3) : (i32, i32) → (i32)
>   return %4 : (i32) → ()
> ```
> 
> Additionally, one can pass the --verbose flag to get more detailed output, including the effectKind, which will be printed as follows (for the first example):
> 
> ```
> Flag `--verbose` was set.
> ⟨EffectKind.pure, ⟨i32, {
>     ^entry(%0 : i32):
>       %1 = InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 8 : () → (i32)
>       %2 = InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 31 : () → (i32)
>       %3 = InstCombine.MOp.binary
>         (ConcreteOrMVar.concrete 32)
>         (InstCombine.MOp.BinaryOp.ashr) (%0, %2) : (i32, i32) → (i32)
>       %4 = InstCombine.MOp.binary
>         (ConcreteOrMVar.concrete 32)
>         (InstCombine.MOp.BinaryOp.and) (%3, %1) : (i32, i32) → (i32)
>       %5 = InstCombine.MOp.binary
>         (ConcreteOrMVar.concrete 32)
>         (InstCombine.MOp.BinaryOp.add) (%4, %3) : (i32, i32) → (i32)
>       return %5 : (i32) → ()
>   }⟩⟩
> ```

Thanks for writing this summary, it helps a lot!

The first thing that jumps out at me is the fact that 
```
"llvm.return"(%3) : (i32) -> ()
```
gets printed as 
```
return %5 : (i32) → ()
```

Since we seem to be printing using the generic syntax for operations, we should keep that convention and use the generic syntax for printing returns also. That is, the output should be the same as the input on those examples you gave (modulo the attributes).

By the way, when asking for a second round of review, it's also good to check your PR description again, and make sure it's still accurate and up-to-date. In particular, I believe we've moved away from the global Repr instance modifications, right? So please update the PR description to reflect that.
The PR description end up in the git history for this PR, so it's important that the description matches the actual changes that get merged at the end accurately!

SSA/Core/Framework.lean
```diff
@@ -270,6 +270,47 @@ instance : Repr (Lets d Γ eff t) := ⟨flip Lets.repr⟩
 
 end Repr
 
+/- # ToString instances for Com and Expr  -/
+section ToString
+variable {d} [DialectSignature d] [Repr d.Op] [Repr d.Ty][ToString d.Ty][ToString d.Op]
```
```suggestion
variable {d} [DialectSignature d] [Repr d.Op] [Repr d.Ty] [ToString d.Ty] [ToString d.Op]
```

Keep a space in between binders!

SSA/Core/Framework.lean
```diff
@@ -270,6 +270,47 @@ instance : Repr (Lets d Γ eff t) := ⟨flip Lets.repr⟩
...
+  "(" ++ String.intercalate ", " args ++ ")"
+/--
+Converts an expression to its string representation.
+Assumes that `toString` instances exist for both the dialect's operations (`d.Op`)
+and types (`d.Ty`). The output includes the operation name, argument list,
+their types, and the resulting output type.
+-/
+partial def Expr.toString [ToString d.Op] : Expr d Γ eff t → String
+  | Expr.mk (op : d.Op) _ _ args _regArgs =>
+    let outTy : d.Ty  := DialectSignature.outTy op
```
```suggestion
    let outTy : d.Ty := DialectSignature.outTy op
```

SSA/Core/Framework.lean
```diff
@@ -270,6 +270,47 @@ instance : Repr (Lets d Γ eff t) := ⟨flip Lets.repr⟩
...
+/-- This function recursivly converts the body of a `Com` into its string representation.
+Each bound variable is printed with its index and corresponding expression. -/
+partial def Com.ToStringBody : Com d Γ eff t → String
+  | .ret v => s!"return {_root_.repr v } : ({toString t}) → ()"
+  | .var e body =>
+    s!" %{_root_.repr <|(Γ.length)} = {Expr.toString e }" ++ "\n" ++
+    Com.ToStringBody body
+
+/- This function implements a toString instance for the type `Com`.  -/
+partial def Com.toString (_com : Com d Γ eff t) : String :=
```
```suggestion
/- `Com.toString` implements a toString instance for the type `Com`.  -/
partial def Com.toString (_com : Com d Γ eff t) : String :=
```

refer to the function being documented by name, instead of "this function" (check this in other defs also, please!)

SSA/Core/Framework.lean
```diff
@@ -270,6 +270,47 @@ instance : Repr (Lets d Γ eff t) := ⟨flip Lets.repr⟩
...
+   "{ \n"
+  ++ "^entry" ++  ((formatFormalArgListTupleStr Γ)) ++ ":" ++ "\n"
+  ++ (Com.ToStringBody _com) ++
```
```suggestion
  ++ (Com.ToStringBody com) ++
```
if you use the variable, it shouldn't have the underscore

SSA/Projects/InstCombine/Base.lean
```diff
@@ -115,6 +115,47 @@ inductive MOp (φ : Nat) : Type
   | const (w : Width φ) (val : ℤ) : MOp φ
 deriving Repr, DecidableEq, Inhabited, Lean.ToExpr
 
+def toStringWithoutFlags (op : MOp.BinaryOp) : String :=
```
alexkeizer: This is called `withoutFlags`, but actually it does have the flags!

I think this way of printing the attributes is good for now, so that at least they are present, but indeed in future we would remove those attributes/flags from this function once we have proper attribute support (so that this function really becomes "withoutFlags". In the present, though, the name is confusing. Either we change the name now (and change it back to withoutFlags in the future), or we keep it as-is, but then you have to add a comment (preferably as a docstring) explaining that *at the moment* it does include the flags, but this will be changed in the future.

SSA/Projects/InstCombine/LLVM/Opt.lean
```diff
@@ -1,19 +1,38 @@
...
 
+
+def verbose_flag (fileName : String ) : IO UInt32 := do
+      let icom? ← parseComFromFile fileName
+      match icom? with
+      | none => return 1
+      | some (Sigma.mk _Γ icom) => do
+      IO.println "Flag `--verbose` was set."
+      IO.println s!"{repr icom}" -- we print everything, meaning effect, return type and com
+      return 0
```
alexkeizer, line 16:
```suggestion
def verboseFlag (fileName : String ) : IO UInt32 := do
      let icom? ← parseComFromFile fileName
      match icom? with
      | none => return 1
      | some (Sigma.mk _Γ icom) => do
      IO.println "Flag `--verbose` was set."
      IO.println s!"{repr icom}" -- we print everything, meaning effect, return type and com
      return 0
```

Check the [naming convention](https://leanprover-community.github.io/contribute/naming.html): regular definitions like this are `lowerCamelCase`.

Also, please add a docstring, in particular, to explain (a) that this function will print stuff (as that's not obvious from the name), and (b) what the meaning of the returned u32 is (I presume it's meant as an exit code?; if so, no need to document each value, but do document that it *is* an exit code and that 0 means success)

SSA/Projects/InstCombine/LLVM/Opt.lean
```diff
@@ -1,19 +1,38 @@
...
+    return code
+  else
+    let code ← wellformed fileName
+    return code
 
 def mainCmd := `[Cli|
     opt VIA runMainCmd;
     "opt: apply verified rewrites"
+    FLAGS:
+      verbose; "Declares a flag `--verbose`. This is the description of the flag."
```
alexkeizer: I think you forgot to replace the example string with an actual description here?

alexkeizer (approved):
Some last comments, but once those (and the two comments preceding this top-level one!) are addressed feel free to merge. Do remember to write docstrings for new definitions!

salinhkuhn:
> > To visualize what this PR does: The input file with the following content, will be printed to the command line by running lake exe opt "file-path".
> > ```
> >   ^bb0(%arg0: i32):
> >     %0 = "llvm.mlir.constant"() {value = 8 : i32} : () -> i32
> >     %1 = "llvm.mlir.constant"() {value = 31 : i32} : () -> i32
> >     %2 = "llvm.ashr"(%arg0, %1) : (i32, i32) -> i32
> >     %3 = "llvm.and"(%2, %0) : (i32, i32) -> i32
> >     %4 = "llvm.add"(%3, %2) : (i32, i32) -> i32
> >     "llvm.return"(%4) : (i32) -> ()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The output will be the following:
> > ```
> > ^entry(%0 : i32):
> >   %1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
> >   %2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
> >   %3 = "llvm.ashr" (%0, %2) : (i32, i32) → (i32)
> >   %4 = "llvm.and" (%3, %1) : (i32, i32) → (i32)
> >   %5 = "llvm.add" (%4, %3) : (i32, i32) → (i32)
> >  return %5 : (i32) → ()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Programs containing flags e.g
> > ```
> >   ^bb0(%arg0: i32):
> >     %0 = "llvm.mlir.constant"() {value = 8 : i32} : () -> i32
> >     %1 = "llvm.mlir.constant"() {value = 31 : i32} : () -> i32
> >     %2 = "llvm.or"(%0, %1) <{isDisjoint}> : (i32, i32) -> i32
> >     %3 = "llvm.sdiv"(%0, %2) <{isExact}> : (i32, i32) -> i32
> >     "llvm.return"(%3) : (i32) -> ()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > will be displayed as follows :
> > ```
> > ^entry(%0 : i32):
> >    %1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
> >    %2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
> >    %3 = "llvm.or disjoint" (%1, %2) : (i32, i32) → (i32)
> >    %4 = "llvm.sdiv exact" (%1, %3) : (i32, i32) → (i32)
> >   return %4 : (i32) → ()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Additionally, one can pass the --verbose flag to get more detailed output, including the effectKind, which will be printed as follows (for the first example):
> > ```
> > Flag `--verbose` was set.
> > ⟨EffectKind.pure, ⟨i32, {
> >     ^entry(%0 : i32):
> >       %1 = InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 8 : () → (i32)
> >       %2 = InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 31 : () → (i32)
> >       %3 = InstCombine.MOp.binary
> >         (ConcreteOrMVar.concrete 32)
> >         (InstCombine.MOp.BinaryOp.ashr) (%0, %2) : (i32, i32) → (i32)
> >       %4 = InstCombine.MOp.binary
> >         (ConcreteOrMVar.concrete 32)
> >         (InstCombine.MOp.BinaryOp.and) (%3, %1) : (i32, i32) → (i32)
> >       %5 = InstCombine.MOp.binary
> >         (ConcreteOrMVar.concrete 32)
> >         (InstCombine.MOp.BinaryOp.add) (%4, %3) : (i32, i32) → (i32)
> >       return %5 : (i32) → ()
> >   }⟩⟩
> > ```
> 
> Thanks for writing this summary, it helps a lot!
> 
> The first thing that jumps out at me is the fact that
> 
> ```
> "llvm.return"(%3) : (i32) -> ()
> ```
> 
> gets printed as
> 
> ```
> return %5 : (i32) → ()
> ```
> 
> Since we seem to be printing using the generic syntax for operations, we should keep that convention and use the generic syntax for printing returns also. That is, the output should be the same as the input on those examples you gave (modulo the attributes).
> 
> By the way, when asking for a second round of review, it's also good to check your PR description again, and make sure it's still accurate and up-to-date. In particular, I believe we've moved away from the global Repr instance modifications, right? So please update the PR description to reflect that. The PR description end up in the git history for this PR, so it's important that the description matches the actual changes that get merged at the end accurately!

I tried to experiment with preserving the input and printing exactly the input to the command line again but did not see a way to properly do that. Especially preserving the name of the free variables seemed hard. I can try to improve that in the future.
Your point regarding the return statement is a good one, but to implement that I would need a toString instance for every dialect. But since the dialect is not an inductive type but a structure, this change requires a bigger change in the framework, I suppose e.g the dialect structure could have a field of type String that asks for a name. 

update: we decided to implement to proper return statement printing and name preservation in a separate PR in the future.

SSA/Projects/InstCombine/LLVM/Opt.lean
salinhkuhn: I implemented the change and renamed the function to toStringWithFlags, so whenever we add attribute in the future we can change it.

SSA/Projects/InstCombine/Base.lean
salinhkuhn: I implemented the change and renamed the function to toStringWithFlags, so whenever we add attribute support in the future we can change it.

SSA/Projects/InstCombine/LLVM/Opt.lean
salinhkuhn: done

tobiasgrosser:
> > > To visualize what this PR does: The input file with the following content, will be printed to the command line by running lake exe opt "file-path".
> > > ```
> > >   ^bb0(%arg0: i32):
> > >     %0 = "llvm.mlir.constant"() {value = 8 : i32} : () -> i32
> > >     %1 = "llvm.mlir.constant"() {value = 31 : i32} : () -> i32
> > >     %2 = "llvm.ashr"(%arg0, %1) : (i32, i32) -> i32
> > >     %3 = "llvm.and"(%2, %0) : (i32, i32) -> i32
> > >     %4 = "llvm.add"(%3, %2) : (i32, i32) -> i32
> > >     "llvm.return"(%4) : (i32) -> ()
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The output will be the following:
> > > ```
> > > ^entry(%0 : i32):
> > >   %1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
> > >   %2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
> > >   %3 = "llvm.ashr" (%0, %2) : (i32, i32) → (i32)
> > >   %4 = "llvm.and" (%3, %1) : (i32, i32) → (i32)
> > >   %5 = "llvm.add" (%4, %3) : (i32, i32) → (i32)
> > >  return %5 : (i32) → ()
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Programs containing flags e.g
> > > ```
> > >   ^bb0(%arg0: i32):
> > >     %0 = "llvm.mlir.constant"() {value = 8 : i32} : () -> i32
> > >     %1 = "llvm.mlir.constant"() {value = 31 : i32} : () -> i32
> > >     %2 = "llvm.or"(%0, %1) <{isDisjoint}> : (i32, i32) -> i32
> > >     %3 = "llvm.sdiv"(%0, %2) <{isExact}> : (i32, i32) -> i32
> > >     "llvm.return"(%3) : (i32) -> ()
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > will be displayed as follows :
> > > ```
> > > ^entry(%0 : i32):
> > >    %1 = "llvm.mlir.constant" { value = 8 : i32 } : () → (i32)
> > >    %2 = "llvm.mlir.constant" { value = 31 : i32 } : () → (i32)
> > >    %3 = "llvm.or disjoint" (%1, %2) : (i32, i32) → (i32)
> > >    %4 = "llvm.sdiv exact" (%1, %3) : (i32, i32) → (i32)
> > >   return %4 : (i32) → ()
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Additionally, one can pass the --verbose flag to get more detailed output, including the effectKind, which will be printed as follows (for the first example):
> > > ```
> > > Flag `--verbose` was set.
> > > ⟨EffectKind.pure, ⟨i32, {
> > >     ^entry(%0 : i32):
> > >       %1 = InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 8 : () → (i32)
> > >       %2 = InstCombine.MOp.const (ConcreteOrMVar.concrete 32) 31 : () → (i32)
> > >       %3 = InstCombine.MOp.binary
> > >         (ConcreteOrMVar.concrete 32)
> > >         (InstCombine.MOp.BinaryOp.ashr) (%0, %2) : (i32, i32) → (i32)
> > >       %4 = InstCombine.MOp.binary
> > >         (ConcreteOrMVar.concrete 32)
> > >         (InstCombine.MOp.BinaryOp.and) (%3, %1) : (i32, i32) → (i32)
> > >       %5 = InstCombine.MOp.binary
> > >         (ConcreteOrMVar.concrete 32)
> > >         (InstCombine.MOp.BinaryOp.add) (%4, %3) : (i32, i32) → (i32)
> > >       return %5 : (i32) → ()
> > >   }⟩⟩
> > > ```
> > 
> > 
> > Thanks for writing this summary, it helps a lot!
> > The first thing that jumps out at me is the fact that
> > ```
> > "llvm.return"(%3) : (i32) -> ()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > gets printed as
> > ```
> > return %5 : (i32) → ()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Since we seem to be printing using the generic syntax for operations, we should keep that convention and use the generic syntax for printing returns also. That is, the output should be the same as the input on those examples you gave (modulo the attributes).
> > By the way, when asking for a second round of review, it's also good to check your PR description again, and make sure it's still accurate and up-to-date. In particular, I believe we've moved away from the global Repr instance modifications, right? So please update the PR description to reflect that. The PR description end up in the git history for this PR, so it's important that the description matches the actual changes that get merged at the end accurately!
> 
> I tried to experiment with preserving the input and printing exactly the input to the command line again but did not see a way to properly do that. Especially preserving the name of the free variables seemed hard. I can try to improve that in the future. Your point regarding the return statement is a good one, but to implement that I would need a toString instance for every dialect. But since the dialect is not an inductive type but a structure, this change requires a bigger change in the framework, I suppose e.g the dialect structure could have a field of type String that asks for a name.

I think that's fine for now and we may want to keep it the way it is in this PR. There is generally no expectation that names are preserved. We can refine the return statement as it becomes a problem.

alexkeizer:
> Especially preserving the name of the free variables seemed hard. I can try to improve that in the future.

Sorry, I should have been more precise there, as @tobiasgrosser mentioned, the names of the variables aren't important

tobiasgrosser:
> > Finally, try to format your PR title as a proper sentence, so put spaces between words rather than dots and dashes, and use imperative style (read [mathlib's PR style guide](https://leanprover-community.github.io/contribute/commit.html) for more detail; although we aren't always as strict as the mathlib guideline, it'd be good to build a habit of following what they prescribe in that document)
> 
> @salinhkuhn did you have a look at the PR style guide I linked in my previous review?

Sarah, did you look at this. In particular the pr title seems not yet compliant.

salinhkuhn:
> > > Finally, try to format your PR title as a proper sentence, so put spaces between words rather than dots and dashes, and use imperative style (read [mathlib's PR style guide](https://leanprover-community.github.io/contribute/commit.html) for more detail; although we aren't always as strict as the mathlib guideline, it'd be good to build a habit of following what they prescribe in that document)
> > 
> > 
> > @salinhkuhn did you have a look at the PR style guide I linked in my previous review?
> 
> Sarah, did you look at this. In particular the pr title seems not yet compliant.

I'm just now reading it and taking notes. I'm doing a PR learning session now for the next hour such that future PR's (there will be quite a few thinking of my instructions lowering) are better quality and easier to review.

Not sure if that's needed except for me but I could also write a few points about it into the knowledge base.

PR #1217, opened 2025-05-14
fix: remove transform error type dependency

salinhkuhn:
This PR removes the type dependency from TransformError. Instead of taking an input argument of type d.Ty, TransformError now takes no arguments. This change was necessary to enable staged parsing, as previously each dialect had its own ExceptM monad. By removing the dependency from TransformError, it's no longer required for each dialect to throw exceptions using its own monad.
To support this change, the PR adds ToString instances to existing projects such as the FHE and the PaperExample/Simple dialects.

### DIFF
--- a/SSA/Projects/InstCombine/LLVM/EDSL.lean
+++ b/SSA/Projects/InstCombine/LLVM/EDSL.lean
@@ -24,7 +24,7 @@ instance instTransformTy : AST.TransformTy (MetaLLVM φ) φ := { mkTy }
 instance : AST.TransformTy (LLVM) 0 := { mkTy }
 
 def getOutputWidth (opStx : MLIR.AST.Op φ) (op : String) :
-    Except (TransformError) (Width φ) := do
+    Except TransformError (Width φ) := do
   match opStx.res with
   | res::[] =>
     match res.2 with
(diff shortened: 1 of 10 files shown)


SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -24,7 +24,7 @@ instance instTransformTy : AST.TransformTy (MetaLLVM φ) φ := { mkTy }
 instance : AST.TransformTy (LLVM) 0 := { mkTy }
 
 def getOutputWidth (opStx : MLIR.AST.Op φ) (op : String) :
-    Except (TransformError (MetaLLVM φ).Ty) (Width φ) := do
+    Except (TransformError) (Width φ) := do
```
alexkeizer:
```suggestion
    Except TransformError (Width φ) := do
```
let's make it even prettier by removing these unnecessary parentheses

alexkeizer (approved):
Beautiful! I've got a minor suggestion, once that is accepted feel free to merge.

I've also ever so slightly re-worded your PR description: it's just a convention that we describe the PR as making the changes, rather than us authors, so "This PR changes ..." rather than "With this PR, I change ...". This is because the PR description ends up in the git history, and this style just tends to read a bit nicer.

PR #1218, opened 2025-05-14
refactor: replace alive_auto with bv_auto in AliveStatements

alexkeizer:
alive_auto is a tactic that runs the whole stack of `simp_alive_{ops, undef, etc.}` simplification tactics before finally calling out to `bv_auto` (the actual bitvec level solver tactic). Since we already run through the simplification stack of tactics in AliveStatements, the latter is thus much more appropriate.

### DIFF
--- a/SSA/Projects/InstCombine/AliveStatements.lean
+++ b/SSA/Projects/InstCombine/AliveStatements.lean
@@ -15,8 +15,10 @@ theorem bv_AddSub_1043 :
   simp_alive_ops
   simp_alive_case_bash
   simp_alive_split
-  try bv_auto
-  all_goals sorry
+  all_goals
+    solve
+    | bv_auto
+    | sorry
 
 theorem bv_AddSub_1152 :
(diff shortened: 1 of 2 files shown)


tobiasgrosser:
This reports 11 failed theorems, vs `3` the we previously had. Also,`SSA/Projects/InstCombine/update_alive_statements.py` must be updated.

alexkeizer:
Ah, as it turns out there are multiple test cases that do in fact generate multiple subgoals to be discharged by the solver, so my fear that our case bashing is fragile was entirely justified.

This was not surfacing before, as `alive_auto` implicitly applied `bv_auto` to all subgoals, so I've fixed it by hoisting up the all_goals in the generated proofs, so that
```
try bv_auto
all_goals sorry
```
is replaced with
```
all_goals
  solve
  | bv_auto
  | sorry
```

This has fixed the issue locally, leaving only the 3 sorries we'd excpect

alexkeizer:
> Also,SSA/Projects/InstCombine/update_alive_statements.py must be updated.

I already updated the script in my first commit, is there anything missing with the changes I made?

tobiasgrosser:
Thank you, @alexkeizer.

PR #1219, opened 2025-05-14
 feat: hybrid RiscV plus LLVM dialect [2/3]

salinhkuhn:
This PR introduces the parsing infrastructure for the RISC-V–LLVM hybrid dialect.
It performs staged parsing and leverages the existing parsers from the RISC-V and LLVM/InstCombine dialects


SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,132 @@ attribute [simp, simp_denote] outTy_map_signature_eq
 attribute [simp, simp_denote] HVector.mapM'
 attribute [simp, simp_denote] HVector.map'
 
+@[simp_denote]
+def transformExprLLVM (e : Expr (InstCombine.MetaLLVM 0) (ctxtTransformToLLVM Γ) eff ty) :
+  MLIR.AST.ReaderM (LLVMPlusRiscV) (Expr LLVMPlusRiscV Γ eff (.llvm ty)) :=
```
alexkeizer, line 139: Can you write a short docstring (i.e., `/-- ... -/` comment) that explains what this functions does? Idem for the transformExprRiscV function

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,132 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+              | .riscv _ =>
+                throw <| .generic s!"INTERNAL ERROR: This case is impossible, LLVM expression is pointing to RISCV variable."
+              | .llvm originalLLVMTy =>
+                if hty' : originalLLVMTy = t then
+                  return ⟨v.val, by rw [h, hty']⟩
+                else
+                  throw <|.generic s!"INTERNAL ERROR: This case is impossible, LLVM expression is pointing to an incorrect bitwidth LLVM argument."
+            | none =>
+              -- this is impossible, because ctxtTransformToLLVM is a `List.map`, which always maintains length.
+              throw <| .generic s!"INTERNAL ERROR: This case is impossible, as 'ctxtTransformToLLVM' is length-preserving."
```
alexkeizer: If this is really impossible, can we convince Lean of that fact? For example, by using `get` instead of `get?` on line 137?

If not (and that's fine, don't waste too much time convincing Lean if it's not straightforward), could you add to this comment why it's hard/annoying to prove this impossibility?

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,132 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+      let ⟨ty₁, v₁⟩ ← MLIR.AST.TypedSSAVal.mkVal Γ v₁Stx
+      match ty₁ with
+      | .riscv (.bv) =>
+              return ⟨ .pure,  .llvm (.bitvec 64) ,⟨ .castRiscv , by rfl, by constructor,
+               .cons v₁ <| .nil,
+                .nil⟩⟩
+      | .llvm (.bitvec 64) =>
+              return ⟨ .pure, .riscv (.bv) ,⟨ .castLLVM , by rfl, by constructor,
+               .cons v₁ <| .nil,
+                .nil⟩⟩
```
alexkeizer: Please have a look at the utilities I added in #1208, and how I used them to clean up the LLVM parser. You should be able to similarly clean up this branch of the `if` to just a few lines of code, using the `mkExprOf`, `parseArgs` and `assumeArity` helpers.

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,132 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+      | _ => throw <| .unsupportedOp s!"unsupported operation {repr opStx}"
+    | _ => throw <| .unsupportedOp s!"unsupported operation {repr opStx}"
```
alexkeizer: Note that unsupportedOp is not the right error here: this would claim that `builtin.unrealized_conversion_cast` is an unsupported operation, even though it most definitely is supported. I believe the actual error is that the number of arguments (i.e, the arity) is wrong, right? So throw an error that says that! (Although, if you follow my last comment and use `assumeArity`, you won't even have to throw the error yourself, the helper should throw the right error for you!)

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,132 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+        let ⟨eff, ty , expr⟩ ← RiscvMkExpr.mkExpr (ctxtTransformToRiscV Γ) opStx (← read)
+        let v ← transformExprRISCV expr
+        return ⟨eff, .riscv ty , v⟩
+
+instance : MLIR.AST.TransformExpr (LLVMPlusRiscV) 0   where
+  mkExpr := mkExpr
+
+-- TO DO: finish this proof. Did not really know how to do yet. But convinced it is provable.
+@[simp_denote]
+def transformVarLLVM (v :  Ctxt.Var (ctxtTransformToLLVM Γ) ty) :   Ctxt.Var Γ (LLVMRiscV.Ty.llvm ty) :=
```
alexkeizer:
```suggestion
def transformVarLLVM (v :  Ctxt.Var (ctxtTransformToLLVM Γ) ty) : Ctxt.Var Γ (LLVMRiscV.Ty.llvm ty) :=
```
Again, please check whitespace throughout the file, you only want to have a single space and after `:`.

In future, please double check whitespace of your PR before asking for a review, please!

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,132 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+  | ⟨h, ty⟩ =>  ⟨h, by sorry ⟩
+
+@[simp_denote]
+def transformVarRISCV (v :  Ctxt.Var (ctxtTransformToRiscV Γ) ty) :   Ctxt.Var Γ (LLVMRiscV.Ty.riscv ty) :=
+  match v with
+  | ⟨h, ty⟩ =>  ⟨h, sorry ⟩
+
+def mkReturn (Γ : Ctxt _) (opStx : MLIR.AST.Op 0) : MLIR.AST.ReaderM (LLVMPlusRiscV)
+    (Σ eff ty, Com LLVMPlusRiscV Γ eff ty) := do
+    let llvmParseReturn := InstcombineTransformDialect.mkReturn (ctxtTransformToLLVM  Γ) opStx (← read)
```
alexkeizer:
```suggestion
  let llvmParseReturn := InstcombineTransformDialect.mkReturn (ctxtTransformToLLVM  Γ) opStx (← read)
```
and the rest of the body should be de-dented one level also.

By the way, if you want a thorough reference of all the style rules, we generally follow the convention set by Mathlib, which you can read here: https://leanprover-community.github.io/contribute/index.html. It's quite detailed, but do have a look at the style guide, naming conventions, and documentation guidelines linked at the start of that page. Don't worry too much about the PR / communication workflow that document describes further down, as our workflow doesn't follow that exactly, although they do have some good examples of what [a PR description](https://leanprover-community.github.io/contribute/commit.html) could look like!

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,132 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+    match llvmParseReturn with
+    | .ok ⟨eff, ty, Com.ret v⟩ =>
+      return ⟨eff, .llvm ty, Com.ret (transformVarLLVM v)⟩
+    | _ =>
+      let ⟨eff, ty , com⟩ ← RiscvMkExpr.mkReturn  (ctxtTransformToRiscV Γ) opStx (← read)
+      match com with
+      |Com.ret v =>
```
alexkeizer:
```suggestion
      | Com.ret v =>
```
We always put a whitespace after the `|` of a match-arm (check this everywhere, please)!

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,132 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+  else
+    let llvmParse := InstcombineTransformDialect.mkExpr (ctxtTransformToLLVM  Γ) opStx (← read)
+    match llvmParse with
+      | .ok ⟨eff, ty, expr⟩ => do
+        let v ← transformExprLLVM expr
+        return ⟨eff, .llvm ty, v⟩
+      | .error  (_) => do
+        let ⟨eff, ty , expr⟩ ← RiscvMkExpr.mkExpr (ctxtTransformToRiscV Γ) opStx (← read)
+        let v ← transformExprRISCV expr
+        return ⟨eff, .riscv ty , v⟩
```
alexkeizer, line 270: We should only attempt to run the RiscV parser if the LLVM parser declares it doesn't feel responsible for the operation (which it does by throwing the unsupportedOp error). Any other errors should be passed to the user right away. 

This is partly a performance thing, but most importantly, it ensures the errors remain relevant. In your version, you suppress the information present in the LLVM parse error, so I believe that if you, e.g., ask it to parse an `llvm.add` which mistakenly has three arguments, the llvm parser will rightfully throw an error that it expected exactly two arguments, but you then call the riscv parser and return whatever error that gave, meaning the user only sees a misleading eror claiming that llvm.add is an unsupported operation.

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,132 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+  mkTy tStx := do
+    try
+      let llvmParse ← InstcombineTransformDialect.mkTy tStx
+      return .llvm llvmParse
+    catch _llvmErr =>
+      try
+        let riscvParse ← RiscvMkExpr.mkTy tStx
+        return .riscv riscvParse
+      catch _riscvErr =>
+        throw <| .generic s!"Unable to parse type as either LLVM type or RISCV type."
```
alexkeizer: You might want to include the information present in the original llvm and riscv errors. Generally the point of an error is to give information as to what might be wrong, so it's best not to ignore or throw away this information, but just to pass it to the user instead!

SSA/Projects/LLVMAndRiscv.lean
alexkeizer: I'm unfortunately less convinced, here's a proof of `False` using that definition:

```
example : False := by
  let Γ : Ctxt LLVMPlusRiscV.Ty := [.riscv .bv]
  let v : Ctxt.Var (ctxtTransformToLLVM Γ) (.bitvec 999) :=
    ⟨0, rfl⟩
  let v' : Ctxt.Var Γ (.llvm <| .bitvec 999) :=
    transformVarLLVM v

  simpa [Γ, v', v, transformVarLLVM] using v'.property
```

It cannot be true that any variable in the transformed context can be mapped onto an LLVM variable in the original context. Remember that in the transformation we've replaced any RiscV types with a placeholder `.bitvec 999`, meaning that if the variable with index `0` in the original context is a riscv type, then the variable with index `0` in the transformed type has type .bitvec 999. From just the existence of the variable in the transformed context, we should thus not be allowed to assume the same variable in the original context indeed has an LLVM type!

Thus, if we *do* want to do this transformation, but realize we cannot do so for all values, then the solution is that the function must be fallible, i.e., it can sometimes throw an error (or return an `Option.none` or something along those lines). Hopefully that also helps with how you'd implement this

alexkeizer (changes requested):
Unfortunately, the sorries are not actually provable. I left a more detailed comment, but you'll have to slightly change your strategy there.

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: done

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: done

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: done.

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: makes sense. Changed it, thank you

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: done

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: thank you, resolved now by making the function fallible.

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: I tried now for some time and I get to the point where I needed to prove that being a valid index in the transformed context is valid iff it is valid in the original context. That's where I then got stuck. I kept my proof attempt on the side. I do not think much is missing, maybe we can pair on it later once and add it with a follow up PR.

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: changed the error to be an arity error

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: I just looked at it and started implementing the change. I then got stuck and couldn't figure out how to use it for my cast that has the same name but should result in different expressions depending on the argument type. I now left the implementation as I had it initially and only changed the error message.
I tried to use the new utilities, but I did not now how to do it when I needed to pass the operand name to the mkExpression function — the problem being that the opStx name isn't unique in my case.

salinhkuhn: Update: I managed to use it. Happy to get feedback if it possible to use them in a even cleaner way. 
I also added the proof that the out-of-bounds-case is actually impossible , hence  changed from get? to get. 
Also fixed the sorries.

SSA/Core/Framework.lean
```diff
@@ -268,8 +268,24 @@ instance : Repr (Expr d Γ eff t) := ⟨flip Expr.repr⟩
 instance : Repr (Com d Γ eff t) := ⟨flip Com.repr⟩
 instance : Repr (Lets d Γ eff t) := ⟨flip Lets.repr⟩
 
+-- modified section for sarah
+variable {dd}[DialectSignature dd][ToString dd.Ty][ToString dd.Op]
+
+
+
+
```
tobiasgrosser: Why so many newlines?

SSA/Core/Framework.lean
```diff
@@ -268,8 +268,24 @@ instance : Repr (Expr d Γ eff t) := ⟨flip Expr.repr⟩
 instance : Repr (Com d Γ eff t) := ⟨flip Com.repr⟩
 instance : Repr (Lets d Γ eff t) := ⟨flip Lets.repr⟩
 
+-- modified section for sarah
```
tobiasgrosser: Why this comment?

SSA/Core/Framework.lean
```diff
@@ -268,8 +268,24 @@ instance : Repr (Expr d Γ eff t) := ⟨flip Expr.repr⟩
...
+variable {dd}[DialectSignature dd][ToString dd.Ty][ToString dd.Op]
+
+
+
+
+  /-- Format string for a Com, with the region parentheses and formal argument list. -/
+  partial def Com.toString(com : Com d Γ eff t) : String :=
+    "{\n hi is there a new line "
+
+
```
tobiasgrosser:
```suggestion

```

SSA/Core/Framework.lean
```diff
@@ -268,8 +268,24 @@ instance : Repr (Expr d Γ eff t) := ⟨flip Expr.repr⟩
...
+instance : ToString (Com dd Γ eff t)  := ⟨Com.toString⟩
+
 end Repr
 
+
+
+
 /-! ### DecidableEq instance -/
```
tobiasgrosser: This change is unrelated?

SSA/Projects/InstCombine/Base.lean
```diff
@@ -105,6 +110,34 @@ def reprWithoutFlags (op : MOp.BinaryOp) (prec : Nat) : Format :=
 instance : Repr (MOp.BinaryOp) where
   reprPrec := reprWithoutFlags
 
+-- to   DO : FORMATTING
```
tobiasgrosser: What is this 'to DO' about?

SSA/Projects/InstCombine/Base.lean
```diff
@@ -105,6 +110,34 @@ def reprWithoutFlags (op : MOp.BinaryOp) (prec : Nat) : Format :=
...
+  | .add  ⟨false, false⟩   => "llvm.add"
+  | .add  ⟨nsw, nuw⟩        => s!"llvm.add {nsw} {nuw}"
+  | .mul  ⟨false, false⟩ => "llvm.mul"
+  | .mul  ⟨nsw, nuw⟩       => s!"llvm.mul {nsw} {nuw}"
+  | .sub ⟨false, false⟩   => "llvm.sub"
+  | .sub ⟨nsw, nuw⟩        => s!"llvm.sub {nsw} {nuw}"
+  | .sdiv ⟨false⟩         => "llvm.sdiv"
+  | .sdiv ⟨true⟩           => "llvm.sdiv exact"
+  | .udiv ⟨false⟩         => "llvm.udiv"
+  | .udiv ⟨true⟩           => "llvm.udiv exact"
```
tobiasgrosser: The `=>` does not seem aligned. Similarly, the whitespace after the operation name and the flags is inconsistent.

SSA/Projects/InstCombine/Base.lean
```diff
@@ -115,6 +148,15 @@ inductive MOp (φ : Nat) : Type
...
 
+instance : ToString (MOp 0) where
+   toString
+     | .unary _w op => s!"\"{repr  op}\""
+     | .binary _w op => s!"\"{repr  op}\""
+     | .select  _w =>  "select"
+     | .icmp  _pred _w => "icmp"
+     | .const w val => s!"\"llvm.mlir.constant\" \{ value = {val} : {w} }"
+
+
```
tobiasgrosser:
```suggestion

```

SSA/Projects/InstCombine/Base.lean
```diff
@@ -353,6 +395,7 @@ instance : Lean.ToFormat LLVM.Ty := by unfold LLVM; infer_instance
 
 instance : ToString (MTy φ) where
   toString t := repr t |>.pretty
+
 instance : ToString LLVM.Ty := by unfold LLVM; infer_instance
```
tobiasgrosser: Unrelated.

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -16,7 +18,7 @@ currently only operates within one dialect.
 
 To make the intermixing of the type system across the dialects work,
 we insert unrealized_conversion_cast like MLIR does during lowering.
-see: section (UnrealizedConversionCastOp)
+see_: section (UnrealizedConversionCastOp)
```
tobiasgrosser: Why an underscore here?

SSA/Projects/RISCV64/Base.lean
```diff
@@ -221,7 +221,7 @@ Again, we mark  it as `simp` and `reducible`.
 -/
 @[simp, reducible]
 def Op.outTy : Op  → Ty
-  |.li _ => Ty.bv
+  | .li _ => Ty.bv
```
tobiasgrosser: This change is unrelated.

SSA/Projects/RISCV64/Base.lean
```diff
@@ -145,7 +145,7 @@ encounters a `sig` it can replace it by its definition.
 -/
 @[simp, reducible]
 def Op.sig : Op → List Ty
-  |.li _ => []
+  | .li _ => []
```
tobiasgrosser: This change is unrelated.

tobiasgrosser (commented):
Hi @salinhkuhn, I pushed some minor style comments. After these have been addressed, I feel we may only need a final comment from @alexkeizer.

Besides addressing these comments, can you please write 1-2 sentences of where this PR is standing and what you would need from @alexkeizer review. It is then clear that you are waiting for @alexkeizer's input to move this forward.

SSA/Projects/InstCombine/Base.lean
salinhkuhn: They should not be part of this PR, I removed it.

SSA/Projects/InstCombine/Base.lean
```diff
@@ -353,6 +353,7 @@ instance : Lean.ToFormat LLVM.Ty := by unfold LLVM; infer_instance
 
 instance : ToString (MTy φ) where
   toString t := repr t |>.pretty
+
```
tobiasgrosser:
```suggestion
```

tobiasgrosser: There is still one unrelated change.

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,184 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+  | .error e =>
+    match e with
+    | .unsupportedOp _s=>
+      let ⟨eff, ty , com⟩ ← RiscvMkExpr.mkReturn (ctxtTransformToRiscV Γ) opStx (← read)
+      match com with
+      | Com.ret v =>
+        return ⟨eff, .riscv ty, Com.ret (← transformVarRISCV v)⟩
+      | _ => throw <| .generic s!"Unable to parse return as either LLVM type or RISCV type."
+    | _ => throw <| .generic s!"Unable to parse return as either LLVM type or RISCV type.
+        LLVM parser threw error diffrent from unsupported Op"
```
tobiasgrosser:
```suggestion
        LLVM parser threw error different from unsupported Op"
```

SSA/Projects/InstCombine/Base.lean
salinhkuhn: this was a whitespace error in InstCombine that I wanted to fix, but yes will pull it out into a single PR.

tobiasgrosser: For such whitespace errors, just open a separate PR and merge it "as obvious" without waiting for a review.

salinhkuhn:
Summary for @alexkeizer   regarding hybrid dialect 2 PR.
- The sorry part is now proven, as we discussed :)
- You had previously commented and asked whether it's possible to prove that the out-of-bounds case when accessing the transformed error won't happen. I looked into it yesterday and managed to prove it. The proof is a bit longer than just a few lines, but I minimized it as much as I could. I'd be curious to get your review on that part.
-  For now, the proof is included only for the RISC-V transform context. Once it's approved, I’ll add the analogous proof for LLVM. It's the same logic.
- In the end I also managed to use the new parsing utilities you implemented. A quick review on whether I used them optimally would be appreciated.
-  All other comments regarding the naming of the errors have been addressed as suggested.

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -224,31 +268,39 @@ def mkExpr (Γ : Ctxt _) (opStx : MLIR.AST.Op 0) :
 instance : MLIR.AST.TransformExpr (LLVMPlusRiscV) 0   where
   mkExpr := mkExpr
 
--- TO DO: finish this proof. Did not really know how to do yet. But convinced it is provable.
 @[simp_denote]
-def transformVarLLVM (v :  Ctxt.Var (ctxtTransformToLLVM Γ) ty) :   Ctxt.Var Γ (LLVMRiscV.Ty.llvm ty) :=
-  match v with
-  | ⟨h, ty⟩ =>  ⟨h, by sorry ⟩
+def transformVarLLVM (v : Ctxt.Var (ctxtTransformToLLVM Γ) ty) : MLIR.AST.ReaderM LLVMPlusRiscV (Ctxt.Var Γ (LLVMRiscV.Ty.llvm ty)) :=
```
alexkeizer: line length

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -224,31 +268,39 @@ def mkExpr (Γ : Ctxt _) (opStx : MLIR.AST.Op 0) :
...
+      let ⟨eff, ty , com⟩ ← RiscvMkExpr.mkReturn (ctxtTransformToRiscV Γ) opStx (← read)
       match com with
-      |Com.ret v =>
-        return ⟨eff, .riscv ty, Com.ret (transformVarRISCV v) ⟩
-      |_ => throw <| .generic s!"Unable to parse return as either LLVM type or RISCV type."
+      | Com.ret v =>
+        return ⟨eff, .riscv ty, Com.ret (← transformVarRISCV v)⟩
+      | _ => throw <| .generic s!"Unable to parse return as either LLVM type or RISCV type."
+    | _ => throw <| .generic s!"Unable to parse return as either LLVM type or RISCV type.
+        LLVM parser threw error different from unsupported Op"
```
alexkeizer: Here, too, you want to pass along the information of the original error!
```suggestion
    | e => throw e
```

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -224,31 +268,39 @@ def mkExpr (Γ : Ctxt _) (opStx : MLIR.AST.Op 0) :
...
+    match e with
+    | .unsupportedOp _s=>
+      let ⟨eff, ty , com⟩ ← RiscvMkExpr.mkReturn (ctxtTransformToRiscV Γ) opStx (← read)
       match com with
-      |Com.ret v =>
-        return ⟨eff, .riscv ty, Com.ret (transformVarRISCV v) ⟩
-      |_ => throw <| .generic s!"Unable to parse return as either LLVM type or RISCV type."
+      | Com.ret v =>
+        return ⟨eff, .riscv ty, Com.ret (← transformVarRISCV v)⟩
+      | _ => throw <| .generic s!"Unable to parse return as either LLVM type or RISCV type."
```
alexkeizer: What's the purpose of matching against the `com` here? Why not just return
`RiscvMkExpr.mkReturn (ctxtTransformToRiscV Γ) opStx` directly?

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -224,31 +268,39 @@ def mkExpr (Γ : Ctxt _) (opStx : MLIR.AST.Op 0) :
...
+    | _ => throw <| .generic s!"Unable to parse return as either LLVM type or RISCV type.
+        LLVM parser threw error different from unsupported Op"
+  | _ => throw <| .generic s!"Unable to parse return as either LLVM type or RISCV type."
```
alexkeizer: I'm not sure what this error case is exactly, but if both LLVM and RiscV give an unsupportedOp error, then your hybrid parser ought to give an unsupportedOp error, too!

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,184 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+  MLIR.AST.ReaderM (LLVMPlusRiscV) (Expr LLVMPlusRiscV Γ eff (.llvm ty)) :=
+    match e with
+    | Expr.mk op1 ty_eq1 eff_le1 args1 regArgs1 => do
+        let args' : HVector (Ctxt.Var Γ) (.llvm <$> DialectSignature.sig op1) ←
+          args1.mapM' fun t v => do
+            match h : Γ.get? v.val with
+            | some ty' => do
+              match hty : ty' with
+              | .riscv _ =>
+                throw <| .generic s!"INTERNAL ERROR: This case is impossible, LLVM expression is pointing to RISCV variable."
```
alexkeizer: Is this truly impossible / an internal error? What error do we get when a user writes:
````
^bb0(%a : !i64, %b : !i64):
  %v1 = llvm.and %a, %b -- using RiscV types in an LLVM operation!
````

This *should* throw an error, but it shouldn't be an internal error. Internal errors usually indicate a bug in the library/function that you're calling, but here the error is that the input to the function is mallformed. If there's something earlier that would've caught this, then please add a brief comment saying:
```
-- This is impossible, because mixing LLVM and RiscV variables would've already been caught by ...
```

File: SSA/Projects/LLVMAndRiscv.lean [hunk 28]
```diff
@@ -125,4 +127,184 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+    let llvmParse ← InstcombineTransformDialect.mkTy tStx
+    return .llvm llvmParse
+  catch _llvmErr =>
+    try
+      let riscvParse ← RiscvMkExpr.mkTy tStx
+      return .riscv riscvParse
+    catch _riscvErr =>
+        throw <|.generic s!" INTERNAL ERROR : While trying to transform from MLIR AST to dialect specific AST
+         the transformation failed. The errors thrown are:
+          s!{(toString (repr _riscvErr))} and s!{(toString (repr _llvmErr ))}"
```
alexkeizer:
```suggestion
  catch llvmErr =>
    try
      let riscvParse ← RiscvMkExpr.mkTy tStx
      return .riscv riscvParse
    catch riscvErr =>
        throw <|.generic s!" INTERNAL ERROR : While trying to transform from MLIR AST to dialect specific AST
         the transformation failed. The errors thrown are:
          s!{(toString (repr riscvErr))} and s!{(toString (repr llvmErr ))}"
```

If you use the variables, they shouldn't get a `_` before them!

Also, is this truly an INTERNAL ERROR? See the last comment, if this error can happen just because of bad input, it isn't an internal error, just a regular error.

alexkeizer (approved):
I've got some last comment about error handling (it is quite tricky). Once those are addressed, it should be fine to go ahead and merge!

SSA/Projects/LLVMAndRiscv.lean
```diff
@@ -125,4 +127,184 @@ attribute [simp, simp_denote] outTy_map_signature_eq
...
+  MLIR.AST.ReaderM (LLVMPlusRiscV) (Σ eff ty, Expr LLVMPlusRiscV Γ eff ty) := do
+  let args ← opStx.parseArgs Γ
+  if (opStx.name = "builtin.unrealized_conversion_cast") then
+    let mkExprOf := opStx.mkExprOf (args? := args) Γ
+    let args ← (← opStx.parseArgs Γ).assumeArity 1
+    let ⟨ty, v⟩ := args[0]
+    match ty with
+      | .riscv (.bv) => mkExprOf <| .castRiscv
+      | .llvm (.bitvec 64) => mkExprOf <| .castLLVM
+      | _ => throw <| .unsupportedOp s!"unsupported operation {repr opStx}"
```
alexkeizer, line 260: This looks beautiful, thanks for making the change!

One final suggestion, we can pass along the parsed expressions to `mkExprOf` to avoid having to parse them again, like so:
```suggestion
    let mkExprOf := opStx.mkExprOf (args? := args) Γ
    let args ← (← opStx.parseArgs Γ).assumeArity 1
    let ⟨ty, v⟩ := args[0]
    match ty with
      | .riscv (.bv) => mkExprOf (args? := args) <| .castRiscv
      | .llvm (.bitvec 64) => mkExprOf (args? := args) <| .castLLVM
      | _ => throw <| .unsupportedOp s!"unsupported operation {repr opStx}"
```
This is just a minor performance thing, to avoid doing the same work of parsing the arguments twice.

salinhkuhn: I already parse it on line 256. I therefore already avoid doing the parsing twice. I think my naming of the partial applied mkExprOf to again mkExprOf might be confusing.

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: I need to access the return variable and transform it back to the hybrid dialect context. Returning directly would give me an error I assume.

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: I added the comment.

SSA/Projects/LLVMAndRiscv.lean
salinhkuhn: changed it to unsupportedOp

SSA/Projects/LLVMAndRiscv.lean
alexkeizer: Ah, fair point. I'd argue "Unable to parse return as either LLVM type or RISCV type" is a bit of a misleading error when the parse succeeds (but just happens to return a `Com.var` for whatever reason), but since it shouldn't happen anyway I guess it's not that bad

SSA/Projects/LLVMAndRiscv.lean
alexkeizer: Ah, no, that's entirely my bad, given I did exactly that same trick in LLVM.

PR #1220, opened 2025-05-14
refactor: replace `simp_alive_split` implementation with just simp.

alexkeizer:
This PR changes the implementation of `simp_alive_split`---which used to call `split` and `split_ifs` and thus introduced new subgoals---with a single simp call (and some `intros` to preserve that behaviour). This is possible as the `if` statements used in the semantics follow a strict pattern, where one branch of the if is always `poison`. Such ifs with poison admit straightforward simplification lemmas, which this PR adds, thus removing the need to split on the condition.

NOTE: I've confirmed the new implementation of `simp_alive_split` successfully closes all the expected goals in`AliveStatements`, but we ought to wait for the full evaluation to run before merging this PR, as there could be breakage there.

Keep in mind that we do expect the numbers of solved theorems reported by the evaluation to go down, given that we are actively trying to reduce the number of subgoals generated by our automation (hence reducing the number of distinct calls to the SAT solvers).


alexkeizer:
This seems to work well enough, but to avoid causing problems with the evaluation I'll not replace the implementation of the existing tactic, but rather make this into a new tactic, which will be specialized for the new SLLVM dialect (which models proper immediate UB).

PR #1221, opened 2025-05-15
chore: disable unused variable error

tobiasgrosser:


SSA/Core/MLIRSyntax/Transform.lean
```diff
@@ -35,7 +35,7 @@ section Monads
   errors.
 -/
 
-abbrev ExceptM  (d : Dialect) := Except TransformError
+abbrev ExceptM  (_ : Dialect) := Except TransformError
 abbrev BuilderM (d : Dialect) := StateT NameMapping (ExceptM d)
 abbrev ReaderM  (d : Dialect) := ReaderT NameMapping (ExceptM d)
```
alexkeizer, line 40: @salinhkuhn Sorry, I should have caught this in the original PR: now that the transform error doesn't take a Ty argument anymore, we can also get rid of the `d : Dialect` argument to `ExceptM`, `BuilderM`, and `ReaderM`.

You should've gotten an unused variable warning in this file, that's why Tobias made this change here (by renaming the argument to `_` he explicitly indicated the variable will not be used, getting rid of the warning). Could you open a second PR where you remove the Dialect argument from these three monads alltogether?

PR #1211, opened 2025-05-12
To int to nat removal semantics: RISCV Merge [1/3]

salinhkuhn:
This is a rather big PR again but it contains the semantics already merged and additionally has rewrites of the semantics into purely bit vector domain statements without calls to toInt and toNat. This file should replace the current smenatics file when complete.

This PR is the first third of the whole undertaking.
For some theorems, we previously already committed their pure variant but left out the toNat version. In this case, we now add the toNat version under the original theorem name, move the pure version to a _bv statement, and prove equivalence of both.

PR #1224, opened 2025-05-15
To int to nat removal semantics: RISCV Merge [2/3] 

salinhkuhn:
This PR continues the migration of the semantics into the pure bit vector domain by adding additional translations and the corresponding correctness proofs. In particular, it is the follow-up PR to #1211 .


tobiasgrosser (approved):
LGTM. Feel free to merge.

PR #1227, opened 2025-05-15
feat: intros_llvm tactic

alexkeizer:
This PR adds an `intros_llvm` tactic which behaves similar to `intros`, in that it introduces as many binders as it can, but it does while giving them accessible names. The tactic is not strictly speaking specific to the LLVM dialect, but it's intended for use with the other alive/llvm tactics, hence the name.


tobiasgrosser:
Nice. LGTM

PR #1235, opened 2025-05-16
feat: add peephole rewrite structure for LLVMAndRiscV dialect

salinhkuhn:
This PR adds peephole rewrite structures for the LLVMAndRiscV hybrid dialect, enabling reuse of the existing Peephole Rewriter to perform rewrites within the hybrid dialect


SSA/Projects/LLVMRiscV/PeepholeRefine.lean
```diff
@@ -0,0 +1,66 @@
...
+
+/-- `RiscVPeepholeRewriteRefine` defines a `PeepholeRewrite` structure
+for RISC-V `Com`s within the hybrid dialect. The refinement relation
+is defined using bitvector refinement, since the return values of
+RISC-V computations are bitvectors. -/
+structure RiscVPeepholeRewriteRefine (Γ : Ctxt Ty) where
+  lhs : Com LLVMPlusRiscV Γ .pure (Ty.riscv (.bv))
+  rhs : Com LLVMPlusRiscV Γ .pure (Ty.riscv (.bv))
+  correct : ∀ V,
+    BitVec.Refinement (lhs.denote V : Option _) (rhs.denote V : Option _)
```
alexkeizer: This definition causes the `Option.some` and `liftm` stuff that we looked at a while back. Remembering that the denotation of a RiscV `.bv` is a plain `BitVec`, why do you insert the type hint `Option _` here? This hint makes Lean coerce your bitvec into an Option, which effectively means it's inserting an implicit `Option.some` here. 

Then, you ask for the `BitVec.Refinement` of those two `some`s, but this is just equivalent to stating that the original bitvecs were equal. That is, the stated `correct`ness criteria should be equivalent to `∀ V, lhs.denote V = rhs.denote V`. This, then, is just the correctness statement of regular PeepholeRewrites. Is there a specific reason you needed to add this structure?

SSA/Projects/LLVMRiscV/PeepholeRefine.lean
```diff
@@ -0,0 +1,66 @@
+import SSA.Projects.LLVMRiscV.LLVMAndRiscv
+
+open LLVMRiscV
+/-
+This file contains the defintions of the Peephole Rewrite
```
alexkeizer:
```suggestion
This file contains the definitions of the Peephole Rewrite
```
type

SSA/Projects/LLVMRiscV/PeepholeRefine.lean
```diff
@@ -0,0 +1,66 @@
...
+a `sorry`. To still leverage the rewrite functionality, we wrap
+our rewrites into a form accepted by the Peephole Rewriter, omitting
+the proof of return value equality. This does not compromise our
+correctness guarantees, since the rewrite itself includes a proof
+that, for any two programs of the correct type, the rewrite is a
+valid refinement. We still hint the unsoudness of the Peephole
+Rewrite because in fact the proof is not provided and can't be
+provided until the Peephole Rewriter accepts refinements.
+ -/
+def LLVMToRiscvPeepholeRewriteRefine.toPeepholeUNSOUND
```
alexkeizer: Please split this comment up into a broader comment on the file, which should be part of the `/-! ... -/` doc comment (the crucial difference between `/-! ... -/` and `/- ... -/` is that the former will show up when generating documentation in the style of https://leanprover-community.github.io/mathlib4_docs/index.html), and also add a more succinct docstring (`/-- ... -/`-type comment), which points out the unsoundness, for the specific definition

alexkeizer (approved):
I think we only need the LLVMPeephole, right? Once you've removed the RiscV one and refined the documentation, feel free to merge!

SSA/Projects/LLVMRiscV/PeepholeRefine.lean
salinhkuhn: Done.

SSA/Projects/LLVMRiscV/PeepholeRefine.lean
salinhkuhn: Removed this definition.

PR #1236, opened 2025-05-16
fix: change llvm parser to throw unsupported op instead of generic error within mkReturn

salinhkuhn:
This PR updates the LLVM parser to throw an "unsupported operation" error instead of a generic error within the mkReturn function. Previously, when encountering an unknown return statement or operand, it would throw a generic error. With this PR, it now throws a more specific "unsupported op" error.
This PR is part of a stagged PR chain, it is based on the PR #1235.


alexkeizer (changes requested):
A more elegant solution would be to change the LLVM parser so that `mkReturn` throws an unsupportedOp error. That way, we'd still preserve the nice error about the wrong number of variables when writing `llvm.return %a, %b`, I worry your change would suppress this error.

@salinhkuhn could you give that a try? If it turns out to cause too many other problems, though, don't waste too much on it, but do ask for help!

salinhkuhn:
> A more elegant solution would be to change the LLVM parser so that `mkReturn` throws an unsupportedOp error. That way, we'd still preserve the nice error about the wrong number of variables when writing `llvm.return %a, %b`, I worry your change would suppress this error.
> 
> @salinhkuhn could you give that a try? If it turns out to cause too many other problems, though, don't waste too much on it, but do ask for help!

It didn't cause any problems at all, so I went ahead and implemented the change. I also tested the parsing again to make sure the issue was resolved, and it is indeed resolved.
As soon as you give me the OK @alexkeizer , it is ready to merge.

tobiasgrosser:
This looks great. Thank you, @salinhkuhn.

PR #1237, opened 2025-05-17
chore: refactor `Comb` parser and remove unnecessary types

luisacicolini:
This PR refactors the `Comb` parser according to #1208, removes unnecessary types (so that everything is a bitvector operation) and removes support for `concat`, since it can't be supported without `HVector`/`hList`/similar.

### DIFF
--- a/SSA/Projects/CIRCT/Comb/Comb.lean
+++ b/SSA/Projects/CIRCT/Comb/Comb.lean
@@ -1,17 +1,24 @@
+import Qq
 import SSA.Core.Framework
 import SSA.Core.Framework.Macro
 import SSA.Core.MLIRSyntax.GenericParser
 import SSA.Core.MLIRSyntax.EDSL2
 import SSA.Core.Tactic.SimpSet
 import SSA.Projects.CIRCT.Comb.CombSemantics
+import SSA.Projects.InstCombine.Base
+import SSA.Core.MLIRSyntax.Transform.Utils
+import SSA.Projects.InstCombine.LLVM.CLITests
+
+open Qq Lean Meta Elab.Term Elab Command
+open MLIR
 
 namespace MLIR2Comb
 
 section Dialect
 
 inductive Ty
-| bv (w : Nat) : Ty -- A bitvector of width `w`.
-deriving DecidableEq, Repr, Lean.ToExpr
+| bitvec (w : Nat) : Ty -- A bitvector of width `w`.
+deriving DecidableEq, Repr, ToExpr
 
 inductive Op
 | add (w : Nat) (arity : Nat)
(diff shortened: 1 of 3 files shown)


SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -154,181 +140,84 @@ def ofList {Γ : Ctxt _} ty : (l : List ((ty : Comb.Ty) × Γ.Var ty)) → (h :
...
+  let binW : AST.ReaderM (Comb) (Nat) := do
+    let args ← args.assumeArity 2
+    return getVarWidth args[0]
+  -- 3-ary ops
+  let terW : AST.ReaderM (Comb) (Nat) := do
+    let args ← args.assumeArity 3
+    return getVarWidth args[0]
+  -- n-ary ops
+  let args' ← opStx.args.mapM (MLIR.AST.TypedSSAVal.mkVal Γ) -- will need to find a better way to do this
+  if h : args'.length = 0 then
```
alexkeizer:
```suggestion
  if h : args.length = 0 then
```

It seems you're parsing the arguments here again, does using the existing ParsedArgs not work? If it complains about `ParsedArgs.length` not existing, you can try `args.toList.length` to look at the underlying list

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -154,181 +140,84 @@ def ofList {Γ : Ctxt _} ty : (l : List ((ty : Comb.Ty) × Γ.Var ty)) → (h :
...
+    throw <| .generic s!" empty list of argument provided for the variadic op {repr opStx.args}"
```
alexkeizer, line 158: NIT: Note that this check is performed eagerly, so IIUC, even for non-variadic ops if you pass an empty list then it will throw this error

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -154,181 +140,84 @@ def ofList {Γ : Ctxt _} ty : (l : List ((ty : Comb.Ty) × Γ.Var ty)) → (h :
...
+      | some p => mkExprOf <| .icmp ps (← binW)
+      | _ => throw <| .generic s!" invalid attribute provided for {repr opStx.args}"
+    | ["Comb.mods"] => mkExprOf <| .mods (← binW)
+    | ["Comb.modu"] => mkExprOf <| .modu (← binW)
+    | ["Comb.shl"] => mkExprOf <| .shl (← binW)
+    | ["Comb.shrs"] => mkExprOf <| .shrs (← binW)
+    | ["Comb.shru"] => mkExprOf <| .shru (← binW)
+    | ["Comb.sub "] => mkExprOf <| .sub (← binW)
+    -- 3-ary
+    | ["Comb.mux"] => mkExprOf <| .mux (← terW)
```
alexkeizer, line 191: You could totally write something like:
```lean
let args <- args.assumeArity 3
if getVarWidth args[2] != 1 then
  -- throw error
else
  mkExprOf ...
```

To get a more informative error!

alexkeizer (commented):
> despite setting ϕ to 0, I can't seem to remove the check on the mvar in MkTy, what am I missing?

Not sure, the code look like it'd ought to work. What error or you seeing?

> I hacked a bit the arguments management for n-ary ops - I believe there are smoother ways to do that, could you please suggest how to do it better?

I left a suggestion

> I would like to throw a more informative error for Comb.mux when the third arg is not .bitvec 1 (condition), since as of now a very generic type error is raised. My solution would be to nest a further match in the parser, which I believe is what we're trying to avoid with the new infrastructure. What's the best way to proceed?

Adding an extra match would be fine, I think. I left a suggestion with a rough outline of what I'd write!

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: indeed, that was the error and using `toList` fixed it, thanks!

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: that should be okay, though, since no op in comb has no arguments

luisacicolini: there is always at least an argument

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: although I am afraid that would conflict with the variadic ops, say if one has 3 args. better leaving it as is I suppose.

SSA/Projects/CIRCT/Comb/Comb.lean
```diff
@@ -119,30 +116,19 @@ def_denote for Comb where
...
-      | some w₁', some w₂' => return .typeSum w₁' w₂'
-      | _, _ => throw .unsupportedType
-    | _ => throw .unsupportedType
+def mkTy : MLIR.AST.MLIRType 0 → MLIR.AST.ExceptM Comb (Comb.Ty)
   | MLIR.AST.MLIRType.int _ w =>
     match w with
-    | .concrete w' => return .bv w'
-    | .mvar _ => throw <| .generic s!"Bitvec size can't be an mvar"
+    | .concrete w' => return .bitvec w'
+    | .mvar _ => throw <| .generic s!" bitVec size can't be an mvar!"
```
alexkeizer: Is the capitalization change here intended? `bitVec` looks a bit funny

alexkeizer (approved):
LGTM, modulo a error message phrasing

SSA/Projects/CIRCT/Comb/Comb.lean
luisacicolini: fixed

PR #1238, opened 2025-05-17
chore: add `dc-comb` functor

luisacicolini:
This PR contains wip regarding the `comb`-`DC` functor and the updates to `DC` that were necessary to complete the fdunctor work.


luisacicolini:
Status update: I pushed the casting function that constructs a `Stream` from a `HVector` - lifting the result of `comb` operations into `dc`-digestible bits. I don't think `Stream.stuck` is what we want, given that we don't want all `none` to be lifted, but _also_ (and most importantly) the result of the comb operation. I thus tried using the function we had for the dialect already (`HVector.replicateToList`) and doing sth like `have := Stream.ofList (HVector.replicateToList h)`, which did not work. I'm now stuck and will context-switch to some other work, but any input would be welcome @ymherklotz @alexkeizer

PR #1241, opened 2025-05-18
feat: modify toString instance of LLVMAndRiscv

salinhkuhn:
This PR modifies the existing toString instance of the LLVMAndRiscV dialect to no longer use the Repr instance internally. This allows toString to differ from Repr.

### DIFF
--- a/SSA/Projects/LLVMRiscV/LLVMAndRiscv.lean
+++ b/SSA/Projects/LLVMRiscV/LLVMAndRiscv.lean
@@ -68,8 +68,11 @@ instance LLVMPlusRiscVSignature : DialectSignature LLVMPlusRiscV where
   | .castLLVM =>
       {sig := [Ty.llvm (.bitvec 64)], outTy := (Ty.riscv .bv), regSig := []}
 
-instance : ToString LLVMPlusRiscV.Ty  where
-  toString t := repr t |>.pretty
+instance : ToString LLVMPlusRiscV.Op  where
+  toString := fun
+  | .llvm llvm    => toString llvm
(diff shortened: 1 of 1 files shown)


SSA/Projects/LLVMRiscV/LLVMAndRiscv.lean
```diff
@@ -68,8 +68,11 @@ instance LLVMPlusRiscVSignature : DialectSignature LLVMPlusRiscV where
   | .castLLVM =>
       {sig := [Ty.llvm (.bitvec 64)], outTy := (Ty.riscv .bv), regSig := []}
 
-instance : ToString LLVMPlusRiscV.Ty  where
-  toString t := repr t |>.pretty
+instance : ToString LLVMPlusRiscV.Op  where
+  toString := fun
+  | .llvm llvm    => toString llvm
+  | .riscv riscv  => toString riscv
+  | _  => "builtin.unrealized_conversion_cast"
```
alexkeizer: It's better to match on the cast instruction explicitly. That way, if you ever add another constructor (and forgot to change this instance) it will give you a compiler error, rather than silently printing "builtin.unrealized_conversion_cast" for your new instruction.

alexkeizer (approved):
LGTM, feel free to merge after addressing my comment

SSA/Projects/LLVMRiscV/LLVMAndRiscv.lean
salinhkuhn: Implemented the change, thx.

PR #1245, opened 2025-05-19
feat: broken simpprocs for the LLVMAndRiscV dialect

salinhkuhn:
This PR adds simprocs for the LLVMAndRiscV dialect so that simp_peephole can recognize and simplify specific patterns. However, the current simprocs are not yet sufficient. In particular, simplification fails in the case of cast instructions—we are unable to eliminate the framework cast, and instead encounter a "move not type correct" error. This issue may be related to the newly introduced LLVM parser.


salinhkuhn:
In combination with the PR  #1247 the simpprocs successfully simplify the framework overhead. Thus they still require to be merged.

tobiasgrosser:
They are merged now. Can you click "update branch" and prepare this PR for a final review?

SSA/Projects/LLVMRiscV/simpproc.lean
```diff
@@ -0,0 +1,144 @@
...
+expressions, like [.const Nat], instead of building `[Nat]`, it builds `[Lean.Expr.Const ``Nat]`
+(i.e.., it seralizes the `Expr` as well!).
+Instead, we want a shallow serialization, where we just build `List.cons (.const Nat) List.nil`.
+-/
+def listExprToExprShallow (type : Option Expr) : List Expr → MetaM Expr
+| .nil => mkAppOptM ``List.nil #[type]
+| .cons x xs => do
+  let tailExpr ← listExprToExprShallow type xs
+  mkAppOptM ``List.cons #[type, x, tailExpr]
+
```
alexkeizer: This feels like it'd ought to exist in the standard library already. A loogle search for the signature shows [`Lean.Meta.mkListLit`](https://leanprover-community.github.io/mathlib4_docs/Lean/Meta/AppBuilder.html#Lean.Meta.mkListLit), whose [implementation](https://github.com/leanprover/lean4/blob/81b85203c90439fb7d7e0cec32130d62d8cb7a2e/src/Lean/Meta/AppBuilder.lean#L517-L524) looks like it does the right thing. Can you check whether replacing listExprToShallow with Lean.Meta.mkListLit (and dropping the former) also works?

SSA/Projects/LLVMRiscV/simpproc.lean
```diff
@@ -0,0 +1,144 @@
...
+    (xs : HVector TyDenote.toType (tys.map LLVMRiscV.Ty.llvm)) :
+  (llvmArgsFromHybrid (tys := ty :: tys) (HVector.cons x xs)) =
+  HVector.cons (f := TyDenote.toType) (a := ty) (as := tys) x (llvmArgsFromHybrid xs)
+   := rfl
+
+open Lean Meta Elab Qq in
+simproc [simp_denote] llvmArgsFromHybrid_cons_eq (llvmArgsFromHybrid (_) ) := fun e => do
+-- TODO: (comment from Alex)This ought to be .cons rather than a wildcard -------------^^^
+--       Unfortunately, then it the simproc again stops being applied when we'd
+--       expect it to, despite liberal use of `no_index`.
```
alexkeizer:
```suggestion
-- TODO: This ought to be .cons rather than a wildcard -------------^^^
--       Unfortunately, then it the simproc again stops being applied when we'd
--       expect it to, despite liberal use of `no_index`.
```

Note I took care so that the `^^^` symbols point towards the "this" the comment referenced, please don't break that!

SSA/Projects/LLVMRiscV/simpproc.lean
```diff
@@ -0,0 +1,144 @@
...
+      equality, only reduced to '{eq}'."
+  logInfo m!"final right-hand-side of equality is: {rhs}"
+  return .visit {
+    expr := rhs,
+    proof? := .some proof
+  }
+
+/- The following additional lemmas are needed to simplify the proof goals for the hybrid dialect.-/
+@[simp]
+theorem poisonOr_mk_some_eq_value (x : α) : { toOption := some x } = PoisonOr.value x := rfl
```
alexkeizer: Can you move this to the Poison.lean file, and name it `mk_some` (in the `PoisonOr` namespace). While you're at it, please also add the `mk_none` analogous lemma!

SSA/Projects/LLVMRiscV/simpproc.lean
```diff
@@ -0,0 +1,144 @@
...
+
+@[simp_denote]
+theorem liftM_eq_some (α : Type u) : @liftM Id Option _  α = some := by rfl
```
alexkeizer: Do you still need this? I thought we got rid of the superfluous option that got introduced?

alexkeizer (changes requested):
Could you write a comment that (to the best of your understanding; it's fine if this is not quite perfect) explains why we need the simprocs?

SSA/Projects/LLVMRiscV/simpproc.lean
salinhkuhn: Thx, I managed to replace the function with the already given implementation.

SSA/Projects/LLVMRiscV/simpproc.lean
salinhkuhn: done

salinhkuhn:
> Could you write a comment that (to the best of your understanding; it's fine if this is not quite perfect) explains why we need the simprocs?

I added a comment at the top of the file.
I implemented all your comments, therefore the PR would be ready for a second round of review @alexkeizer

alexkeizer (approved):
LGTM, although I don't really like that we have to resort to simprocs for this, and the current simprocs could likely be cleaned up, it's probably not worth to spend the effort that would take, so this hack shall live on.

(For posterity (i.e., git-blame purposes), I believe these simprocs were originally authored by @bollu)

PR #1246, opened 2025-05-20
chore: better document what is needed to use new EDSL elaborator

alexkeizer:


tobiasgrosser:
This looks great to me.

PR #1247, opened 2025-05-20
refactor: use new EDSL elaborator for LLVMAndRiscV Hybrid dialect

alexkeizer:
This PR changes the hybrid dialect to use the new EDSL elaboration strategy, which should completely eliminate any mention of `ParsedArgs` and related parser internal details from the elaborated programs.

@salinhkuhn can you confirm whether this PR fixes the issues you were running into with your `simproc` branch?


salinhkuhn:
> This PR changes the hybrid dialect to use the new EDSL elaboration strategy, which should completely eliminate any mention of `ParsedArgs` and related parser internal details from the elaborated programs.
> 
> @salinhkuhn can you confirm whether this PR fixes the issues you were running into with your `simproc` branch?

Yes, thank you — it enables me to write proofs, but we still need the existing simpprocs. Together, the framework overhead gets eliminated.

PR #1248, opened 2025-05-20
feat: extend hybrid dialect to support generic width

salinhkuhn:
This PR extends the hybrid dialect to support generic widths for LLVM IR instructions.
It inlines the getOutputWidth function so that any errors are thrown within the same monad as the mkExpr operation and since the output width of an instruction is not used elsewhere in the file.

### DIFF
--- a/SSA/Projects/LLVMRiscV/LLVMAndRiscv.lean
+++ b/SSA/Projects/LLVMRiscV/LLVMAndRiscv.lean
@@ -2,7 +2,8 @@ import SSA.Projects.InstCombine.Base
 import SSA.Projects.RISCV64.Base
 import SSA.Projects.RISCV64.Syntax
 import SSA.Projects.InstCombine.LLVM.EDSL
-
+import SSA.Projects.RISCV64.PrettyEDSL
+import SSA.Projects.InstCombine.LLVM.PrettyEDSL
 open InstCombine(LLVM)
 namespace LLVMRiscV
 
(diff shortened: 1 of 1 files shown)


SSA/Projects/LLVMRiscV/LLVMAndRiscv.lean
```diff
@@ -264,9 +265,14 @@ def mkExpr (Γ : Ctxt _) (opStx : MLIR.AST.Op 0) :
...
-      | .riscv (.bv) => mkExprOf <| .castRiscv
-      | .llvm (.bitvec 64) => mkExprOf <| .castLLVM
-      | _ => throw <| .unsupportedOp s!"unsupported operation {repr opStx}"
+      | .riscv (.bv) =>
+         match opStx.res with
+        | res::[] =>
+           match res.2 with
+          | .int _ w =>  mkExprOf <| .castRiscv w.toConcrete
+          | _ => throw <| .generic s!"A cast operation must output an integer type"
+        | _ => throw <| .generic s!"A cast operation must only have one return type"
```
alexkeizer, line 274: It would still be good to put this in a function (defined in the current file), just to keep the code a bit more readable, even if we copy-paste the implementation of said function.

alexkeizer (approved):
LGTM; it would IMO be a bit cleaner to keep the output width code in a function (when I suggested to copy it, I merely meant to copy the function, not to inline the whole thing). Either way, feel free to merge

SSA/Projects/LLVMRiscV/LLVMAndRiscv.lean
salinhkuhn: okay, I will do it in a follow up PR then later. I know want to test if I can write correct proofs using the new widths.

PR #1249, opened 2025-05-20
feat: new LLVM dialect with proper UB (for division)

alexkeizer:
This PR adds SLLVM, a new dialect which will eventually model memory, structured-control flow and UB, but for now is just a wrapper around the existing LLVM arithmetic operations, with proper UB-as-a-side-effect semantics.

For now, the operations that may throw immediate UB are udiv, sdiv, urem and srem, all when the divisor is either poison or `0`.

Follow up PRs will adapt (a copy of) AliveAutoGenerated to use these UB semantics, and build up the tactic infrastructure to decide refinement.


tobiasgrosser (approved):
LGTM

PR #1250, opened 2025-05-20
feat: generalize PoisonOr refinement to build on top of a refinement relation of underlying values

alexkeizer:
This PR adapts refinement of PoisonOr values s.t. `value a` is refined by `value b` whenever `a` is refined by `b` (rather than hard-coding `a = b` as the condition), by requiring some underlying refinement relation on the wrapped values. This will become relevant when modelling UB, as then we'll have two layers of optionals (UB and poison), which we'd ideally both model as `PoisonOr`. This change thus ensures that stack gets the right poison semantics, so that `value poison` indeed is refined by `value (value 1#w)`.

To make the existing LLVM semantics work, we add a scoped instance of refinement for pure bitvectors, based on equality, so that the meaning of refinement of LLVM.IntW is the same as before. We make this instance scoped so that other dialects can still make other choices as to what refinement of bitvectors means, but it does mean that some lemmas won't apply out of the box. Instead of having to remember to open the right namespace for each proof file, I fixed this by using `simp` instead (which is able to unify against the not-in-scope instance in the proof goal, rather than trying to synthesize a new instance).

### DIFF
--- a/SSA/Core/Framework/Refinement.lean
+++ b/SSA/Core/Framework/Refinement.lean
@@ -43,12 +43,28 @@ NOTE: This typeclass is not intended for dialect implementors. Please implement
 class Refinement (α : Type) where
   IsRefinedBy : α → α → Prop
 
-instance [Refinement α] : HRefinement α α where
+instance instHRefinementOfRefinement [Refinement α] : HRefinement α α where
   IsRefinedBy := Refinement.IsRefinedBy
+
+/-! #### Trivial Refinement -/
+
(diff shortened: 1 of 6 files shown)


tobiasgrosser (approved):

tobiasgrosser:
This breaks our CI.

alexkeizer:
CI is fixed by opening the appropriate scope (so the scoped instance of Refinement of bitvectors is brought into scope).

PR #1261, opened 2025-05-22
feat: add instruction lowering 64-bit & according PoisonOr lemmas 

salinhkuhn:
This PR adds the proof for 64-bit addition lowering. The proofs are structured so that each one follows the same sequence of tactics.
To support this, we added missing lemmas to the PoisonOr file and marked two cast definitions in the LLVMAndRiscV dialect with the simp_riscv attribute.

### DIFF
--- a/SSA/Core/Util/Poison.lean
+++ b/SSA/Core/Util/Poison.lean
@@ -129,6 +129,11 @@ variable {a : α}
 @[simp] theorem mk_some (x : α) : { toOption := some x } = PoisonOr.value x := rfl
 @[simp] theorem mk_none : { toOption := none (α := α) } = PoisonOr.poison := rfl
 
+@[simp_denote, simp]
+theorem toOption_getSome : (PoisonOr.value x).toOption.getD y = x := by rfl
+@[simp_denote, simp]
+theorem toOption_getNone : (PoisonOr.poison).toOption.getD y = y := by rfl
+
 end Lemmas
 
 /-! ### Refinement -/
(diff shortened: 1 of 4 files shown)


tobiasgrosser (approved):
Nice!

alexkeizer:
@salinhkuhn this PR broke CI!

The reported error is a bit confusing, it complains about a change in AliveStatements. If we look at the diff, though, we see the core of the issue is that this PR introduced some warnings:
```
 +    warning: /home/runner/work/lean-mlir/lean-mlir/SSA/Core/Util/Poison.lean:187:0: automatically included section variable(s) unused in theorem 'PoisonOr.PoisonOr.eq_squb':
+      [HRefinement α α]
+    consider restructuring your `variable` declarations so that the variables are not in scope or explicitly omit them:
+      omit [HRefinement α α] in theorem ...
+    note: this linter can be disabled with `set_option linter.unusedSectionVars false`
+    warning: /home/runner/work/lean-mlir/lean-mlir/SSA/Core/Util/Poison.lean:187:8: The namespace 'PoisonOr' is duplicated in the declaration 'PoisonOr.PoisonOr.eq_squb'
+    note: this linter can be disabled with `set_option linter.dupNamespace false`
+     := by
```

Please don't ignore warnings, sadly our CI doesn't test for them, and we do have warnings in places we shouldn't have, but we really should try our best not to introduce any more warnings. If you don't understand what a warning means, just ask on Zulip, and we can tell you if it's harmless or not.

In this case, there are two warnings:
* Firstly, it warns that there is already a `variable [HRefinement _ _]` in scope, so the extra `[HRefinement]` bound on the theorem is redundant. You likely can just remove the theorem bound.
* Secondly, the file already has a `namespace PoisonOr` command at the top, so you should name the theorem just `eq_squb` and it'll be put in the right namespace. Currently, the full name of the theorem is `PoisonOr.PoisonOr.sq_ub`

salinhkuhn:
@alexkeizer  I now fixed the errors in the following PR #1268

PR #1268, opened 2025-05-23
fix: fix warnings due to naming error and redundant variables

salinhkuhn:
This PR removes the warnings due to the Poison.eq_squb lemma that were introduced because of adding a typeclass constraint redundantly. It additionally improves the naming of the lemma.


tobiasgrosser:
Cool.

alexkeizer:
Thanks for fixing this!

PR #1273, opened 2025-05-26
fix: change cast semantics to signExtend.

salinhkuhn:
This PR addresses a bug discovered while verifying instruction lowering for operand widths other than 64-bit.


salinhkuhn:
@alexkeizer @tobiasgrosser  Could one of you approve this change quickly ?
It needed to be fixed because now we lower different widths and therefore setWidth is wrong because it performs zeroExtension and not signExtension.

tobiasgrosser (approved):

alexkeizer:
@salinhkuhn note for the future: if a PR fixes a bug then it's super helpful if in the PR description you either put a short description of the bug (1 or 2 sentences) or a reference to a GH issue if it exists.

That way, if someone in ~6 months wonders why the cast semantics are what they are, they'll find this PR and immediately get the counter-example of why zero-extending wouldn't work.

PR #1274, opened 2025-05-26
feat: add lowering for truncate and cmp instruction from LLVM to RiscV

salinhkuhn:
This PR adds the lowering for the truncate and compare LLVM instructions.


salinhkuhn:
@luisacicolini  This is also ready for syntax review

SSA/Projects/LLVMRiscV/Pipeline/icmp.lean
```diff
@@ -0,0 +1,523 @@
...
+    simp_alive_undef
+    simp_alive_ops
+    simp_alive_case_bash
+    simp_alive_split
+    all_goals
+    simp only [PoisonOr.toOption_getSome, BitVec.signExtend_eq]
+    bv_decide
+  }
+
+def icmp_ugt_riscv_i32 := [LV| {
```
luisacicolini, line 41:
```suggestion
def icmp_ugt_riscv_i32 := [LV| {
```
is this name correct? I see you are only using `sltu`, could you update the def to be consistent?

luisacicolini: and could you check this for all the defs in the file?

salinhkuhn: Yes, its confusing but the lowering of the icmp ugt instructions is implemented using sltu. I always name the definition according to the llvm instruction I am lowering and it happens that it lowers to sltu with the arguments reversed.

salinhkuhn: Yes, its confusing but the lowering of the icmp ugt instructions is implemented using sltu. I always name the definition according to the llvm instruction I am lowering and it happens that it lowers to sltu with the arguments reversed.

luisacicolini (approved):
lgtm, thanks!

PR #1276, opened 2025-05-26
fix: finish proof goals in trunc

salinhkuhn:
This PR closes proof goals that broke due to the change in PR  #1273


salinhkuhn:
@luisacicolini

SSA/Projects/LLVMRiscV/Pipeline/trunc.lean
```diff
@@ -37,7 +37,9 @@ def llvm_trunc_riscv_i64_to_i32 : LLVMPeepholeRewriteRefine 32 [Ty.llvm (.bitvec
     simp_alive_ops
     simp_alive_case_bash
     simp_alive_split
-    all_goals simp
+    all_goals
+    simp
```
luisacicolini: I would feel safer using `simp only` here, what's the usual convention? cc @tobiasgrosser

tobiasgrosser: Yes, its `simp only`, as this is not a terminal simp but it is followed by `bv_decide`.

salinhkuhn: fixed now

luisacicolini (approved):
lgtm

PR #1277, opened 2025-05-26
feat: improve multiPeepholeRewriter

salinhkuhn:
This PR updates the multiPeepholeRewriter to support a dependently typed list of rewrites. This enables handling rewrites with varying contexts and return types in a single pass. Previously, all rewrites in a list had to share the same context and return type, requiring multiple calls to the rewriter for different cases.

### DIFF
--- a/SSA/Core/Framework.lean
+++ b/SSA/Core/Framework.lean
@@ -2634,12 +2634,12 @@ variable {d : Dialect} [DialectSignature d] [DecidableEq (Dialect.Ty d)] [Decida
 [TyDenote d.Ty] [DialectDenote d] [Monad d.m] in
 /--  rewrite with the list of peephole optimizations `prs` at the `target` program, at location `ix`
 and later, running at most `fuel` steps. -/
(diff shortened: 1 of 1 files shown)


tobiasgrosser (approved):

SSA/Core/Framework.lean
```diff
@@ -2634,32 +2634,33 @@ variable {d : Dialect} [DialectSignature d] [DecidableEq (Dialect.Ty d)] [Decida
 [TyDenote d.Ty] [DialectDenote d] [Monad d.m] in
 /--  rewrite with the list of peephole optimizations `prs` at the `target` program, at location `ix`
 and later, running at most `fuel` steps. -/
-def multiRewritePeepholeAt (fuel : ℕ) (prs : List (PeepholeRewrite d Γ t))
+def multiRewritePeepholeAt (fuel : ℕ) (prs : List (Σ Γ, Σ ty, PeepholeRewrite d Γ ty))
```
alexkeizer, line 2637: @salinhkuhn note that you can abbreviate `Σ Γ, Σ ty, ` with a single sigma `Σ Γ ty, ` which means the same.

PR #1280, opened 2025-05-29
feat: llvm.select lowering

salinhkuhn:
This PR adds the lowering for the llvm.select operation.


salinhkuhn:
@alexkeizer @luisacicolini hi, could one of you quickly check this PR. I did not use a simp only before bv_decide because the simp is applied to all subgoals. What is the usual approach here ? Splitting up every case separately and then using simp only or is it fine how I did it ?

SSA/Projects/LLVMRiscV/select.lean
```diff
@@ -0,0 +1,108 @@
...
+      split <;> simp
+    case value.poison.value =>
+      intro x x'
+      split <;> simp <;> bv_decide
+    case value.value.poison =>
+      intro x x'
+      split <;> simp <;> bv_decide
+    case value.value.value =>
+      intro x x' x''
+      split <;> simp <;> bv_decide
```
luisacicolini: you have two approaches: 
1. separate each case and have `simp only ...` and `bv decide` there (each case will start with `\.`, e.g. [here](https://github.com/opencompl/lean4/blob/1fd7206f0038e851b645422ec1dc77f75b104387/src/Init/Data/BitVec/Lemmas.lean#L115))
2. if you `simp?` at this point you'll see all the theorems used in all branches: you can put all them together in the same `simp only` - the theorems that are not necessary in some branc will be ignored. However, this might not work, as some rewrites might overlap in weird ways in `simp`. Give it a shot :)

salinhkuhn: I fixed the simps. simp? unfortunately did not suffice,  as you mentioned  it did not work properly because of some rewrites that are overlapping.

luisacicolini: this lgtm now, but I'd like to wait and see if @alexkeizer has any sugggestions to make the proof more coincise :)

SSA/Projects/LLVMRiscV/select.lean
```diff
@@ -0,0 +1,136 @@
+import SSA.Projects.LLVMRiscV.PeepholeRefine
+import SSA.Projects.LLVMRiscV.simpproc
+import SSA.Projects.RISCV64.Tactic.SimpRiscV
+
+open LLVMRiscV
+set_option Elab.async true
+
+/-
```
alexkeizer:
```suggestion
/!-
```

SSA/Projects/LLVMRiscV/select.lean
```diff
@@ -0,0 +1,136 @@
...
+LLVM version 11.0 lowers `select` to a conditional branch regardless of the
+optimization level used. However, our current setup does not support branches.
+
+To handle this, we manually encode the semantics of the `select` instruction
+using bitwise operations. While this results in a longer RISC-V instruction
+sequence compared to LLVM's native lowering, it allows us to support lowering
+the instruction within our framework.
+-/
+
+/- # select, RiscV (leveraging bit-wise operations)  -/
```
alexkeizer:
```suggestion
/!- # select, RiscV (leveraging bit-wise operations)  -/
```

SSA/Projects/LLVMRiscV/select.lean
```diff
@@ -0,0 +1,136 @@
...
+    llvm.return %10 : i64
+  }]
+
+def select_llvm_64 := [LV| {
+    ^entry (%cond : i1, %arg0: i64, %arg1: i64):
+    %0 = "llvm.select"(%cond, %arg0, %arg1) <{"fastmathFlags" = #llvm.fastmath<none>}> : (i1, i64, i64) -> i64
+    llvm.return %0 : i64
+  }]
+
+def select_riscv_select_llvm_64 : LLVMPeepholeRewriteRefine 64 [Ty.llvm (.bitvec 64), Ty.llvm (.bitvec 64), Ty.llvm (.bitvec 1)] where
```
alexkeizer: Check your line lengths throughout the file

SSA/Projects/LLVMRiscV/select.lean
```diff
@@ -0,0 +1,136 @@
...
+  lhs:= select_llvm_64
```
alexkeizer:
```suggestion
  lhs := select_llvm_64
```
whitespacing

SSA/Projects/LLVMRiscV/select.lean
```diff
@@ -0,0 +1,136 @@
...
+  rhs:= select_riscv
+  correct := by
+    unfold select_riscv select_llvm_64
+    simp_peephole
+    simp_riscv
+    simp [LLVM.select]
```
alexkeizer: This is a [non-terminal simp](https://lean-lang.org/doc/reference/latest/The-Simplifier/Terminal-vs-Non-Terminal-Positions/) (read the link!).

In this case, please use the standard tactics we have for this:
```
simp_alive_ops
simp_alive_undef
```

(NOTE: I haven't double-checked which order they go, or if it's called _alive_ or _llvm_, we are somewhat inconsistent with this, but have a look at AliveStatements to see the standard invocation which has the right names and orders in every proof in that file)

SSA/Projects/LLVMRiscV/select.lean
```diff
@@ -0,0 +1,136 @@
...
+    simp_alive_case_bash
```
alexkeizer, line 52: Here, too, you're missing the `simp_alive_split` tactic
```suggestion
    simp_alive_case_bash
    simp_alive_split
```

SSA/Projects/LLVMRiscV/select.lean
```diff
@@ -0,0 +1,136 @@
...
+    case value.poison.poison =>
+      intro x
+      split <;> simp
+    case value.poison.value =>
+      simp [PoisonOr.toOption_getSome, PoisonOr.toOption_getNone, BitVec.and_zero,
```
alexkeizer, line 57: This touches upon the proof automation tips I sent you on Zulip, but `PoisonOr.toOption_getSome` and `PoisonOr.toOption_getNone` should both be part of at least the `simp_llvm` simpset, maybe some of the others as well if the pattern still shows up. I suspect once you add these rewrites, a decent number of the cases should be solved automatically, and those that aren't ought to be simpler

alexkeizer (commented):
@salinhkuhn please make sure you check for (and fix) any too-long-lines and other formatting mishaps.

Otherwise, I left some suggestions on how to improve the proofs, but it touches on largely the same stuff that I already sent you in the proof-automation Zulip thread. Try to implement some of the suggestions, but if it doesn't work, just merge the PR as-is (post fixing formatting) and we'll clean up the proof once you're reading to address proof automation together.

SSA/Projects/LLVMRiscV/select.lean
salinhkuhn: That one here was on purpose because when I use it, it loose the ability to name the hypothesis that gets introduced on the split. Maybe we should look at this in person once

SSA/Projects/LLVMRiscV/select.lean
salinhkuhn: I will merge the PR as it is and then fix this in a subsequent merge right now as you suggested. But when I first merge this PR now with the wonky proof then I can fix it more efficiently.

PR #1281, opened 2025-05-29
feat: examples that are labour intensive 

salinhkuhn:
This PR contains an example.lean file to illustrate what is currently annoying and labour-intensive for the author  to write, and won't scale well.


alexkeizer:
@salinhkuhn just to be sure, did you actually want to merge this, or did you just open this PR to have something to show for the Zulip discussion? In the latter case, a PR is perfect for that, but you could mark it as "draft" to signal you're not actually looking for a merge/review. If the former, let me know and I'll give a proper review (but I would suggest we wait on that until after we've paired on the proof automation)

tobiasgrosser:
Sarah's work has meanwhile all been merged.

PR #1282, opened 2025-05-29
feat: adding lowerings for constants 

salinhkuhn:
This PR adds the lowerings for constants. We use a python script to generate the rewrites between -25 and 25. Since Lean accepts the proofs, we have guaranteed correctness for our rewrites no matter how they are generated.

The script generates the proof of lowering a llvm constant (rhs) to a RISC-V load immediate instruction (lhs). Below is an example for the constant one being lowered to a RISC-V load of an immediate with value one.
```
def llvm_const_lower_riscv_li1 : LLVMPeepholeRewriteRefine 64 [] :=
  {lhs:= const_llvm_1 rhs:= li_riscv_1,
   correct := by
    unfold const_llvm_1 li_riscv_1
    simp_peephole
    simp_riscv
    simp_alive_ops
    simp
  }
```


salinhkuhn:
@alexkeizer @luisacicolini  This PR is also ready for review from my side. Just checking the script should be enough.

SSA/Projects/LLVMRiscV/Pipeline/scripts/auto_constants.py
```diff
@@ -0,0 +1,52 @@
...
+    return header + body + array_definition
+
+# Add global array of rewrites
+
+# get the parent directory because I placed the script in a dedicated folder.
+script_dir = Path(__file__).parent
+parent_dir = script_dir.parent
+output_file = parent_dir / "const.lean"
+
+# writting to lean file
```
luisacicolini, line 67: typo

SSA/Projects/LLVMRiscV/Pipeline/scripts/auto_constants.py
```diff
@@ -0,0 +1,52 @@
+from pathlib import Path
+def generate_lean_proofs(max_val=25):
```
luisacicolini, line 19: could you add in a comment a single example of a proof you are generating?

salinhkuhn: I added it at the top of the file containing the script. I will also add one to the PR description.

SSA/Projects/LLVMRiscV/Pipeline/scripts/auto_constants.py
```diff
@@ -0,0 +1,68 @@
+from pathlib import Path
+"""
+# This script generates Lean proofs for constant lowering of RISC-V li instructions.
+It generates a riscv lefthandside and an llvm constant right hand side for each integer in the
```
luisacicolini: could you be consistent with the naming? e.g. `lefthandside` vs. `right hand side`

salinhkuhn: Yes, its wrong there.

SSA/Projects/LLVMRiscV/Pipeline/scripts/auto_constants.py
salinhkuhn: fixed

SSA/Projects/LLVMRiscV/Pipeline/const.lean
```diff
@@ -0,0 +1,2233 @@
+-- AUTOGENERATED Lean file: Constant lowering for RISC-V li instructions
```
alexkeizer, line 1: Good thinking to put this comment here! For completeness sake, you might also add a comment that explicitly points to which script was used to generate it, or with instructions in case someone else wanted to tweak/regenerate the lowerings

SSA/Projects/LLVMRiscV/Pipeline/const.lean
```diff
@@ -0,0 +1,2233 @@
...
+
+ open LLVMRiscV
+
+    def li_riscv_n50 := [LV| {
+    ^entry ():
+      %0 = "li"() {imm = -50 : !i64}  : (!i64) -> (!i64)
+      %1 = "builtin.unrealized_conversion_cast"(%0) : (!i64) -> (i64)
+      llvm.return %1 : i64
+  }]
+def const_llvm_n50 : Com  LLVMPlusRiscV [] .pure (.llvm (.bitvec 64))  := [LV| {
```
alexkeizer: Your whitespace looks off here, can you try to fix your script to format things a bit nicer?

In particular, there shouldn't be spaces before that `open`, nor before that `def`, and there should an empty line after the definition (i.e., between lines 14 and 15)

SSA/Projects/LLVMRiscV/Pipeline/const.lean
```diff
@@ -0,0 +1,2233 @@
...
+    ^entry ():
+      %1 = llvm.mlir.constant (-50) : i64
+      llvm.return %1 : i64
+  }]
+def llvm_const_lower_riscv_lin50 : LLVMPeepholeRewriteRefine 64 [] :=
+  {lhs:= const_llvm_n50, rhs:= li_riscv_n50,
```
alexkeizer:
```suggestion
  {lhs := const_llvm_n50, rhs := li_riscv_n50,
```
mind your whitespaces around `:=`.

NIT: You could also consider using `where` instead of `{ ... }`, but I leave that choice up to you

SSA/Projects/LLVMRiscV/Pipeline/const.lean
```diff
@@ -0,0 +1,2233 @@
...
+   correct := by
+    unfold const_llvm_n50 li_riscv_n50
+    simp_peephole
+    simp_riscv
+    simp_alive_ops
+    simp
+  }
+
+def li_riscv_n49 := [LV| {
```
alexkeizer:
```suggestion
def liRiscvN49 := [LV| {
```
Remember that definitions ought to have lowerCamelCase names (i.e., the first letter is lowercase, then every new "word"/component starts with an Uppercase character, without any underscores/dashes to separate components)

alexkeizer (approved):
Looks good! I just have some suggestions about naming & formatting, feel free to merge once that is fixed

PR #1285, opened 2025-05-29
chore: edit to be in-line with what chris implemented

bollu:


bollu:
supereded by https://github.com/opencompl/lean-mlir/pull/1286

PR #1293, opened 2025-06-02
K induction cycle cond

bollu:
We add the cyclic condition check to our implementation of verified k-induction, which makes the solver complete even in the presence of cycles: https://web.stanford.edu/class/cs357/lecture12.pdf


bollu:
Superceded by #1319 , which incorporates our code into the verified solver.

PR #1295, opened 2025-06-02
add: basic block extracting script& parsing tests& processing script

salinhkuhn:
This PR adds example files to the LLVMAndRISCV test folder that contain the LLVM IR dialect lowering for each supported instruction in our fragment. These files were then used to verify that they parts (basic blocks) can be successfully parsed in Lean-MLIR.
Additionally, this PR adds the following scripts:
A script to extract a basic block from a file.
A script to preprocess an LLVM IR file so that it can be passed to the opt tool.
A script that performs the full pipeline: preprocessing, passing the file to opt, and writing the output to a new file.


SSA/Projects/InstCombine/Base.lean
```diff
@@ -153,7 +153,7 @@ instance : ToString (MOp 0) where
      | .unary _w op => s!"\"{toString op}\""
      | .binary _w op => s!"\"{toString  op}\""
      | .select  _w => "select"
-     | .icmp  _pred _w => "icmp"
+     | .icmp  _pred _w => s!"llvm.icmp.{_pred}"
```
alexkeizer: Why do you change this? Besides the constant, none of the other instructions include the llvm. prefix either.

SSA/Projects/LLVMRiscV/Pipeline/scripts/asm-to-ssa.py
alexkeizer, line 1: Is this supposed to be an empty file?

SSA/Projects/LLVMRiscV/Pipeline/use/ReadMe.txt
```diff
@@ -0,0 +1,24 @@
+Here we describe how to apply use our instruction selection and apply to an LLVM IR  file.
+There are several options dependeing on the desired output format.
```
alexkeizer:
```suggestion
There are several options depending on the desired output format.
```

SSA/Projects/LLVMRiscV/Pipeline/use/ReadMe.txt
```diff
@@ -0,0 +1,24 @@
...
+
+define <rty> @function(<ty> %i, ....) {
+    entry:
+    ....
+     ret <rty> %some
+    }
+    
+     define i32 @main(....) {
+      %out = call <rty> @function(...)
+      ret <rty>%out
```
alexkeizer, line 14: If this is supposed to be a code example, you could use a markdown code block.
That is, write it like so
````
```
code goes here
```
````
That just makes it a bit easier to see where the example starts and ends

File: SSA/Projects/LLVMRiscV/Pipeline/use/ReadMe.txt [hunk 4]
```diff
@@ -0,0 +1,24 @@
...
+    
+
+      Where the function will be lowered by our pass. 
+
+1.)  Either pass the file to the opt tool and observe the lowerig the terminal:
+to do: the complete command.
+
+2.) Add a file to this folder and run <to do> and observe the extracted lowering of your function in
+equally named output file.
```
alexkeizer, line 29: Could you include in your readme which (non-standard) programs you expect to be installed for the scripts to work.

SSA/Projects/LLVMRiscV/tests/LLVMIR/preprocess.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Use it as follows:
+ #./preprocess.sh <input_file>
+
+INPUT_FILE="$1"
+
```
alexkeizer, line 7: NIT: It's not super crucial, but it'd be nice if the scripts still worked properly when called from a different working directory.

For example, if I'm in the main directory, and called `./SSA/Projects/LLVMRiscV/tests/LLVMIR/preprocess.sh <input_file>`, I believe it currently would fail due to the use of relative paths. I believe adding
```
cd $(dirname "$0")
```
would fix this (by explicitly changing the working directory to the dir that contains the script).

You could do the same for all your other scripts

SSA/Projects/LLVMRiscV/tests/LLVMIR/preprocess.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
alexkeizer:
```suggestion
#!/usr/bin/env bash
```
On most systems what you wrote is perfectly fine; it just so happens that my system is weird and has bash in a non-standard location, so it's better to lookup bash with `/usr/bin/env`

SSA/Projects/LLVMRiscV/tests/LLVMIR/process.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
alexkeizer:
```suggestion
#!/usr/bin/env bash
```

alexkeizer (approved):
I didn't review the scripts too thoroughly, as I assume you've tested that they work properly?

Otherwise, LGTM, but you seem to have a stray change change to the printing, let's keep that in a separate PR, as I'm not 100% sure what the right solution is there.

SSA/Projects/LLVMRiscV/tests/LLVMIR/preprocess.sh
salinhkuhn: Cool ,thx. Didn't know that but yes its nicer if they work from everywhere

SSA/Projects/LLVMRiscV/Pipeline/use/ReadMe.txt
salinhkuhn: done

PR #1301, opened 2025-06-04
feat: riscv printing& parsing in opt same as in llvm.

salinhkuhn:
This PR builds the infrastructure to parse RISC-V using the opt tool. It also fixes a semantic bug in the RISC-V dialect: the load immediate instruction (li). This was the only instruction not derived from Sail, as it is a pseudoinstruction. This PR is very similar to a previous PR doing the same for llvm and the hybrid dialect.


SSA/Projects/LLVMRiscV/ParseAndTransform.lean
```diff
@@ -44,11 +44,12 @@ def passriscv64 (fileName : String) : IO UInt32 := do
...
-          IO.println s!"{repr lowered}"
+          IO.println s!"{toString lowered}"
           return 0
         | _ =>
         IO.println s!" debug: WRONG RETURN TYPE : expected Ty.llvm (Ty.bitvec 64) "
         return 1
       | _ =>
       IO.println s!" debug: WRONG EFFECT KIND : expected pure program "
       return 1
+#eval! passriscv64 "test/LLVMDialect/InstCombine/bb0.mlir"
```
alexkeizer, line 55: This line will be noisy, meaning that it will evaluate, and print the result to standard out on every build.

If you used this eval line to debug, it's better to remove it. If you want to keep it around as a test of sorts, then you should put a `#guard_msgs in` in front of it: that way the output is checked against the preceding docstring and *not* printed to standard out. Hence, it removes the noise from the build log *and* ensures it actually functions as an automated test and fails to build if the output has changed.

If you're not sure how to use `#guard_msgs in` properly, please ask on Zulip!

SSA/Projects/RISCV64/Base.lean
```diff
@@ -134,6 +135,10 @@ instance (ty : Ty) : Inhabited (TyDenote.toType ty) where
   default := match ty with
   | .bv  => 0#64
 
+instance : ToString (Ty) where
+  toString t :=  match t with
+  | Ty.bv => "!riscv.reg"
+
```
alexkeizer, line 141: I think you already have a `ToString Ty` instance on line 122!

It seems that this instance has the desired behaviour ~~(although I thought we called the type `!i64`, or did you already change the parser to expect `!riscv.reg`?)~~, so I believe we should remove the earlier instance.

EDIT: I see you've indeed changed the parser

SSA/Projects/RISCV64/ParseAndTransform.lean
```diff
@@ -0,0 +1,45 @@
...
+
+/-- `regionTransform_RiscV` attempts to transform a region into a `Com` of type `RV64`.
+It throws an error if the transformation fails. -/
+def regionTransform_RISCV (region : Region 0) : Except ParseError
+  (Σ (Γ' : Ctxt RISCV64.Ty ) (eff : EffectKind)
+  (ty : RISCV64.Ty ), Com RV64 Γ' eff ty) :=
+  let res := mkCom (d:= RV64) region
+  match res with
+  | Except.error e => Except.error s!"Error:\n{reprStr e}"
+  | Except.ok res => Except.ok res
```
alexkeizer, line 23: Be mindful of indentation! The type of a def ought to be indented 2 spaces more than the body (i.e., the thing that comes after `:=`)

SSA/Projects/RISCV64/ParseAndTransform.lean
```diff
@@ -0,0 +1,45 @@
...
+ parseRegionFromFile fileName regionTransform_RISCV
+
+def parseAsRiscv (fileName : String ) : IO UInt32 := do
+    let icom? ← parseComFromFile_LLVMRISCV fileName
+    match icom? with
+    | none => return 1
+    | some (Sigma.mk _Γ ⟨_eff, ⟨_retTy, c⟩⟩) => do
+      IO.println s!"{toString c}"
+      return 0
+private def test_simple := [RV64_com| {
```
alexkeizer, line 40: Keep at least one newline in between definitions, please

SSA/Projects/RISCV64/ParseAndTransform.lean
```diff
@@ -0,0 +1,45 @@
...
+  ^bb0(%e1 : !i64, %e2 : !i64 ):
+  %1 = add %e1, %e2 : !i64
+       ret %1 : !i64
+}]
+#eval! test_simple
```
alexkeizer, line 45: See my earlier comment about `#eval!`. If you want this to be a proper test, please add a `#guard_msgs in` in front of it.

alexkeizer (commented):
@salinhkuhn I did a post-merge review of this one, and there are a few (minor!) issues. Could you make some follow up PR to address my comments?

PR #1311, opened 2025-06-09
fix: remove "unsafe" in CSE

salinhkuhn:
This PR removes the unsafe keyword from the definitions in CSE.lean.
Previously, Lean did not know that the types were indeed inhabited.


tobiasgrosser:
Nice. Maybe @bollu can also give a 👍 and then its ready to merge?

tobiasgrosser (approved):

tobiasgrosser:
I guess @bollu is busy today. Let's do a post-merge review.

alexkeizer:
Thanks, it should be much nicer to use this without the unsafes!

I did a quick post-merge review and LGTM

PR #1318, opened 2025-06-13
fix :  Opt tool fails to parse i1 types.

salinhkuhn:
When running lake exec opt bb0i1.mlir, an error occurs indicating that the contents of the file bb0i1.mlir are not parsed correctly. The error can be resolved by modifying the file in one of the following ways:
Replacing false with 0
Annotating the attribute with an explicit type, such as i64 or i1. (where i64 is wrong).

It is not clear to me where the parser assumes the i64 type and how to change it such that the example parses without (wrong) modification.


alexkeizer:
@salinhkuhn Is this supposed to be merged? It claims to fix something, but I don't really see how; should this have been an issue report?

salinhkuhn:
@alexkeizer No, I think Tobias request it back then we I encountered the issues that false/true couldn't be parsed. I think the solution was then just not generate this, but I will double check. But for sure no mergeable.

tobiasgrosser:
This is likely resolved already.

PR #1325, opened 2025-06-14
fix: compile failure of `SSA.Projects.ModArith.Examples`

znssong:
The command `lake build SSA.Projects.ModArith.Examples` failed due to recent changes in the `SSA` formalization framework. We fixed it to ensure successful compilation.


SSA/Projects/ModArith/Basic.lean
```diff
@@ -54,6 +54,14 @@ inductive Ty (q : ℕ) where
 | modLike
 deriving DecidableEq, Repr, Inhabited
 
+/--
+We provide a `ToString` instance: this is a human-readable name for each type.
+-/
+instance {q} : ToString (Ty q) where
+  toString := fun
+  | .integer => "ℤ"
+  | .modLike => "ℤ/qℤ"
```
alexkeizer:
```suggestion
instance {q} : ToString (Ty q) where
  toString := fun
  | .integer => "ℤ"
  | .modLike => "!R"
```

The intention here is to convert the types into the same string that the parser would expect, which judging from the examples would be `!R`. I couldn't see any examples that use the integer type, so I'm not sure what that ought to be. @Maokami could you perhaps clarify?

alexkeizer (changes requested):
Thanks for your contribution, and sorry for breaking this!

PR #1345, opened 2025-06-20
chore: print number of goals seen for InstCombine

luisacicolini:
This PR adds a print of the total #goals seen in the `InstCombine` benchmark and uncomments the printing of theorems for which a counter example is found by both solvers.


bv-evaluation/collect-data-llvm.py
```diff
@@ -244,8 +244,11 @@
                 print("something is broken: bitwuzla output "+str(output_bw.value)+" leanSAT output "+str(output_ls.value))
                 break
         thmTot += thm
+        # print("file "+file+ " contains "+str(thm)+" goals\n.")
```
tobiasgrosser: Why do we add a commented line?

bv-evaluation/collect-data-llvm.py
```diff
@@ -207,7 +207,7 @@
                 ls_only_failed += 1
                 thm = thm + 1
             elif output_bw == output.counterxample and output_ls == output.counterxample:
-                # print("bitwuzla and leanSAT provided counterexample for theorem "+str(thm)+ "in file "+file)
+                print("bitwuzla and leanSAT provided counterexample for theorem "+str(thm)+ " in file "+file)
```
tobiasgrosser, line 210: Is this not a bit noisy?

tobiasgrosser (approved):

bv-evaluation/collect-data-llvm.py
luisacicolini: I thought having CI printing something for every file might make it too noisy, do we want this print instead?

bv-evaluation/collect-data-llvm.py
luisacicolini: I think it would make sense to have this in CI, as we do for all other cases (I don't know why this one specifically was commented). But if this is noisy I can move it to a different PR

bollu:
I would suggest  to switch to using the logging module. We can use `logging.debug` for debug-print statement, and `logging.info` for information. Setting it up this way gives us timestamped logs, along with an extra log file. Code [available here](https://github.com/opencompl/lean-mlir/blob/main/bv-evaluation/compare-automata-automata-circuit-jsonl.py#L131-L135):

```py
def setup_logging(db_name : str):
    # Set up the logging configuration
    logging.basicConfig(level=logging.DEBUG,
        format='%(asctime)s - %(levelname)s - %(message)s',
        handlers=[logging.FileHandler(f'{db_name}.log', mode='a'), logging.StreamHandler()])
```

bollu:
Let's do it in the following way: This PR will only contain the new `print`, and the next PR will contain `logging.info` and `logging.debug` calls to ensure we get both good debug logging as well as not-too-noisy CI. Does that sound good?

bv-evaluation/collect-data-llvm.py
tobiasgrosser: No, just drop it? Or add an option that one can enable it.

bv-evaluation/collect-data-llvm.py
tobiasgrosser: OK, then let's do this.

tobiasgrosser:
> Let's do it in the following way: This PR will only contain the new `print`, and the next PR will contain `logging.info` and `logging.debug` calls to ensure we get both good debug logging as well as not-too-noisy CI. Does that sound good?

Yes, the logging stuff should be a separate PR.

bv-evaluation/collect-data-llvm.py
luisacicolini: removed

bollu (approved):
LGTM.

luisacicolini:
waiting for ci to finish and pass, will merge afterwards.

PR #1346, opened 2025-06-20
chore: make comparison of HackersDelight results more stable

luisacicolini:
This PR makes the evaluation of HackersDelight problems with different bitwithds more stable and updates the comparison script to have the same structure of the other comparison scripts.


tobiasgrosser (commented):
What does `more stable` mean? Can you explain what changed or was the problem? Also, it seems you now use a symbolic width. Why is that?

With respect to making the script similar to the others, this is great. You may want to consider to factor the common code into a library file and import this code, instead of replicating it across files. This would make it even clearer that the script are the same. Git history wise, first factoring out the functions (in a separate commit) and then using them here would make for an easier-to-read history, as your claim that this is the same code is trivially true as one sees the very same function is being called.

luisacicolini:
the previous version of the code used `sed` to replace e.g. the `64` in `BitVec 64` with another width, which does not seem super robust as it relied on that value alwasy being 64 (or known). Now it's `WIDTH` and we replace it with a concrete value (4, 8, 16, 32, 64) only _when we run the file_. In this way, if we want to try other widths, the replacement is easier (we don't have to look into the file what's the current width value to replace it).

luisacicolini:
I agree with the factoring out functions that are the same and will do so.

luisacicolini:
Actually, if we all agree, I'd rather first merge this PR - pinky promising that I'll immediately factor-out the common functions (with the old script there would be nothing to factor out, as it's too different than all other scripts). Otherwise, the only alternative I can think of, is to have both the factoring out and the edit in this pr, which seems not ideal really.

tobiasgrosser (approved):

PR #1353, opened 2025-06-20
chore: add flag to avoid deep kernel recursion error

luisacicolini:
This PR enables the flag `set_option debug.skipKernelTC true` to avoid deep kernel recursion error _only_ in the theorems from the InstCombine benchmark that suffer from the issue.

### DIFF
--- a/bv-evaluation/collect-data-llvm.py
+++ b/bv-evaluation/collect-data-llvm.py
@@ -187,7 +187,7 @@ for file in os.listdir(benchmark_dir):
                         errs = errs + 1  
                 elif (("error:" in l or "PANIC" in l) and "Lean" not in l and r == 0):
                     err_loc_tot.append(l.split("error: ")[0].split("/")[-1][0:-1])
-                    err_msg_tot.append((l.split("error: ")[1])[0:-1]+", file: "+file+", goal = "+str(thm))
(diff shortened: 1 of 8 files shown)


bv-evaluation/collect-data-llvm.py
```diff
@@ -187,7 +187,7 @@
                         errs = errs + 1  
                 elif (("error:" in l or "PANIC" in l) and "Lean" not in l and r == 0):
                     err_loc_tot.append(l.split("error: ")[0].split("/")[-1][0:-1])
-                    err_msg_tot.append((l.split("error: ")[1])[0:-1])
```
tobiasgrosser: Is this intentional?

tobiasgrosser (approved):

bv-evaluation/collect-data-llvm.py
luisacicolini: fixed, thanks

alexkeizer (changes requested):
Note: when https://github.com/opencompl/lean-mlir/pull/1362 gets merged we should not see these deep recursion errors anymore, so there should no longer be a need for a flag that disables the kernel checking. That is, we probably should not merge this PR.

tobiasgrosser:
Yes, let's close this.

PR #1357, opened 2025-06-23
chore: Prove correctness of new k-induction

bollu:


bollu:
We merge the current status as a stable checkpoint, as the problem has been reduced to graph-theoretic axioms which will be filled in subsequent PRs.

PR #1358, opened 2025-06-23
chore: unify evaluation of `instcombine` and `hackersdelight`

luisacicolini:
This PR deletes:
* all the `Alive`-related evaluation, since it's not in use anymore, and all its folders
* all the symbolic `hackersdelight` and `instcombine` results (unused, deleted from CI as well)


bollu (changes requested):
Let's separate this PR into two PRs: one that drops the un-necessary files, and one that adds the new script? That makes doing a `git revert` in the future massively easier in case we want to recover data.

bollu (approved):
LGTM

luisacicolini:
i removed the symbolic evaluation of hackersdelight from CI too, as it was not used.

luisacicolini:
superceded by #1361 : I would rather implement the changes to ci one step at a time.

PR #1360, opened 2025-06-23
chore: add unified `compare` script

luisacicolini:
Following the comment in #1358, this PR adds the unified `compare` script that merges evaluation of both `instcombine` and `hackersdelight` benchmarks.


bv-evaluation/compare.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    Runs a single Lean file and logs its output.
+    file_to_run: The full path to the .lean file to execute.
+    log_file_base_name: The base name for the log file.
+    specific_arg_for_log_name: An optional argument to append to the log file name, useful for hackersdelight bit-width.
+    """
+    cmd_prefix = 'lake lean '
+    if benchmark == "hackersdelight":
+        try:
+            for r in range(REPS):
```
bollu: Since `REPS = 1`, I think we can safely delete this code.

bv-evaluation/compare.py
```diff
@@ -0,0 +1,164 @@
...
+                        log_file.truncate(0)
+                        log_file.write(f"time out of {TIMEOUT} seconds reached\n")
+                        print(f"{file_to_run} - time out of {TIMEOUT} seconds reached")
+        finally:
+            # Clean up the temporary file created for this specific bit-width and original file
+            if os.path.exists(file_to_run):
+                os.remove(file_to_run)
+                print(f"Deleted temporary file: {file_to_run}")
+
+    elif benchmark == "instcombine":
```
bollu:
```suggestion
    elif benchmark == "instcombine":
```

We should have an `else` branch that throws an error, in case the user passes nothing.

bv-evaluation/compare.py
```diff
@@ -0,0 +1,164 @@
...
+                log_file_path = os.path.join(RESULTS_DIR_HACKERSDELIGHT, log_file_name)
+                with open(log_file_path, 'w') as log_file:
+                    cmd = cmd_prefix + file_to_run
+                    print(f"Running: {cmd}")
+                    try:
+                        subprocess.Popen(cmd, cwd=ROOT_DIR, stdout=log_file, stderr=log_file, shell=True).wait(timeout=TIMEOUT)
+                    except subprocess.TimeoutExpired:
+                        log_file.truncate(0)
+                        log_file.write(f"time out of {TIMEOUT} seconds reached\n")
+                        print(f"{file_to_run} - time out of {TIMEOUT} seconds reached")
```
bollu: I think the `try/except` logic is the exact same, so we should extract these as well into a single function?

bv-evaluation/compare.py
```diff
@@ -0,0 +1,164 @@
...
+                    with open(temp_file_path, 'w', encoding='utf-8') as temp_file:
+                        temp_file.write(modified_content)
+                    
+                    # Submit the temporary file to be run.
+                    # log_file_base_name will be 'original_file_base'
+                    # specific_arg_for_log_name will be the 'width'
+                    future = executor.submit(run_file, "hackersdelight", temp_file_path, original_file_base, str(width))
+                    futures[future] = temp_filename # Store the name of the temporary file for progress reporting
+
+        elif benchmark == "instcombine":
```
bollu, line 130: We should have an `else` case that throws an error.

bv-evaluation/compare.py
```diff
@@ -0,0 +1,164 @@
...
+                shutil.rmtree(item_path)
+        except Exception as e:
+            print(f'Failed to delete {item_path}. Reason: {e}')
+
+def run_file(benchmark : str, file_to_run : str, log_file_base_name : str, specific_arg_for_log_name: str = None) :
+    """
+    Runs a single Lean file and logs its output.
+    file_to_run: The full path to the .lean file to execute.
+    log_file_base_name: The base name for the log file.
+    specific_arg_for_log_name: An optional argument to append to the log file name, useful for hackersdelight bit-width.
```
bollu: Why is `specific_arg_for_log_name` useful for hackersdelight? An explanaion would be nice!

bollu (changes requested):
Nice thank you for unifying the scripts! I feel there's more unification that's possible, so I request these as changes.

bv-evaluation/compare.py
luisacicolini: done

bv-evaluation/compare.py
luisacicolini: done

bv-evaluation/compare.py
luisacicolini: done

bv-evaluation/compare.py
luisacicolini: done

bv-evaluation/compare.py
luisacicolini: explained and renamed

bv-evaluation/compare.py
```diff
@@ -0,0 +1,152 @@
...
+    # Construct log file name including original file's base name and bit-width
+    with open(log_file_path, 'w') as log_file:
+        cmd = cmd_prefix + file_to_run
+        print(f"Running: {cmd}")
+        try:
+            subprocess.Popen(cmd, cwd=ROOT_DIR, stdout=log_file, stderr=log_file, shell=True).wait(timeout=TIMEOUT)
+        except subprocess.TimeoutExpired:
+                    log_file.truncate(0)
+                    log_file.write(f"time out of {TIMEOUT} seconds reached\n")
+                    print(f"{file_to_run} - time out of {TIMEOUT} seconds reached")
```
bollu:
```suggestion
        except subprocess.TimeoutExpired:
                log_file.truncate(0)
                log_file.write(f"time out of {TIMEOUT} seconds reached\n")
                print(f"{file_to_run} - time out of {TIMEOUT} seconds reached")
```

bv-evaluation/compare.py
```diff
@@ -0,0 +1,152 @@
...
+
+    if (benchmark == "hackersdelight"):
+        # Clean up the temporary file created for this specific bit-width and original file for hackersdelight
+        if os.path.exists(file_to_run):
+            os.remove(file_to_run)
+            print(f"Deleted temporary file: {file_to_run}")
```
bollu: This code should be moved to `compare`, so that `run_file` is a real library utility that doesn't know anything about which benchmark is calling it.

bv-evaluation/compare.py
```diff
@@ -0,0 +1,152 @@
...
+            clear_folder(RESULTS_DIR_INSTCOMBINE)
+            os.makedirs(RESULTS_DIR_INSTCOMBINE, exist_ok=True)
+
+            for file in os.listdir(BENCHMARK_DIR_INSTCOMBINE):
+                if "_proof" in file and file.endswith('.lean'): # Ensure it's a Lean file
+                    file_path = os.path.join(BENCHMARK_DIR_INSTCOMBINE, file)
+                    file_title = os.path.splitext(file)[0]
+                    log_file_path = os.path.join(RESULTS_DIR_INSTCOMBINE, f'{file_title}.txt')
+                    future = executor.submit(run_file, "instcombine", file_path, log_file_path)
+                    futures[future] = file
```
bollu:
```suggestion
                    futures[future] = file_path
```

Let's associate the future with the full path so we can print an error.

bv-evaluation/compare.py
```diff
@@ -0,0 +1,152 @@
...
+                    with open(temp_file_path, 'w', encoding='utf-8') as temp_file:
+                        temp_file.write(modified_content)
+                    
+                    # Submit the temporary file to be run.
+                    # log_file_base_name will be 'original_file_base'
+                    # specific_arg_for_log_name will be the 'width'
+                    log_file_name = f'{original_file_base}_{str(width)}.txt'
+                    log_file_path = os.path.join(RESULTS_DIR_HACKERSDELIGHT, log_file_name)
+                    future = executor.submit(run_file, "hackersdelight", temp_file_path, log_file_path)
+                    futures[future] = temp_filename # Store the name of the temporary file for progress reporting
```
bollu:
```suggestion
                    futures[future] = temp_file_path # Store the name of the temporary file for progress reporting
```

Let's associate the future with the full path so we can print an error.

bv-evaluation/compare.py
```diff
@@ -0,0 +1,152 @@
...
+
+        total = len(futures)
+        for idx, future in enumerate(concurrent.futures.as_completed(futures)):
+            file_or_name = futures[future]
+            try:
+                future.result()
+            except Exception as exc:
+                print(f'{file_or_name} generated an exception: {exc}')
+            percentage = ((idx + 1) / total) * 100
+            print(f'{file_or_name} completed, {percentage:.2f}%')
```
bollu:
```suggestion
            file_path = futures[future]
            try:
                future.result()
            except Exception as exc:
                print(f'{file_path} generated an exception: {exc}')
            percentage = ((idx + 1) / total) * 100
            print(f'{file_path} completed, {percentage:.2f}%')
```

bollu (changes requested):
We're getting there! Thank you for working on the onerous task of evaluation scripting ^_^

bollu:
As a final cleanup, it would be great to run `ruff format collect.py` and `ruff check collect.py`: [Ruff](https://docs.astral.sh/ruff/installation/) is a python formatted and linter that catches a bunch of non-obvious mistakes, so it's nice to use best practices and make sure that our scripts are as nice as possible. 

I am happy to do this myself in a follow-up PR.

bollu (approved):
LGTM. Let's merge this in, and `ruff` the code in a follow up PR.

luisacicolini:
thank you @bollu, `ruff` checked and formatted. will merge once CI passes.

PR #1362, opened 2025-06-24
fix: implement workaround for kernel deep recursion issue

alexkeizer:
This PR implements a workaround to avoid the kernel deep recursion issue we've been seeing in the evaluation (https://github.com/leanprover/lean4/issues/8898).

Concretely, we add simprocs which match for `BitVec.ofInt w x` or `LLVM.const? w x`, with `x` a constant (e.g.., literals like 65000 or -1), and rewrites them to `hide (BitVec.ofInt w x)` (resp, `hide (LLVM.const? w x)`). Here, `hide` is a new opaque definition which is just the identity function (and for which we have a proof that it is the identity function), but because it is opaque it prevents the kernel from trying to reduce the argument.
The workaround is encapsulated in two new tactics: `hide_constant` (which applies aforementioned simprocs), and `unhide_constants` (which eliminates all `hide` applications).

It seems that both `simp_alive_undef` and `simp_alive_ops` can trigger the deep recursion, so we call `hide_constant` at the start of the former, and `unhide_constant` at the end of the latter. This does mean that any proof which only uses the former will break---but these tactics aren't really meant to be used in isolation to begin with so it shouldn't cause too many problems.

I ran the evaluation locally, with this workaround and confirmed no more deep recursion issues, and all proofs still seem to work as expected:
```
In total we found 7978 goals.
leanSAT and Bitwuzla solved: 7920
leanSAT and Bitwuzla provided 8 counterexamples
leanSAT and Bitwuzla both failed on 25 theorems
leanSAT failed and Bitwuzla succeeded on 25 theorems
leanSAT succeeded and Bitwuzla failed on 0 theorems
There were 0 inconsistencies
Errors raised: 75
ran rg 'Bitwuzla failed' | wc -l, expected 25, found 25, SUCCESS
ran rg 'LeanSAT failed' | wc -l, expected 50, found 50, SUCCESS
ran rg 'LeanSAT provided a counter' | wc -l, expected 8, found 8, SUCCESS
ran rg 'Bitwuzla provided a counter' | wc -l, expected 8, found 8, SUCCESS
ran rg 'LeanSAT proved' | wc -l, expected 7920, found 7920, SUCCESS
ran rg 'Bitwuzla proved' | wc -l, expected 7945, found 7945, SUCCESS
error The SAT solver timed out while solving the problem. was raised 50 times
error The SMT solver timed out while solving the problem. was raised 25 times
```

Finally, I tweaked the shell.nix file to include some missing python dependencies, and remove vscode. The latter was tracking the "latest" version of the Lean extension, meaning it would trigger a SHA mismatch whenever the extension got an update. In any case, I'd much rather use my global vscode config, so it seems an anti-pattern to have vscode in shell.nix. I made these changes as the previous version broke (and I believe I'm the only one currently using nix), but I'm happy to pull this out into a separate PR if it is at all controversial.

### DIFF
--- a/SSA/Projects/InstCombine/Tactic/SimpLLVM.lean
+++ b/SSA/Projects/InstCombine/Tactic/SimpLLVM.lean
@@ -1,6 +1,7 @@
 /-
 Released under Apache 2.0 license as described in the file LICENSE.
 -/
+import SSA.Core.Framework.Trace
 import SSA.Projects.InstCombine.ForLean
 import SSA.Projects.InstCombine.LLVM.EDSL
 
@@ -111,7 +112,7 @@ The following section defined two tactics `hide_constants` and `unhide_constants
 
 This serves as a workaround for https://github.com/leanprover/lean4/issues/8898
 -/
-section HideConstants
+namespace HideConstants
 open Meta
 
 /-- Definition of `hide` and `hide_eq` -/
@@ -136,6 +137,7 @@ def simpHide (e : Expr) : Meta.SimpM Meta.Simp.Step := do
   let ctx ← Simp.getContext
   if let some parent := ctx.parent? then
     if parent.isAppOf ``hide then
+      trace[LeanMLIR.Elab] "{Lean.crossEmoji}: parent ({parent}) is an application of `hide`"
       return .continue
 
   let expr ← Meta.mkAppM ``hide #[e]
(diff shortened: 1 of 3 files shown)


SSA/Projects/InstCombine/Tactic/SimpLLVM.lean
```diff
@@ -1,6 +1,8 @@
 /-
 Released under Apache 2.0 license as described in the file LICENSE.
 -/
+import SSA.Core.Framework.Trace
+import SSA.Core.Framework.Trace
```
tobiasgrosser: Double import?

SSA/Projects/InstCombine/Tactic/SimpLLVM.lean
```diff
@@ -102,11 +104,94 @@ macro_rules
     )
   )
 
+/-! ### Constant Hiding Workaround
+The following section defined two tactics `hide_constants` and `unhide_constants`.
```
tobiasgrosser:
```suggestion
The following section defines two tactics, `hide_constants` and `unhide_constants`.
```

SSA/Projects/InstCombine/Tactic/SimpLLVM.lean
```diff
@@ -102,11 +104,94 @@ macro_rules
...
+By making it opaque, we prevent the kernel from trying to reduce the argument.
+-/
+def hide : α → α := hide_def.1
+
+/-- A proof that `hide` is the identity. Note that this is deliberately *not*
+a def-eq, since we made `hide` opaque. -/
+theorem hide_eq {a : α} : hide a = a := by
+  rw [hide, hide_def.2]; rfl
+
+/-- symmetric version of `hide_eq` -/
```
tobiasgrosser:
```suggestion
/-- Symmetric version of `hide_eq`. -/
```

tobiasgrosser:
The `nix` change sounds fine, but please pull this out into a separate PR.

PR #1363, opened 2025-06-24
fix: add python dependencies to shell.nix

alexkeizer:
This PR adds matplotlib and pandas to shell.nix, as these are required to run the bv-evaluation scripts, but were not being installed yet.

Also, we remove the vscode package, as it doesn't track a specific version, meaning it continually triggers SHA mismatches. Furthermore, I would prefer to just use my own vscode, that is globally installed, so I don't think vscode ought to be part of the shell.nix file.


tobiasgrosser:
Thank you, @alexkeizer!

PR #1366, opened 2025-06-25
feat: weekly GH action to gc toolchains

alexkeizer:
This PR adds a Github action which will run `elan toolchain gc --delete` weekly on the self-hosted evaluation VM, in an effort to reclaim disk space.

NOTE: I'm not super familiar with the GH action syntax, so please do check my work carefully to make sure I haven't missed anything significant :sweat_smile:


alexkeizer:
Tobias gave an LGTM in-person, so merging

PR #1367, opened 2025-06-25
feat: check in CI that nix-shell has the right python depencies

alexkeizer:
This PR adds a new action to CI which will confirm that (a) nix-shell is functional (guarding against any SHAs going out of date, as has happened previously), and (b) that shell.nix has all python dependencies required to run the bv-evaluation scripts.

Furthermore, despite #1363 shell.nix wasn't quite up to date yet. So this PR also adds `pandas` and `ripgrep` to `shell.nix`.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,22 @@
+name: Nix ci
+on:
+  push:
+    branches:
+      - "main"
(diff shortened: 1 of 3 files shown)


tobiasgrosser:
Nice. Thank you!

PR #1368, opened 2025-06-25
feat: enable multiple repetitions for evaluation of `instcombine` and `hackersdelight`

luisacicolini:
this pr re-introduces the infrastructure to repeat the tests `REPS` times (with `REPS` given as user input in `compare.py`) and compute final solving time returned as a result as a mean (`np.mean`) of these runs, for each goal solved and for each file.


tobiasgrosser:
Yes, I think this should be an argument to the compare script. Also, why do you delete `.placeholder`. This seems unrelated.

bv-evaluation/compare.py
```diff
@@ -24,7 +24,6 @@
 
 TIMEOUT = 1800
 
-
 def clear_folder(results_dir):
```
tobiasgrosser: Unrelated change. Can you drop this change?

bv-evaluation/results/HackersDelight/.plaeholder
tobiasgrosser, line 1: Unrelated change. Please drop.

tobiasgrosser (approved):
This LGTM. Good to merge, modulo the minor comments.

bv-evaluation/compare.py
luisacicolini:
```suggestion

def clear_folder(results_dir):
```

luisacicolini: dropped

bv-evaluation/results/HackersDelight/.plaeholder
luisacicolini: fixed

PR #1369, opened 2025-06-26
feat: add scripts and Dockerfile commands to run SMT-LIB benchmarks

abdoo8080:
This WIP PR sets up the initial SMT-LIB benchmarking framework. It installs the relevant solvers and introduces `results/run_smtlib_benchmarks.py`, which executes benchmarks in parallel and stores results under `bv-evaluation/results`. TODO:
- Develop scripts to aggregate benchmark data into CSV format.
- Integrate with plotting and statistics scripts.
- Refactor `results/run_smtlib_benchmarks.py` to:
  - Reorganize output paths for clarity.
  - Enable shared task pools across solvers when possible (low priority).

### DIFF
--- a/artifacts/oopsla25-bv-decide/Dockerfile
+++ b/artifacts/oopsla25-bv-decide/Dockerfile
@@ -28,34 +28,98 @@ RUN apt-get update && \
 RUN useradd -m -s /bin/bash user && \
     adduser user sudo && \
     echo "user ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
-RUN mkdir -p /code && chown user:user /code # Create /code as root before switching
-
 USER user
+
+# HACK around Lean-MLIR self-reference incompatibility with docker cache:
+# Do all the work in the home directory for now and move it to the lean-mlir directory later
+RUN mkdir -p /home/user/lean-mlir
+WORKDIR /home/user/lean-mlir
 
 # Install elan and update environment
 RUN curl https://elan.lean-lang.org/elan-init.sh -sSf | sh -s -- -y --default-toolchain none
 ENV PATH=/home/user/.elan/bin:$PATH
 
+# Download and unzip SMT-LIB QF_BV benchmarks
+RUN mkdir -p bv-evaluation/SMT-LIB
+RUN curl -L -o QF_BV.tar.zst https://zenodo.org/records/15493090/files/QF_BV.tar.zst?download=1 && \
+    tar -xf QF_BV.tar.zst -C bv-evaluation/SMT-LIB && \
+    rm QF_BV.tar.zst
+
+# Create Solvers directory
+RUN mkdir -p bv-evaluation/solvers
 
 # Install Bitwuzla (commit: 22.6.25)
-RUN cd /home/user && git clone https://github.com/bitwuzla/bitwuzla
-# && git checkout 0dca38d0f62fa9002ad6278ca6374838a69ade19
-WORKDIR /home/user/bitwuzla
-RUN ./configure.py && cd build && ninja && sudo ninja install
+RUN cd bv-evaluation/solvers && git clone https://github.com/bitwuzla/bitwuzla && cd bitwuzla && git checkout 0dca38d0f62fa9002ad6278ca6374838a69ade19
+RUN cd bv-evaluation/solvers/bitwuzla && ./configure.py && cd build && ninja && sudo ninja install
 
+# Install CoqQFBV (commit: 09.04.23)
+RUN cd bv-evaluation/solvers && git clone https://github.com/fmlab-iis/coq-qfbv && cd coq-qfbv && git checkout bb055f7b05dd81a5ddf89c734bde140a9feecf70
+WORKDIR bv-evaluation/solvers/coq-qfbv
+RUN yes "/usr/local/bin" | bash -c "sh <(curl -fsSL https://opam.ocaml.org/install.sh)" && \
+    opam init --disable-sandboxing --bare
+RUN opam switch create ocaml-base-compiler.4.14.0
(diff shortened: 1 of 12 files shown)


artifacts/oopsla25-bv-decide/Dockerfile
```diff
@@ -1,55 +1,124 @@
...
+
+# Set up non-interactive mode for apt-get
+ENV DEBIAN_FRONTEND=noninteractive
+ENV TZ=Etc/UTC
 
 # install dependencies
-RUN apt-get update -y \
-    && DEBIAN_FRONTEND="noninteractive" apt-get install --yes \
+RUN apt-get update && \
+    apt-get install --yes \
```
alexkeizer:
```suggestion
RUN apt-get update -y && \
    apt-get install --yes \
```
Was the loss of the `-y` flag intentional?

artifacts/oopsla25-bv-decide/Dockerfile
```diff
@@ -1,55 +1,124 @@
...
         tzdata \
-	pip \
-	python3-venv \
+        pip \
+        python3-venv \
```
alexkeizer: What happened with the whitespace here?

artifacts/oopsla25-bv-decide/Dockerfile
```diff
@@ -1,55 +1,124 @@
...
+# Install elan and update environment
+RUN curl https://elan.lean-lang.org/elan-init.sh -sSf | sh -s -- -y --default-toolchain none
+ENV PATH=/home/user/.elan/bin:$PATH
 
-# install Lean-MLIR (commit: 22.06.25)
-RUN git clone https://github.com/opencompl/lean-mlir && cd lean-mlir && git checkout b0253c5f5dbc6cc507b559d76fd659da38a8e565 
-WORKDIR /code/lean-mlir
+# Install Lean-MLIR (commit: 22.06.25)
+RUN cd /home/user && git clone https://github.com/opencompl/lean-mlir && cd lean-mlir && git checkout b0253c5f5dbc6cc507b559d76fd659da38a8e565 
+WORKDIR /home/user/lean-mlir
```
alexkeizer: A bunch of seemingly unrelated changes here, why:
* did we re-order the various installation bits (e.g., where did the bitwuzla install go)?
* was the URL we install elan from changed?
* did the lean-mlir checkout move from `/code/` to `/home/user`?

For the last one: note that this is reflected in the instructions PDF also, so we'd have to update that as well

alexkeizer (changes requested):
I only looked through the Dockerfile, but the diff there seems very messy. Could you keep the old stuff in the same place (instead of moving things around), so that it's clearer what exactly changed?

artifacts/oopsla25-bv-decide/Dockerfile
luisacicolini: @alexkeizer a bunch of these changes may be due to my last updates to the dockerfile on main, which I merged this morning! Sorry about that, @abdoo8080

bv-evaluation/collect-stats-bv-decide.py
```diff
@@ -416,22 +417,19 @@ def slowdown_smtlib_stats():
 
 
 def smtlib_stats(performance_smtlib_dir):
-    df_bitwuzla_46k = pd.read_csv("../raw-data/SMTLIB/bitwuzla_46k.csv")
-    df_leanwuzla_46k = pd.read_csv("../raw-data/SMTLIB/leanwuzla_46k.csv")
-    assert len(df_bitwuzla_46k) == len(df_leanwuzla_46k)
+    df_bitwuzla = pd.read_csv("../raw-data/SMT-LIB/bitwuzla_data.csv")
```
luisacicolini:
```suggestion
    df_bitwuzla = pd.read_csv("../raw-data/SMT-LIB/bitwuzla_data.csv")
```
sorry just a double check - `raw-data` should now be at the same level of the `collect-stats-script`, is the `../` necessary?

abdoo8080: No. I removed `../` in my local workspace. I didn't know you already fixed this issue (~8 mins ago)!

abdoo8080: Resolved!

luisacicolini: thanks :)

artifacts/oopsla25-bv-decide/Dockerfile
abdoo8080: Yes!

artifacts/oopsla25-bv-decide/Dockerfile
abdoo8080: Caused by https://github.com/opencompl/lean-mlir/pull/1369#discussion_r2171968621. Resolved!

artifacts/oopsla25-bv-decide/Dockerfile
abdoo8080: > did we re-order the various installation bits (e.g., where did the bitwuzla install go)?

I moved bitwuzla to `/home/user/lean-mlir/bv-evaluation/solvers` with other solvers. As for `lean-mlir`, I am pushed it to the end to avoid rebuilding 90% of the docker image every time we make a change to a script (see note on self-reference hack).

> was the URL we install elan from changed?

No. We used to install a default toolchain, which increases build time and size of the docker image. I changed to the code to install only the toolchain(s) needed to build lake projects

> did the lean-mlir checkout move from `/code/` to `/home/user`?

Yes. I moved everything to `/home/user` for simplicity (and not having to deal with root issues)

> For the last one: note that this is reflected in the instructions PDF also, so we'd have to update that as well

Good point! Will update!

PR #1371, opened 2025-06-26
fix: tweak `hide_constants` to avoid exponentiation threshold

alexkeizer:
This PR tweaks the hide_constants so that `LLVM.const? w x` is now rewritten to `hide (LLVM.const? (hide w) x`, instead of `hide (LLVM.const? _ _)`. This was not necessary to avoid the original deep recursion errors, but, in some examples it seems that `2 ^ w` was still being evaluated, leading to errors talking about the exponentiation threshold. (NOTE: these did not seem to be kernel errors, rather this seemed to be happening in the elaborator). By explicitly hiding the `w` value, the problematic expression becomes `2 ^ (hide w)`, which Lean will not attempt to evaluate, silencing the aforementioned exponentation threshold error.

We did have to insert a cast, as the return type now changed from `IntW w` to `IntW (hide w)`, so we had to tweak unhide_constants as well to properly eliminate this cast. Furthermore, this PR removes the check whether `x` is a "constant", instead choosing to hide any expression. There is no real cost to doing the rewrite, so this isConstant check was unneccessary complexity, which would've been even more complex to maintain after this change.

### DIFF
--- a/SSA/Projects/InstCombine/AliveHandwrittenLargeExamples.lean
+++ b/SSA/Projects/InstCombine/AliveHandwrittenLargeExamples.lean
@@ -504,7 +504,9 @@ def alive_simplifySelect764 (w : Nat) :
   simp_alive_ops
   simp_alive_case_bash
   intro A
-  simp only [BitVec.slt, ofBool_eq_one, toInt_zero, decide_eq_true_eq, BitVec.zero_sub]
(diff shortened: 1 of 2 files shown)


tobiasgrosser:
I close this as this was resolved in https://github.com/opencompl/lean-mlir/pull/1379. @alexkeizer, please reopen if there is still a need for this PR.

PR #1381, opened 2025-06-26
chore: `hackersdelight` collect fix

luisacicolini:
this pr fixes a mistake with the data collection script for `hackersdelight`, which I had not previously noticed. Will merge immediately, as this is a very limited fix of 2 lines only on the concerned file.


tobiasgrosser:
Thank you.

PR #1384, opened 2025-06-26
feat: add `plot` scripts

luisacicolini:
This PR moves the plotting script and its corresponding `requirements.txt` file from the paper repository to `lean-mlir` repository to enable plotting within the artifact. I also `ruff` formatted and checked the files, removing uncalled functions.

PR #1385, opened 2025-06-26
chore: update `Dockerfile` for `bv_decide` artifact

luisacicolini:
Updated the dockerfile to run with the scripts currently in main. The plotting scripts are still missing (need to merge #1384 first), but at least the docker should be functioning and running things. Also updated the `makefile` to run the correct commands.

### DIFF
--- a/artifacts/oopsla25-bv-decide/Dockerfile
+++ b/artifacts/oopsla25-bv-decide/Dockerfile
@@ -61,10 +61,9 @@ RUN rm -rf .lake/packages/mathlib && lake update && lake build
 RUN python3 -m venv .venv && \
     .venv/bin/pip install -r requirements.txt
 
-# setup python env
+# Setup python env
 ENV VIRTUAL_ENV=/opt/venv
 ENV PATH="/opt/venv/bin:$PATH"
-
 
 # Set the evaluation script directory as working dir
 WORKDIR /code/lean-mlir/bv-evaluation/
--- a/artifacts/oopsla25-bv-decide/Dockerfile
+++ b/artifacts/oopsla25-bv-decide/Dockerfile
@@ -49,11 +49,6 @@ RUN cd /code && git clone https://github.com/opencompl/lean-mlir && cd lean-mlir
 #&& git checkout 7bd8c716a4a0bd5608279f0b73bd6c87f78c9f1c
 WORKDIR /code/lean-mlir
 
-# Install and set the default Lean toolchain for elan
-# Get the toolchain from .lean-toolchain file (example: 'leanprover/lean4:nightly-2025-06-20')
-RUN /home/user/.elan/bin/elan toolchain install nightly && \
-    /home/user/.elan/bin/elan default nightly
-
 # Ensure the .lake/packages directory is clean before running lake update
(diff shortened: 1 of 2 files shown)


luisacicolini:
at the moment, the `Dockerfile` fails to run the `compare` script (`file not found`). no idea why, but it works with a previous `checkout` of `lean-mlir`

luisacicolini:
that error seems to have disappeared, at least on the lean machine and dockerfile does what it's supposed to without throwing errors 🥳 
I am very unsure about the whole versioning of lean/lean-mlir/mathlib: I will need some help with that.

artifacts/oopsla25-bv-decide/Dockerfile
```diff
@@ -1,55 +1,79 @@
...
-# build lean-mlir
-RUN lake exe cache get
-RUN lake build
+# Install and set the default Lean toolchain for elan
+# Get the toolchain from .lean-toolchain file (example: 'leanprover/lean4:nightly-2025-06-20')
+RUN /home/user/.elan/bin/elan toolchain install nightly && \
+    /home/user/.elan/bin/elan default nightly
+
+# Ensure the .lake/packages directory is clean before running lake update
+RUN rm -rf .lake/packages/mathlib && lake update && lake build
```
abdoo8080: > I am very unsure about the whole versioning of lean/lean-mlir/mathlib: I will need some help with that.

`lake` should automatically figure out the correct Lean toolchain version to use in a clean build with `lake update && lake build`. So, you don't need lines 52-56. I don't know what happens when you start with an old build (which I am guessing you do from `rm -rf .lake/packages/mathlib`).

luisacicolini: I see, the problem is that without those lines kept getting an error in the docker, which was unable to find the `lean-toolchain` or `lakefile`

abdoo8080: Oh! That's interesting... It just works for me...

luisacicolini: let me give it another try

luisacicolini: @abdoo8080 seems to have worked now! Guess the problem was somewhere else yesterday. Finishing the run and then I'll update the dockerfile in this pr.

tobiasgrosser (approved):

artifacts/oopsla25-bv-decide/Dockerfile
```diff
@@ -1,55 +1,79 @@
...
-COPY . .
 
-# build lean-mlir
-RUN lake exe cache get
-RUN lake build
+# Install and set the default Lean toolchain for elan
+# Get the toolchain from .lean-toolchain file (example: 'leanprover/lean4:nightly-2025-06-20')
+RUN /home/user/.elan/bin/elan toolchain install nightly && \
+    /home/user/.elan/bin/elan default nightly
+
```
luisacicolini:
```suggestion
```

artifacts/oopsla25-bv-decide/Dockerfile
luisacicolini: removed.

PR #1386, opened 2025-06-26
chore: update and move `collect-stats`

luisacicolini:
This PR updates the `collect-stats` script and its use in CI. 

Changelist: 
* ran `ruff check` and `ruff format` on the `stats` script, removed unused variables and reformatted, renamed script
* updated CI with the new script name 
* added `tables` subfolder to store the output of the script
* Given that we won't eventually need the `for-paper` subdir (everything will live in `raw-data`, `plots`, `tables` subdirs for the artifact), I moved the script out and updated it to save the tables in `tables`. I also `ruff` formatted and checked the script.
* updated `requirements.txt` with the package the script needs


tobiasgrosser:
@luisacicolini, this generally seems to go in a good direction. However, can you change the commit message from "this updates the script" to actually talk about the changes you made.

abdoo8080:
Any reason why this script is different from [collect-stats.py](https://github.com/opencompl/paper-lean-bitvectors/blob/main/tools/collect-stats.py) in paper repo?

luisacicolini:
@abdoo8080 I removed functions that were not used and ran `ruff` check + format, happy to leave the old script (I still suggest we move it, though)

tobiasgrosser:
> @abdoo8080 I removed functions that were not used and ran `ruff` check + format, happy to leave the old script (I still suggest we move it, though)

Let's not keep old stuff around. We should have a single source of truth. I vote for moving.

abdoo8080:
I see! Some changes I made there are not carried over. I will fix that once this PR is merged.

luisacicolini:
yes @abdoo8080, feel also free to push to this branch if that works for you.

bollu:
LGTM, and thanks for consistently running a linter on this stuff.

PR #1387, opened 2025-06-27
feat: unify data collection scripts

luisacicolini:
This PR unifies the data collection scripts for `instcombine` and `hackersdelight`. 
changelist: 
* remove old `data-collection-llvm.py`, `data-collection-hackersdelight.py`
* add `collect.py`
* use the updated scripts in CI


bollu (commented):
LGTM, modulo the fact that the script has not been reviewed carefully, as it is quite large.

PR #1392, opened 2025-06-27
feat: add zenodo link for bv-decide paper

luisacicolini:
change list: 
- [x] updated zenodo link in `artifact.tex`
- [x] removed python scripts running when the docker container is built, scripts can now be ran via `run.sh` from within the artifact folder
- [x] update `makefile` to build the docker image and create a `tar.ball` of the built image
- [x] update `artifact.tex` with instructions on using the artifact, including the new commands in `run.sh`

everything was paired on with/supervised by @bollu


artifacts/oopsla25-bv-decide/makefile
```diff
@@ -4,10 +4,8 @@ artifact.pdf: artifact.tex sigplanconf.cls
 	latexmk -pdf -shell-escape artifact.tex
 
 docker:
-	docker build -t lean-image
-	docker create --name temp-container lean-image
-	docker cp temp-container:/code/lean-mlir/bv-evaluation/raw-data ./docker-raw-results
-	docker rm temp-container
+	podman build -t oopsla25-bv-decide --progress plain .
+	podman save -o ./oopsla25-bv-decide.tar oopsla25-bv-decide
```
alexkeizer, line 8: Why the switch from podman to docker? The latter seems more standard: even if the user has podman installed, it's common to provide a docker-compatible alias, but I haven't heard of docker providing a podman alias.

PR #1398, opened 2025-06-27
feat: improve readability of `collect.py`

luisacicolini:
as suggested by @bollu, this PR simplifies some of the logic in `collect.py` making it more readable.


bollu:
LGTM, as this seems much clearer. I prefer that while there is code repeated, it is very easy to audit. The previous version was non-obvious code.

luisacicolini:
will merge once CI passes and I've made sure that the numbers are still consistent.

PR #1401, opened 2025-06-27
chore: fix directories and dataframe names in `collect-stats-bv-decide`

luisacicolini:
This pr updates `collect-stats-bv-decide` to work with the new dataframes. co-authored/supervised by @bollu


bollu (approved):
LGTM.

PR #1404, opened 2025-06-27
chore update artifact

luisacicolini:


artifacts/oopsla25-bv-decide/artifact.tex
```diff
@@ -12,14 +12,14 @@
 % CC BY 4.0 license
 %
 
-\documentclass[onecolumn, sigconf]{acmart}
+\documentclass[acmlarge, nonacm]{acmart}
```
bollu, line 15: why did this change?

artifacts/oopsla25-bv-decide/artifact.tex
```diff
@@ -102,63 +110,64 @@ \subsection{Artifact check-list (meta-information)}
...
-Access the tar file of the docker image from \url{https://doi.org/10.5281/zenodo.15755236}
+\section{Getting Started}
 
+Access the docker image from \url{10.5281/zenodo.15755236}: 
 \begin{script}
 $ docker load -i oopsla25-bv-decide.tar
 $ docker run -it oopsla25-bv-decide
-# | This clears the build cache,
-# | fetches the maths library from the build cache,
-# | and builds our framework.
```
bollu: why did we remove the explanation?

artifacts/oopsla25-bv-decide/artifact.tex
luisacicolini: copied from width-indep artifact

bollu (commented):
Thanks! I think we should also add the text about grabbing the docker image from DockerHub if the zenodo image does not exist.

artifacts/oopsla25-bv-decide/artifact.tex
luisacicolini: my mistake, we should put it back

bollu:
I'd say something like:

> Please access the image from DockerHub (https://hub.docker.com/r/abdoo8080/oopsla25-bv-decide-base/) with the command `docker pull abdoo8080/oopsla25-bv-decide-base:v1`). A long-term archive of the Dockher hub image will be uploaded to Zenodo, but we support dockerhub as the primary form of access"

luisacicolini:
lgtm, thank you!

PR #1408, opened 2025-06-28
chore: use --platform to force use x86_64 even on MacOS with ARM

bollu:


artifacts/oopsla25-bv-decide/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:24.04
+FROM --platform=linux/amd64 ubuntu:25.04
```
luisacicolini:
```suggestion
FROM --platform=linux/amd64
```

PR #1410, opened 2025-06-28
feat: include smtlib benchmarking in run.sh

abdoo8080:
This PR also includes changes in `compare.py` to automatically select random benchmarks to run (given a random seed provided by the user).


bollu:
LGTM, thanks for picking up the right `MLTon` from source.

PR #1415, opened 2025-06-30
chore: enable native compilation

ineol:


tobiasgrosser:
Updated to latest `main` to check if `native_compilation` works again.

PR #1422, opened 2025-07-01
feat: add test file for the run

bollu:
Also add `lake exe cache get` to the instrctions so that users actually grab from a lean cache.


artifacts/oopsla25-bv-decide/artifact.tex
```diff
@@ -137,7 +137,7 @@ \section{Getting Started}
 # | This clears the build cache,
 # | fetches the maths library from the build cache,
 # | and builds our framework.
-$ lake update && lake build
+$ lake update && lake exe cache get && lake build
```
bollu:
```suggestion
$ lake update && lake build
```

PR #1433, opened 2025-07-03
refactor: replace shell.nix with flake, build nix-based Docker image in ci

bollu:
This PR replaces our existing `shell.nix` setup with a `flake.nix` file, which lets us lock our nixpkgs version, meaning less flakiness in the setup (no pun intended); this means that to enter the dev shell, one now uses `nix develop` instead of `nix-shell`. 

Additionally, the flake also exposes a default package which transitively installs all dependencies loaded by the devshell, which makes for an easy install target for, say, nix profile. We use this to create a new nix-based Dockerfile, which builds Lean-MLIR using nix profile-installed dependencies. Furthermore, we add CI to automatically build an Docker image, and push it to `ghcr.io/opencompl/lean-mlir` (Github container registry, which makes the image show up in: https://github.com/opencompl/lean-mlir/pkgs/container/lean-mlir). In future PRs, I would like to run our evaluation CI inside this Docker image, to automatically test it, meaning we have a known-good Docker setup for future artifacts.

A noticable shortcoming of the current Dockerfile is that it does not yet cache the .lake folder between builds, making those builds slower than the existing CI build infrastructure. This will be addressed in a follow-up PR.

---

PR description written by @alexkeizer, who adopted the PR


alexkeizer:
> Make nix run isolated from the rest of the system, so that when I perform a run, I really am testing an isolated system,

How do we achieve this? And does it affect `nix develop` also? If we move to a flake, I'd like it to fully replace the shell.nix setup we have today, which means I'd like to use it for daily development, where I don't necessarily want to completely isolate it

alexkeizer:
@bollu I noticed `flake.lock` is .gitignored. Do we have a strong reason for this? I'd expect we want this to be checked in, so that we use the same lock locally as will be backed into the Docker image

bollu:
@alexkeizer what's the status here?

alexkeizer:
@bollu I had a look, and the the bitwuzla version in latest nixpkgs is 0.7.0, which is one release behind the latest (0.8.0), so I decided to override the version in our flake. This does increase the complexity of the setup, not sure how that affects the discussion we were having before.

I think regardless it'd be good to have bitwuzla in my devshell be that version, but I can concede that we might not want to introduce such overlays/overrides into the critical path for artifacts (I *think* it's fine, but will gladly concede if you disagree). In which case, we just install bitwuzla from source in the Dockerfile as we already do today. On the other hand, the fact that we can relatively easily add an override for a dependency which is already packaged in nixpkgs to a new version (assuming that the build-process and dependencies didn't change) might be a good argument for using nix here.

I've also added some chonky comments to flake.nix to make the file slightly more approachable for non-nixers.

bollu:
Nice, this is very well documented! ^_^

PR #1436, opened 2025-07-07
chore: run small width independent test evaluation in CI

bollu:
This allows us to make sure that our width independent code runs correctly, which will help with the larger refactor in mind of the test pipeline


tobiasgrosser:
This is not needed anymore.

PR #1447, opened 2025-07-14
refactor: move term-model to stand-alone file

alexkeizer:
This PR moves the term model development from the core Framework.lean file to its own file. It seems not to be used anywhere, so I'm not sure if we want to keep it around, but this is a first step towards organizing the behemoth that is our Framework.lean file.


alexkeizer:
I'm merging this, as it's a simple refactor

PR #1452, opened 2025-07-16
refactor: make `Ctxt` a newtype structure instead of a def

alexkeizer:
This PR changes the `Ctxt Ty` type to be a structure with a single field of type `List Ty` instead of a `def` defined as the same. This way, the conversion from Ctxt to List becomes an explicit `toList / ofList` definition in the expression, which we can then match on in simp rewrites.
More importantly, it ensures we can add instances on Ctxt which might behave differently from the corresponding List instance, without inviting dependent-typed hell. For example, we might want to define what it means to append a list of types to a context, and we might want the order of the result to be slightly different from what appending the underlying lists would give us.

I'm not quite sure why, but this change triggered some bugs related to termination checking in some of the trickier mutually recursive functions (e.g., `matchVar`/`matchArg`). I added explicit termination_by hints to guide the checker, but this changed the error from a failure to check termination into:
```
Cannot derive matchArg._mutual.eq_def
  failed to generate splitter for match auxiliary declaration 'matchVar.match_3', unsolved subgoal:
```
I eventually figured that moving the match on the variable `w : Var Δ_out t` from the top-level matcharms down into a separate match just on `w` avoids the error. Why? Who knows? 

This error seems to have been the root-cause of the termination checker failure: after this match-change we can remove the explicit termination hints and the termination checker will find the right bounds automatically again. However, it seemed more robust to me to leave them in, in case we run into similar issues in the future.


alexkeizer:
This PR is likely to cause breakage in any other branches that work with contexts. I plan to merge this next week when I'm back again, so @luisacicolini @salinhkuhn, if you have any long-running branches it might be good to get this merged soon (of course, if that's not feasible, feel free to ping me post merge of this PR if it breaks your stuff).

In particular, this PR makes `Ctxt` a structure, meaning the type is no longer strictly (def-)equal to Lists. There is a coercion, but it doesn't apply in each situation, so if you encounter an error along the lines of "expected a `Ctxt`, found a `List`" the solution is to enclose the list term with the langle-rangle constructor brackets (`\<` ... `\>`) to call the constructor explicitly.

Another error that might happen is "Invalid dotted identifier notation: The expected type of `.foo` could not be determined", where `.foo` is using dot-notation to construct a type in your dialect, e.g., this could happen in `Com LLVMPlusRiscV [.llvm (.bitvec 32), .llvm (.bitvec 32)]`. The solution is to either not use dot-notation (i.e., by either opening the relevant namespace and using `llvm`, or fully qualifying the constructor), or by once again enclosing the context-list with brackets, as in `Com LLVMPlusRiscV ⟨[.llvm (.bitvec 32), .llvm (.bitvec 32)]⟩`.

tobiasgrosser:
> This PR is likely to cause breakage in any other branches that work with contexts. I plan to merge this next week when I'm back again, so @luisacicolini @salinhkuhn, if you have any long-running branches it might be good to get this merged soon (of course, if that's not feasible, feel free to ping me post merge of this PR if it breaks your stuff).
> 
> In particular, this PR makes `Ctxt` a structure, meaning the type is no longer strictly (def-)equal to Lists. There is a coercion, but it doesn't apply in each situation, so if you encounter an error along the lines of "expected a `Ctxt`, found a `List`" the solution is to enclose the list term with the langle-rangle constructor brackets (`\<` ... `\>`) to call the constructor explicitly.
> 
> Another error that might happen is "Invalid dotted identifier notation: The expected type of `.foo` could not be determined", where `.foo` is using dot-notation to construct a type in your dialect, e.g., this could happen in `Com LLVMPlusRiscV [.llvm (.bitvec 32), .llvm (.bitvec 32)]`. The solution is to either not use dot-notation (i.e., by either opening the relevant namespace and using `llvm`, or fully qualifying the constructor), or by once again enclosing the context-list with brackets, as in `Com LLVMPlusRiscV ⟨[.llvm (.bitvec 32), .llvm (.bitvec 32)]⟩`.

Nice. This is a great change. Let's get it merged when you are back.

PR #1456, opened 2025-07-17
refactor: replace DCE-specific API with existing Ctxt.Hom API

alexkeizer:
This simplifies a lot of the proof development in DCE

### DIFF
--- a/SSA/Projects/DCE/DCE.lean
+++ b/SSA/Projects/DCE/DCE.lean
@@ -4,10 +4,34 @@ Released under Apache 2.0 license as described in the file LICENSE.
 import SSA.Core.Framework
 import Mathlib.Tactic.Linarith
 
+open Ctxt (Var Valuation Hom)
+
 
 /-- Delete a variable from a list. -/
 def Ctxt.delete (Γ : Ctxt Ty) (v : Γ.Var α) : Ctxt Ty :=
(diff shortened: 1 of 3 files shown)


SSA/Projects/DCE/DCE.lean
```diff
@@ -4,294 +4,151 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+
 
 /-- Delete a variable from a list. -/
 def Ctxt.delete (Γ : Ctxt Ty) (v : Γ.Var α) : Ctxt Ty :=
   Γ.eraseIdx v.val
 
+@[simp] theorem Ctxt.delete_snoc_toSnoc {Γ : Ctxt Ty} {v : Γ.Var t} :
+    (Γ.snoc u).delete (v.toSnoc) = (Γ.delete v).snoc u := rfl
+
+def Hom.delete {Γ : Ctxt Ty} (delv : Γ.Var t) : Hom (Γ.delete delv) Γ :=
```
bollu, line 17:
```suggestion
def Hom.deleteDomain {Γ : Ctxt Ty} (delv : Γ.Var t) : Hom (Γ.delete delv) Γ :=
```

Nice! I am a huge fan of this API ^_^ 
Might be useful to say that we delete from the domain, as there is another operation where we can delete from the *codomain*.

PR #1491, opened 2025-08-01
chore: throw a CI failure when Alive statistics are not as expected

alexkeizer:
This PR changes the CI to fail whenever the Alive statistics aren't as they should be, instead of posting a comment to every PR. Since these statistics aren't expected to change, these comments just cause unhelpful noise, and are thus easy to overlook.


alexkeizer:
I've tested and confirmed that CI will indeed not pass when the numbers are not as expected: see https://github.com/opencompl/lean-mlir/actions/runs/16678590421/job/47211372191?pr=1491

tobiasgrosser:
Nice. Thank you, @alexkeizer.

PR #1503, opened 2025-08-04
feat: dc.select: transact on only values it inspects.

bollu:
Thanks for the explanation, @luisacicolini ! I've tagged you as a reviewer.


luisacicolini:
lgtm, thanks @bollu!

PR #1508, opened 2025-08-05
refactor: denote Expr into a Valuation map

alexkeizer:
This PR changes `Expr.denote` to return the new Valuation (which is the input valuation with the result of the current expression added to the end). The result of *only* the current expression is now called `Expr.denoteOp`.

This serves to deduplicate the effect that an Expr has on Valuations, which previously was expressed both in Com and in Lets.

This PR prepares for multiple return values.


alexkeizer:
This should hopefully now be ready to merge :crossed_fingers:

PR #1510, opened 2025-08-06
refactor: make Com.rec' recursor more flexible

alexkeizer:
This PR changes the Com.rec' custom recursor for simple structural recursion on `Com`s (i.e., without mutually recursion on the contained Exprs) more flexible.

Specifically, it ensures the effect and return type--which never actually change--are treated as parameters rather than index. This means we'll be allowed to do `induction com` even when the effect or return type are specific expressions (rather than free variables).

This also means there's no need for `recPure` anymore, which previously we had to use to be able to recurse on known-pure `Com`s; this now becomes just an instance of the more flexible recursion principle.

### DIFF
--- a/SSA/Core/Framework.lean
+++ b/SSA/Core/Framework.lean
@@ -380,72 +380,75 @@ abbrev Expr.outContext (_ : Expr d Γ eff t) : Ctxt d.Ty :=
 ### Com recursors
 -/
 
+section Rec
+variable {eff t} {motive : ∀ {Γ}, Com d Γ eff t → Sort u}
+          (ret : ∀ {Γ} , (v : Var Γ t) → motive (Com.ret v))
+          (var : ∀ {Γ} {u : d.Ty},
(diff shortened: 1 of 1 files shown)


SSA/Core/Framework.lean
```diff
@@ -380,72 +380,75 @@ abbrev Expr.outContext (_ : Expr d Γ eff t) : Ctxt d.Ty :=
...
-    (ret : ∀ {Γ t}, (v : Var Γ t) → motive (Com.ret v))
-    (var : ∀ {Γ} {t u : d.Ty},
-      (e : Expr d Γ .pure t) → (body : Com d (Γ.snoc t) .pure u) →
+end Rec
+
+/-- Alternative recursion principle for known-pure `Com`s.
+
+NOTE: this can now be achieved through `Com.rec'` direclty;
+`recPure` is deprecated and will be removed soon. -/
+-- TODO: eventually remove this
```
tobiasgrosser, line 443: Any reason you did not drop this immediately? It seems you moved most (all?) instances over already?

PR #1511, opened 2025-08-06
feat: Add support for multiple return values

alexkeizer:
This PR adds support for multiple return values in `Com`!

The core change is that `DialectSignature.outTy` now becomes `DialectSignature.returnTypes`, and is generalized to return a List of types. All the other changes are effectively the natural consequence of this.

- The proof strategy in `Match.lean` has changed to rely more heavily on Subtypes. This means that we can have simp lemmas that are effectively keyed on the goals, instead of having to rely on remembering which rewrite to apply (since they are disambiguated by side-condition).
- DCE had to be generalized to delete a (consecutive) range of variables
- We drive-by fix some printing instances so that (a) HVector's printing no longer includes stray extra brackets and commas and (b) so that the Com/Expr printing code does not include brackets/list-notation, but instead look like, e.g., `(i64, i64)`
- The TermModel code got broken, but it's not actually used anywhere, so :shrug: 
- Finally, we adjust the entire rest of the project to align with the possibility of multiple return values:
  - Dialect operation semantics now have to return an hvector of values (since there could be multiple); given that existing dialects only have a single return for each operation, we just wrap the semantics in `[...]\_h`
  - Similarly, we have to introduce some list literals whenever a `Com` or `Expr` type is referenced:
  - There is also a `LeanMLIR.SingleReturnCompat` namespace which contains backwards compatible aliasses, so for simple cases a straightforward fix is `open LeanMLIR.SingleReturnCompat` to recover the old behaviour.
  - Finally, we had to fix the proof automation to deal with the new shape of semantics:
    - In particular, I wrote `simpValuationApply`, which is a (d)simproc that normalizes "applications of *fully-concrete* valuations to *fully-concrete* variables", so for example, `(\empty ::\_v x ::\_v y) \<1, _\>` is rewritten to just `x`. 
    - This used to be done by composing simp lemmas and hoping for the best; this somehow broke but rather than figuring out why exactly moving to a simproc gives us much more control and thus robustness


SSA/Core/Framework/ToExpr.lean
```diff
@@ -155,12 +153,15 @@ partial def Regions.toExprAux {regSig : List (Ctxt d.Ty × d.Ty)}
 
 end
 
+
+
```
alexkeizer:
```suggestion
```

SSA/Core/Transforms/Rewrite/Match.lean
```diff
@@ -101,40 +101,239 @@ def matchVar {Γ_in Γ_out Δ_in Δ_out : Ctxt d.Ty} {t : d.Ty} [DecidableEq d.O
...
+--     → α
+--   | ⟨_, h⟩ => False.elim <| by
+--       simp [matchVar] at h
+--       sorry
+-- def eqvVarRight  :
+--     MatchVarResult lets v (.var matchLets matchExpr) w.appendInr ma
+--     ≃ MatchVarResult lets v matchLets w ma where
+--   toFun := fun ⟨x, h⟩ => ⟨x, by simp_all⟩
+--   invFun := fun ⟨x, h⟩ => ⟨x, by simp_all⟩
+
```
alexkeizer:
```suggestion
```

SSA/Core/Framework.lean
```diff
@@ -686,30 +772,31 @@ An alternative representation of a program as a `Lets` with a return `Var`
...
   simp [denoteLets]
 
-theorem FlatCom.denote_eq [DialectDenote d] (flatCom : FlatCom d Γ eff Γ_out t) :
-    flatCom.denote = fun Γv => flatCom.lets.denote Γv >>= fun Γ'v => return (Γ'v flatCom.ret) := by
-  funext Γv
-  simp [denote]
+-- theorem FlatCom.denote_eq [DialectDenote d] (flatCom : FlatCom d Γ eff Γ_out ts) :
+--     flatCom.denote = fun Γv => flatCom.lets.denote Γv >>= fun Γ'v => return (Γ'v flatCom.rets) := by
+--   funext Γv
+--   simp [denote]
```
```suggestion
```

SSA/Core/Framework.lean
```diff
@@ -903,41 +993,18 @@ assignment of that variable in the input valuation -/
...
---   Valuation.comap_outContextHom_denoteLets, Com.denoteLets_returnVar_pure,
---   denote_mk]
---   unfold Valuation.reassignVar
---   congr; funext V_mid; congr
---   funext t' v'
---   simp only [dite_eq_right_iff, forall_exists_index]
---   rintro rfl rfl
---   simpa using h _ (sorry : V_mid ∈ Functor.supp (lets.denote V_in))
-
 end DenoteInsert
```
alexkeizer, line 984: This got implemented in the Zipper file

SSA/Core/Framework.lean
```diff
@@ -292,66 +299,72 @@ def Com.rec' {Γ} (com : Com d Γ eff t) : motive com :=
...
-    HVector (fun t : Ctxt d.Ty × d.Ty => Com d t.1 .impure t.2) (DialectSignature.regSig e.op) :=
+def Expr.regArgs {Γ ts} (e : Expr d Γ eff ts) :
+    Regions d (DialectSignature.regSig e.op) :=
   Expr.casesOn e (fun _ _ _ _ regArgs => regArgs)
 
+-- TODO: I don't think `returnVar` is used, remove it
 /-- `e.returnVar` is the variable in `e.outContext` which is bound by `e`. -/
-def Expr.returnVar (e : Expr d Γ eff ty) : e.outContext.Var ty :=
+def Expr.returnVar (e : Expr d Γ eff [ty]) : e.outContext.Var ty :=
   Var.last _ _
```
alexkeizer: I should follow this TODO

SSA/Core/Transforms/Rewrite/Match.lean
```diff
@@ -101,40 +101,218 @@ def matchVar {Γ_in Γ_out Δ_in Δ_out : Ctxt d.Ty} {t : d.Ty} [DecidableEq d.O
...
+### Subtypes
+-/
+attribute [refl] List.Subset.refl -- TODO: upstream somewhere
+
+variable (lets v matchLets w) (mapIn : Mapping _ _) in
+def MatchVarResult := { mapOut : Mapping _ _ //
+  ∃ (mapIn' mapOut' : Mapping _ _),
+    mapIn.entries ⊆ mapIn'.entries
+    ∧ mapOut'.entries ⊆ mapOut.entries
+    ∧  matchVar lets v matchLets w mapIn' = some ((), mapOut') }
```
alexkeizer:
```suggestion
    ∧ matchVar lets v matchLets w mapIn' = some ((), mapOut') }
```

whitespace

alexkeizer:
@luisacicolini This is currently building :tada: Could you review / skim over the RiscV bits to double check I didn't drop anything important, and to get a feel for the shape of changes necessary once this is merged?

Also, if you merge anything else into main, could you let me know (or even merge those changes into this branch (post merging into main) if you feel particularly inspired :crossed_fingers:)?

luisacicolini:
thanks a lot for taking care of this @alexkeizer! It all looks good to me. The PRs that are likely to be merged over the next days (#1545, #1585) should not be impacted by what you are doing, AFAICT, and even if they were the fixes look pretty manageable :) 

I won't merge in this branch as the code is not stable yet, once that happens I will consider it, unless this PR is merged already.

SSA/Core/Tactic.lean
```diff
@@ -132,6 +141,71 @@ simproc [simp_denote] elimValuation (∀ (_ : Ctxt.Valuation _), _) := fun e =>
...
+          \n\t{V}"
+      Meta.check e
+      return .continue
+
+end SimpValuationApply
+
+/-!
+### HVector.cons injectivity
+
+WORKAROUND: Simplifying the semantics can yields equalities of the form:
```
alexkeizer:
```suggestion
WORKAROUND: Simplifying the semantics can yield equalities of the form:
```

PR #1514, opened 2025-08-06
fix: compute coqQFBV numbers and make solved numbers absolute

abdoo8080:


bollu:
Thanks abdal; I've cherry-picked this commit into https://github.com/opencompl/lean-mlir/pull/1516, which I will merge into `main`. That will ensure that we have (a) a stable commit https://github.com/opencompl/lean-mlir/pull/1516/commits/9ed595567e409513d6063e54f88223ac0c80d8d9 to update the artifact with, and (b) that this commit lives in the history of the `main` branch.

bollu:
@abdoo8080 If you want to make further changes, please push to the branch https://github.com/opencompl/lean-mlir/pull/1516 directly, which you can since you have write access to `opencompl/lean-mlir` ^_^

PR #1527, opened 2025-08-14
feat: add example of CIRCT lowering with comb

luisacicolini:
This PR adds an example lowering of a circuit involving the handshake and DC functors with comb. It also fixes two minor parsing bugs that I noticed during the process in the DC and Handshake functors.


luisacicolini:
merging once CI passes as the changes are limited to DC.

luisacicolini:
Removed one last file containing the `SyncMap` which I will upstream separately, merging the examples in the meantime.

PR #1532, opened 2025-08-21
chore:Multi-width Arithmetic Relations (slt, ult, sle, ule, neq)

bollu:
This completes our support for the arithmetic fragment. I realized that `neq` is more nontrivial than I originally thought due to a change in encoding.


bollu:
Superceded by #1633

PR #1533, opened 2025-08-21
feat: Make extract goals produce just the theorem statement, with imports commented out

bollu:
This allows for easier external scripting, since the external script can grab the contents of the test file and slap it into a lean file with the right imports. 
We keep the suggested list of imports in comments for easy editing.


alexkeizer (approved):
I don't *love* that the .lean files under tests/ no longer compile as-is, but after discussion with @bollu I see why doing so makes the scripting nicer, so I'll concede

PR #1537, opened 2025-08-22
chore: refactor the scripts and folder currently in `LLVMRiscV/Evaluation`

luisacicolini:
This PR refactors the scripts to generate the tests and their folder. In particular: 
* `mlir_splitter`, `mlir_bb0_extracter`, `extractbb0` are now merged in a single script, with one function per lowering/pass
* the output of each lowering/pass is stored in a separate file 

I `ruff format generate.py && ruff check generate.py`and everything is okay. The `README` describes the whole toolchain. 

I rebased this PR on top of #1539, such that the changes to `main` are the same.

PR #1538, opened 2025-08-22
feat: add prototype evaluation for LLVMRiscV

luisacicolini:
A few notes: 

* I won't merge `simpproc.lean` from this PR, because the version that currently is on main is identical and only has more comments - which we want. I will thus maintain the version on main.

* I will not merge `ParseAndTransform.lean` because again the version on main is the correct one AFAICT. In fact, on this **branch** we have 
```lean
          let lowered := selectionPipeFuelWithCSE c true /- calls the instruction selector defined in `
                                                  InstructionLowering`, true indicates pseudo variable lowering-/
          IO.println s!"{Com.toPrint (lowered) }"
          return 0
```
while on **main** we have
```lean          
          let lowered := selectionPipeFuelSafe c /- calls the instruction selector defined in `
                                                  InstructionLowering`-/
          IO.println s!"{toString lowered}"
```
and an extra
```
#eval! passriscv64 "test/LLVMDialect/InstCombine/bb0.mlir"
```

* the folder `Evaluation/Comparison` except for the file `mlirtest.mlir`, while `prototypeEval.mlir` has been renamed to `out_100.mlir` and now lives in #1537 with the rest of the refactored testing infra.

* the folder `Evaluation/LLVM-RISCV-testSuite` now lives in #1537 and has been refactored.

* the folder `Evaluation/ProofAutomation` now lives in #1537 and has been refactored.

* folder `tests/*` have not been ported anywhere, as to the best of my understanding they are not strictly part of the evaulation. The only exceptions are files `out.mlir` and `out2.mlir` which have been renamed to `out_1k.mlir` and `out_100k.mlir`, respectively, and now live in #1537, in `Evaluation/benchmarks/MLIR_multi/`


luisacicolini:
closing as verything has been refactored and upstreamed

PR #1576, opened 2025-08-25
chore: write a simproc to simplify `HVector.getN` applied to a fully concrete HVector

alexkeizer:
This ought to fix the recent opt breakage


alexkeizer:
superseded by / folded into #1580

PR #1584, opened 2025-08-25
chore: add `CIRCT` build target

alexkeizer:
This PR takes all the CIRCT files currently built in CI, adds them to `Projects.CIRCT.CIRCT` as imports and then adds a new build target with this file as root. This makes it easier to build/test the CIRCT files locally without having to build each file individually.

Since there was some overlap between names defined in the DC and Handshake example files, we introduce an Example namespace in both files


bollu:
brilliant, thank you for performing this refactoring!

PR #1585, opened 2025-08-25
Add GlobalISel compilation flow to eval scripts.

salinhkuhn:
Adding a third compilation flow to generate.py and mc-anaylsis.py to compare against not only SelectionDAG but also GlobalISel.

I only changed generate.py and the Mac script by adding self-contained additional functions without touching @luisacicolini code or changing any of it. I adapt her style and added a third folder for the GlobalISel flow.


tobiasgrosser:
Sarah's work as meanwhile all been merged.

PR #1587, opened 2025-08-26
chore: peel the functionality in bv_evaluation to limit memory/time.

bollu:
We refactor this into a separate 'liblimits.py, which will be shared across our evaluation scripts.

### DIFF
--- a/bv-evaluation/runwithlimits.py
+++ b/bv-evaluation/runwithlimits.py
@@ -91,6 +91,3 @@ def run_with_limits(cmd: List[str], timeout: int, memout_mb: int) -> (str, str, 
             return "TIMEOUT", stdout, stderr
     except Exception as e:
         return "ERROR", "", str(e)
-
-
-
--- a/bv-evaluation/runwithlimits.py
+++ b/bv-evaluation/runwithlimits.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python3
+import os
+import subprocess
+import shutil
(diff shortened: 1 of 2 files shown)


bollu:
CC @tobiasgrosser. It would be good to get this in.

alexkeizer:
I don't really grok the python bits well enough to understand exactly what it's doing at first glance, but AFAICT this is just moving existing code verbatim to a new file? I'm always a fan of modularizing our code and organizing it with a bit more structure, so this seems like a clear improvement to me :+1:

bv-evaluation/runwithlimits.py
```diff
@@ -0,0 +1,96 @@
...
+            flag["done"] = True
+            kill_process_tree(proc.pid)
+            stdout, stderr = proc.communicate()
+            monitor_thread.join()
+            return "TIMEOUT", stdout, stderr
+    except Exception as e:
+        return "ERROR", "", str(e)
+
+
+
```
tobiasgrosser: Do we need these empty lines at the end?

tobiasgrosser:
This overall looks good to me. However, I would appreciate an LGTM from @luisacicolini, who was very involved with the evaluation and maybe a quick one from @ineol, who also looked into some of this code. Let's look at the code and do a brief review as the original code was written a bit in a rush.

luisacicolini (approved):

ineol (approved):

PR #1599, opened 2025-08-26
feat: run evaluation steps concurrently

tobiasgrosser:
Split off the building of Lean into a separate step that caches its outputs, then extract three evaluation steps: `Check extracted goals`, `Evaluate LLVM`, `Evaluate Alive & Hacker's Delight`. These evaluation steps can run concurrently, all reusing the previous Lean-MLIR build. We build Mathlib from scratch (not using the pre-built Mathlib) as this allows us to build only the files in Mathlib that we actually use, which reduces the size of the cache we store on GitHub.

We use GitHub's `action/cache` to pass the build data from the `Build Lean-MLIR` step to the individual evaluation steps, which is fast and seems less complex than building an artifact. When the cache is hit, the `Build Lean-MLIR` step takes 7 seconds. Our full cache with Lean-MLIR, Mathlib, and Lean requires 1200 MB and can be downloaded in 16 seconds.

This reduces the runtime of our evaluation from 45 minutes to 20 minutes.


.github/workflows/evaluation.yml
```diff
@@ -84,6 +148,40 @@ jobs:
...
+          lake -R build
+
+      - name: Install Python Modules
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y python3-matplotlib python3-pandas python3-num2words python3-psutil ripgrep
+
       - name: Run Alive Symbolic
         run: |
           (cd bv-evaluation; python3 ./compare-leansat-vs-bitwuzla-alive-sym.py -j48)
```
bollu, line 187: I think this rule is also unnecessary, we can comment it out, since we never use this anymore. We can delete this in the next PR.

.github/workflows/evaluation.yml
```diff
@@ -84,6 +148,40 @@ jobs:
...
+      - name: Load Lean and Lean-MLIR from Cache
+        id: cache-lake
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            ~/.elan
+            .lake/packages
+            .lake/build
+            TacBench/.lake
+          key: ${{ runner.os }}-lake-packages-${{ hashFiles('lake-manifest.json') }}
```
bollu, line 170:
```suggestion
          key: ${{ runner.os }}-lake-packages-${{ hashFiles('**/lake-manifest.json') }}
```

While this is extra-paranoid, I think it's cleaner to hash by all manifests.

alexkeizer: Are the manifests of dependencies even respected in a `lake build`? I thought not.

Regardless, it's not actually a soundness concern, since lake will do its own verification of the cache, so this is not defensive paranoia; rather it's premature optimization and I'd suggest we stick to the old version. (But, I don't feel super strongly about it, so take it or leave it)

.github/workflows/evaluation.yml
alexkeizer: If that is true, can we also delete the relevant scripts / Lean files?

bollu: @alexkeizer I'd like @luisacicolini 's OK on that, but yes, we can as per my undestanding! We only need `compare.py`, `collect.py`,`plot.py`, `compare-automata-automata-circuit-jsonl.py`, and`plot-automata-automata-circuit.py`.

tobiasgrosser: I suggest we move cleaning up the scripts / CI to a separate PR. This PR does not change the number of scripts that are run, but I appreciate a follow-up PR that drops scripts that are not needed.

.github/workflows/evaluation.yml
tobiasgrosser: Happy to discuss this in person, to go either way. It seems none is a critical concern at the moment, so I merge this as is and can then address this in a follow-up conversation tomorrow.

PR #1601, opened 2025-08-27
chore: rename 'lemma' to 'theorem' in SSA/Core

tobiasgrosser:
This frees SSA/Core from another dependency on mathlib. In this particular case, this is purely cosmetic.

Reducing our dependency on mathlib has a chance to speedup CI, as less of mathlib needs to be build and cached.

This is nudging us towards minimizing the dependencies on mathlib.

### DIFF
--- a/SSA/Core/Transforms/Rewrite/Match.lean
+++ b/SSA/Core/Transforms/Rewrite/Match.lean
@@ -226,7 +226,7 @@ def MatchArgResult.mk {mapOut : Mapping _ _}
   ⟨mapOut, mapIn, mapOut, ⟨by rfl, by rfl, h⟩⟩
 
 /-!
-### Subtype-based theorems
+### Subtype-based lemmas
 -/
 
 @[simp] theorem lookup_matchVar_nil (m : MatchVarResult lets v .nil w ma) :
@@ -245,14 +245,14 @@ theorem ofSubsetEntries {varMap₁ : MatchVarResult lets v matchLets w ma} {varM
   refine ⟨⟨varMap₂, mapIn', map', h₁, ?_, h₃⟩, rfl⟩
   trans; exact h₂; exact h_sub
 
-/-! ### nil theorems -/
+/-! ### nil lemmas -/
 
 variable [TyDenote d.Ty] [∀ (t : d.Ty), Inhabited ⟦t⟧] in
 @[simp] theorem mapValuation_nil (mapOut : MatchVarResult lets v .nil w mapIn) (V) :
     mapOut.val.mapValuation V w = V v := by
   simp [Mapping.mapValuation]
 
-/-! ### var_appendInl theorems -/
+/-! ### var_appendInl lemmas -/
 section Left
 variable {w : Δ_out.Var t}
 
@@ -271,7 +271,7 @@ variable {mapIn} (mapOut : MatchVarResult lets v (.var matchLets matchExpr) w.ap
 @[simp, defeq] theorem entries_eqvVar : (eqvVarLeft mapOut).val.entries = mapOut.val.entries := rfl
 
 end Left
-/-! ### var_appendInr theorems -/
+/-! ### var_appendInr lemmas -/
 variable {w : Var ⟨te⟩ _} {mapIn}
 
 theorem getPureExpr_eq_some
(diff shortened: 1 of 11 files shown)


bollu:
LGTM, seems purely cosmetic to me, and clearly a Good Thing ™️  Thanks for this! 💖

luisacicolini (approved):

alexkeizer:
I personally like the aesthetics of `lemma`, but I'll concede it pragmatically is not worth the mathlib dependency.

> less of mathlib needs to be build

We don't really build Mathlib to begin with (unless we're building an executable like `opt`), and `lake exe cache get` will download everything irrespective of what is actually used, so I expect that minimizing Mathlib dependencies won't really affect our build times *until* we manage to eliminate *all* of Mathlib. That's not to say we shouldn't start nudging ourselves in that direction already, but we should at some point have a discussion about whether eliminating all of Mathlib is feasible/desirable.

Currently, not downloading all of mathlib via `lake exe cache get` but instead just calling `lake build` will lead to a roughly 50% smaller .lake directory and save a couple 100 MB of disk/cache space. This has a some impact on compilation time but not sure it is impressive.

I feel slowly nudging towards minimizing our mathlib dependencies where they are needed is a good code-health practice. I agree that before we invest heavily, we probably want to have a discussion first.

SSA/Core/Util/Poison.lean
```diff
@@ -66,7 +66,7 @@ instance [ToString α] : ToString (PoisonOr α) where
   | .poison  => "poison"
   | .value a => "(value " ++ addParenHeuristic (toString a) ++ ")"
 
-/-! ### Monad instance and lemmas -/
+/-! ### Monad instance and theorems -/
```
alexkeizer:
```suggestion
/-! ### Monad instance and lemmas -/
```

We could at least keep lemma around in documentation

SSA/Core/Transforms/Rewrite/Rewrite.lean
```diff
@@ -237,10 +237,10 @@ def multiRewritePeephole (fuel : ℕ)
     (prs : List (Σ Γ, Σ ty, PeepholeRewrite d Γ ty)) (target : Com d Γ₂ eff t₂) : (Com d Γ₂ eff t₂) :=
   multiRewritePeepholeAt fuel prs 0 target
 
-/-- helper lemma for the proof of `denote_rewritePeephole_go_multi`. It proofs that folding
+/-- helper theorem for the proof of `denote_rewritePeephole_go_multi`. It proofs that folding
```
alexkeizer: Here, too, I prefer calling it a lemma in documentation at least, to signify it's not a particularly interesting result, just API/plumbing

alexkeizer:
```suggestion
/-- helper theorem for the proof of `denote_rewritePeephole_go_multi`. It proves that folding
```
While we're at it, might as well fix that typo

SSA/Core/Transforms/Rewrite/Match.lean
```diff
@@ -263,15 +263,15 @@ def eqvVarLeft  :
...
   (eqvVarLeft (matchExpr:=matchExpr) |>.symm  varMap).val.entries = varMap.val.entries := rfl
 
 variable {mapIn} (mapOut : MatchVarResult lets v (.var matchLets matchExpr) w.appendInl mapIn)
 
-@[simp, defeq] lemma entries_eqvVar : (eqvVarLeft mapOut).val.entries = mapOut.val.entries := rfl
+@[simp, defeq] theorem entries_eqvVar : (eqvVarLeft mapOut).val.entries = mapOut.val.entries := rfl
 
 end Left
-/-! ### var_appendInr lemmas -/
+/-! ### var_appendInr theorems -/
```
alexkeizer: This one, too

SSA/Core/Transforms/Rewrite/Rewrite.lean
tobiasgrosser: I went over the comments and removed any changes in comments. Thank you for catching this.

SSA/Core/Transforms/Rewrite/Match.lean
tobiasgrosser: Fixed.

tobiasgrosser:
Thank you, all. This is now cleaned and -- as no further concerns have been raised -- ready to go. I note that we want to think about a general mathlib strategy for the future.

PR #1604, opened 2025-08-27
feat: add stride/offset options to extract_goals to allow for batch processing.

bollu:
We also take the opportunity to clean up the verbosity levels in the script, thereby allowing us to ensure that we do not flood CI.


bollu:
Once https://github.com/opencompl/lean-mlir/pull/1599 is in, we can use stride/offset info to run goal extraction in parallel.

PR #1605, opened 2025-08-27
chore: improve caching in Dockerfile builds

alexkeizer:
This PR tweaks the Dockerfile to use a cached .lake contents from previous builds, speeding up (**non-CI**) builds of the Docker image.

We also clean up the Dockerfile to no longer copy the lakefile and manifest first. This was originally done to maximize the use of the image layer cache by downloading dependencies first; but now that we have proper .lake directory caching that is no longer necessary. This cleans up the Dockerfile, and should make the introduction of further subproject less complex.

Do note that, since cache mounts are broken in GH-provided action runners, this change likely makes CI builds a bit slower.
However, those were super slow already, and as per the rant below, the fix for that is to move to non-ephemeral build environments with proper local (garbage-collected!) cache mechanisms.

-----

### Regarding CI

Note that the cache-mounts are persisted by the Docker deamon. Since Github CI uses ephemeral environments, this cache is lost between runs; so CI builds are not actually sped up. This is true despite us specifying the GHA (GitHub Action) cache backend, because this backend explicitly only persists cached image layers and *not* cache mounts.

I've tried the "official" workaround ([reproducible-containers/buildkit-cache-dance](https://github.com/reproducible-containers/buildkit-cache-dance), but this didn't work:
- Firstly, when using this action, CI was no longer hitting the cache for image layers, causing those to be unneccesarily rebuilt, slowing down the build massively
- Moreover, it actually caused the builds to fail with some error message about `.lake/bin/cache` not being able to executed;
  - The workaround involved bind-mounting files and copying them; I conjecture permission bits somehow got lost during this copy, but I failed to properly debug this

Then I tried to use bind-mounts directly, but those don't work either as even when passing the `rw` option, the written files are *not* saved to the host file-system, meaning that subsequent runs cannot benefit from newly cached files.

Another workaround involves uploading the entire buildkit cache directory to GH Cache, but this then runs into issues arround the cache growing unboundedly, so there is some ad-hoc workaround for the workaround, which is optimistically labelled temporary, but links to an issue from 2020 which has seen no activity towards resolution.

It was at this point that I decided to preserve my sanity and declare the GitHub Action caching story broken; we should just run our own Docker builder instead (i.e, implement #1606).

### DIFF
--- a/SSA/Core/Framework.lean
+++ b/SSA/Core/Framework.lean
@@ -3,7 +3,6 @@ Released under Apache 2.0 license as described in the file LICENSE.
 -/
 -- Investigations on asymptotic behavior of representing programs with large explicit contexts
 
-
 import SSA.Core.ErasedContext
 import SSA.Core.HVector
 import SSA.Core.EffectKind
--- a/SSA/Core/Framework.lean
+++ b/SSA/Core/Framework.lean
@@ -3,6 +3,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 -/
 -- Investigations on asymptotic behavior of representing programs with large explicit contexts
 
+
 import SSA.Core.ErasedContext
 import SSA.Core.HVector
 import SSA.Core.EffectKind
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+Dockerfile
+*.pdf
+.lake/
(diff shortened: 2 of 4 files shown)


SSA/Core/Framework.lean
```diff
@@ -3,6 +3,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 -/
 -- Investigations on asymptotic behavior of representing programs with large explicit contexts
 
+
```
```suggestion
```

PR #1608, opened 2025-08-27
chore: reduce loglevel for extract-goal CI

tobiasgrosser:
The current log-level was very verbose (linear in # of test cases). Thanks to Sid's 6d0b3d168, the verbosity can now easily be reduced.


bollu:
LGTM

PR #1613, opened 2025-08-27
refactor: use namespace.so runners to build Docker image

alexkeizer:


alexkeizer:
After discussing with @tobiasgrosser, we decided to merge this and experiment with using namespace's runner to build the core Docker image only, but keep it out of the critical path so that we can easily revert this decision if desired in future.

PR #1621, opened 2025-08-28
feat: peel automata decision procedures into separate lake project for customers

bollu:


Blase/lean-toolchain
```diff
@@ -0,0 +1 @@
+leanprover/lean4:nightly-2025-08-11
```
alexkeizer: @bollu can you make this file a symlink to the root lean-toolchain file, so that they are kept in sync?

Would anything break if we do so?

bollu: I think that would work just fine!

bollu: Actually, I could imagine that a symlink could lead to really weird artifacts with e.g. bundling into a tar with `lake pack`. I am happy to write us a script that polls the lean toolchain from `Mathlib` and bumps this up everywhere.

SSA/Projects/InstCombine/TacticAuto.lean
```diff
@@ -6,9 +6,9 @@ import Std.Tactic.BVDecide
 import Leanwuzla
 
 import SSA.Projects.InstCombine.Tactic.SimpLLVM
-import SSA.Experimental.Bits.SingleWidth.Tactic
-import SSA.Experimental.Bits.Fast.MBA
-import SSA.Experimental.Bits.AutoStructs.ForLean
+import Blase.SingleWidth.Tactic
+import Blase.Fast.MBA
+-- import Blase.AutoStructs.ForLean
```
alexkeizer:
```suggestion
```
Let's drop this import completely if it's not needed, rather than comment it out

bollu: done.

bollu:
and CI is green! 🎉

alexkeizer:
LGTM! Thanks for cleaning up the diff

PR #1626, opened 2025-08-28
chore: drop AliveHandwritten files

### LINKED ISSUE #1595: Can we drop AliveHandwritten(Large)Examples?
`AliveHandwrittenLargeExamples.lean` is starting to become the bane of my existence: these proofs are very brittle, by virtue of being handwritten and thus needing handwritten adjustment whenever framework stuff changes. What is the benefit of keeping these handwritten examples around? We have a pretty solid automated evaluation now, right, so could we just get rid of these handwritten example files?

alexkeizer:
These files had a tendency to break often, and paying the cost of maintaining them doesn't make much sense now that they have been superseded by our automated evaluation.

Fixes #1595


bollu:
LGTM

PR #1631, opened 2025-08-29
chore: drop custom bitvector notation

alexkeizer:
This PR removes the custom notation we had for signed & unsigned comparisons of bitvectors.

We still need to regenerate the extracted goals

### DIFF
--- a/SSA/Projects/InstCombine/LLVM/Semantics.lean
+++ b/SSA/Projects/InstCombine/LLVM/Semantics.lean
@@ -492,14 +492,14 @@ def icmp' {w : Nat} (c : IntPred) (x y : BitVec w) : Bool :=
   match c with
     | .eq => (x == y)
     | .ne => (x != y)
-    | .sgt => (x >ₛ y)
-    | .sge => (x ≥ₛ y)
-    | .slt => (x <ₛ y)
-    | .sle => (x ≤ₛ y)
-    | .ugt => (x >ᵤ y)
-    | .uge => (x ≥ᵤ y)
-    | .ult => (x <ᵤ y)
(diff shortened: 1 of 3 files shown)


tobiasgrosser:
This looks like a great idea. But why do we have so much snakemake here?

alexkeizer:
> This looks like a great idea. But why do we have so much snakemake here?

Because I picked the wrong place to start my branch, it's now rebased

tobiasgrosser:
This breaks CI, but is fine for me. As I am currently on leave, maybe @bollu can review this. This patch seems to go in a direction he is also interested in.

bollu:
Yes, happy to help here @alexkeizer , as this would help me create cleaner goal states as well

SSA/Projects/InstCombine/LLVM/Semantics.lean
```diff
@@ -492,14 +492,14 @@ def icmp' {w : Nat} (c : IntPred) (x y : BitVec w) : Bool :=
...
-    | .ult => (x <ᵤ y)
-    | .ule => (x ≤ᵤ y)
+    | .sgt => (y.slt x)
+    | .sge => (y.sle x)
+    | .slt => (x.slt y)
+    | .sle => (x.sle y)
+    | .ugt => (y.ult x)
+    | .uge => (y.ule x)
+    | .ult => (x.ult y)
+    | .ule => (x.ule y)
```
bollu, line 502: LGTM!

alexkeizer:
@bollu With our fix this is now building properly, could you run the extraction scripts and update the goals files to make the extract goals CI happy?

PR #1632, opened 2025-08-29
feat: add generic MLIR syntax support for overflow flag attributes in LLVM

salinhkuhn:
This PR extends the parsing logic of the LLVM dialect to support parsing overflow flags in their generic syntax form, as produced by mlir-opt when using the -generic flag.
The previous parsing logic for the custom syntax format is preserved, and a new match case is added for the generic syntax. The parser first reads the expected integer attribute, then derives the corresponding overflow flag based on its value. The mapping between integer values and overflow flags was extracted from compiled LLVM IR files.
With this change, lake exe opt can parse the generic flagged version of `sub`, `mul`,`add`and `shl`.


SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -40,20 +40,27 @@ In future, this might start throwing errors, if the type is not a bitvec.
...
+             return (⟨true, true⟩)
+        | _ =>  /- This case covers generic MLIR syntax for operations with attributes as
+                MLIR encodes them using integer attributes. -/
+              let ⟨n, _ ⟩ ← op.getIntAttr "overflowFlags"
+              match n with
+              | 0 => return (⟨false, false⟩)
+              | 1 => return (⟨true, false⟩)
+              | 2 => return (⟨false, true⟩)
+              | 3 => return (⟨true, true⟩)
+              | s => throw <| .generic s!"The overflow flag with prediacte {s} not allowed. \
```
luisacicolini:
```suggestion
              | s => throw <| .generic s!"The overflow flag with predicate {s} is not allowed. \
```
typo

luisacicolini (approved):
Just found a small typo, except for that LGTM!

SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -40,20 +40,27 @@ In future, this might start throwing errors, if the type is not a bitvec.
...
+        | .list [.opaque_ "llvm.overflow" "nsw", .opaque_ "llvm.overflow" "nuw"] =>
+             return (⟨true, true⟩)
+        | _ =>  /- This case covers generic MLIR syntax for operations with attributes as
+                MLIR encodes them using integer attributes. -/
+              let ⟨n, _ ⟩ ← op.getIntAttr "overflowFlags"
+              match n with
+              | 0 => return (⟨false, false⟩)
+              | 1 => return (⟨true, false⟩)
+              | 2 => return (⟨false, true⟩)
+              | 3 => return (⟨true, true⟩)
```
alexkeizer: NIT: if you write it like follows it's self-documenting what each number maps to, where currently it's a bit obfuscated

```suggestion
              | 0 => return { nsw := false, nuw := false}
              | 1 => return { nsw := true, nuw := false}
              | 2 => return { nsw := false, nuw := true}
              | 3 => return { nsw := true, nuw := true}
```

SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -40,20 +40,27 @@ In future, this might start throwing errors, if the type is not a bitvec.
...
+        | .opaque_ "llvm.overflow" "nsw" => return (⟨true, false⟩)
+        | .opaque_ "llvm.overflow" "nuw" => return (⟨false, true⟩)
+        | .opaque_ "llvm.overflow" "none" => return (⟨false, false⟩)
+        | .list [.opaque_ "llvm.overflow" "nuw", .opaque_ "llvm.overflow" "nsw"]
+        | .list [.opaque_ "llvm.overflow" "nsw", .opaque_ "llvm.overflow" "nuw"] =>
+             return (⟨true, true⟩)
+        | _ =>  /- This case covers generic MLIR syntax for operations with attributes as
+                MLIR encodes them using integer attributes. -/
+              let ⟨n, _ ⟩ ← op.getIntAttr "overflowFlags"
+              match n with
```
alexkeizer: Why did the spacing change here?

luisacicolini: fixed

SSA/Projects/InstCombine/Base.lean
```diff
@@ -136,7 +136,7 @@ def toStringWithFlags (op : MOp.BinaryOp) : String :=
     | .sdiv ⟨true⟩          => "sdiv exact"
     | .udiv ⟨false⟩         => "udiv"
     | .udiv ⟨true⟩          => "udiv exact"
-  s!"llvm.{op}"
+  s!"{op}"
```
alexkeizer: Can you drop this change? The double-printing should've been fixed in main already

SSA/Projects/InstCombine/LLVM/EDSL.lean
```diff
@@ -50,10 +50,16 @@ def parseOverflowFlags (op : AST.Op φ) : ReaderM φ LLVM.NoWrapFlags :=
     | .list [.opaque_ "llvm.overflow" "nuw", .opaque_ "llvm.overflow" "nsw"]
     | .list [.opaque_ "llvm.overflow" "nsw", .opaque_ "llvm.overflow" "nuw"] =>
         return ⟨true, true⟩
-    | .opaque_ "llvm.overflow" s => throw <| .generic s!"The overflow flag {s} not allowed. \
-        We currently support nsw (no signed wrap) and nuw (no unsigned wrap)"
-    | _ => throw <| .generic s!"Unrecognised overflow flag found: {MLIR.AST.docAttrVal y}. \
-        We currently support nsw (no signed wrap) and nuw (no unsigned wrap)"
+    | _ =>  /- This case covers generic MLIR syntax for operations with attributes as
+            MLIR encodes them usin integer attributes. -/
```
alexkeizer: I don't really understand the comment here. @luisacicolini I guess you didn't write it?
Maybe the following is clearer.

```suggestion
    | _ => 
      -- Parse oveflowFlags passed as integer attribute
```

SSA/Projects/InstCombine/Tests/Parse.lean
```diff
@@ -0,0 +1,14 @@
...
+^bb0(%0 : i64, %1 : i64):
+  %2 = "llvm."add true true""(%0, %1)<{overflowFlags = #llvm.overflow<nsw,nuw>}> : (i64, i64) -> (i64)
+  "llvm.return"(%2) : (i64) -> ()
+ }
+-/
+#guard_msgs in #eval String.toFormat <| Com.toPrint [llvm| {
+  ^bb0(%0: i64, %1 : i64):
+    %2 = "llvm.add"(%0, %1) {overflowFlags = 3 : i64} : (i64, i64) -> i64
+    "llvm.return"(%2) : (i64) -> ()
+}]
```
alexkeizer: For completeness sake, could you copy this test for each value of the overflowFlag (0, 1, 2 and 3)?

By the way, the existing "Print.lean" file really tests both printing and parsing; you could move these test-cases into that file in the subsection for the `add` instruction.

NIT: I went a little over-board and originally added test-cases for *every* instruction that uses overflowFlags. ChatGPT / co-pilot was actually quite useful in generating those, if you feel like expanding those test-cases to also test the numeric flags for each instruction that takes them. 
If not, though, I think having just the 4 tests for `add` is enough already

alexkeizer (commented):
I left some comments, feel free to merge after fixing those!

SSA/Projects/InstCombine/Tests/Parse.lean
```diff
@@ -0,0 +1,14 @@
+import SSA.Projects.InstCombine.LLVM.EDSL
+
+/--
+info: builtin.module { ⏎
+^bb0(%0 : i64, %1 : i64):
+  %2 = "llvm."add true true""(%0, %1)<{overflowFlags = #llvm.overflow<nsw,nuw>}> : (i64, i64) -> (i64)
```
tobiasgrosser: This `add true true` looks wrong. We can fix this in a subsequent PR if the bug is orthogonal.

alexkeizer: I believe this bug has been fixed already, @luisacicolini you might want to rebase on main to pull in that fix so that the test doesn't break post-merge.

If you still get this output post-rebase, feel free to ping me!

SSA/Projects/InstCombine/Base.lean
luisacicolini: merged main

luisacicolini: and fixed

SSA/Projects/InstCombine/Tests/Parse.lean
luisacicolini: added all the tests for add, I want to get this PR in but will open an issue to add more tests

SSA/Projects/InstCombine/Tests/Parse.lean
luisacicolini: updated

tobiasgrosser (approved):

SSA/Projects/InstCombine/Tests/Parse.lean
alexkeizer: That's fine, but can you move the tests to the existing `Print.lean` file, in the `/-! #### add (flags) -/` section?

PR #1635, opened 2025-09-03
feat: add testing for Blase/ multi-width

bollu:
We refactor existing single-width tests in the process to use the multi-width backend.


tobiasgrosser:
Does this drop our test-coverage for the single-width code?

bollu:
No, it just moves single-width to `SingleWidth`, and creates a new `MultiWidth` test

bollu:
Note that multi-width will eventually subsume single-width, at which point I will deprecate the single-width solver.

bollu:
For some reason that is entirely mysterious to me, even though the algorithm is *verified*, it claims that addition is not associative

bollu:
I have no idea what's going on, so I need to think as to whether I proved a theorem statement that is "weaker": I now suspect that the algorithm encoding is sound, but not complete?

My first thought is that due to my encoding of widths, it maybe that we produce SAT queries that checks with 'width' vectors that are binary, but we actually want width vectors that in fact unary! So I may have to `scanAnd`

PR #1636, opened 2025-09-04
feat: update DC semantics to use a common `syncMap` primitive

luisacicolini:
We add an alternative semantics for DC dialect that relies on a common `syncMap` for all operations that involve synchronizing input streams


luisacicolini:
Merging without review given that I'm only adding files concerning DC semantics.

PR #1642, opened 2025-09-10
feat: dataflow example rewrites 

luisacicolini:
We add the dataflow rewrites proposed in https://dl.acm.org/doi/pdf/10.1145/3676641.3715993 (including the additional operators needed by the rewrites) and attempt to prove them in our DC/Handshake framework.


luisacicolini:
merging this given that it's self contained and only touches DC.

PR #1643, opened 2025-09-12
feat: LLVMRiscV evaluation changes

luisacicolini:
This PR updates the evaluation of `LLVMAndRiscV`. 

Change log: 
- adding `passriscv64_optimized` to `Opt`
- add `.gitignore` for `LLVMAndRISCV/Evaluation` so we don't push benchmarks/results/plots 
- updated `benchmarks/README`
- adding benchmarks for `GlobalISel` and for `passriscv64_optimized` to `generate.py`
- remove hard-coded LLVM dir from `generate.py` similarly to #1669 
- updated `mca-analysis/README`
- add plotting scripts for mca analysis to a separate file (`mca-analysis/plot.py`)
- add `GlobalISel` and for `passriscv64_optimized` to `mca-analysis.py`
- enable plotting of all parameters output by the mca analysis (`"tot_instructions", "tot_cycles", "tot_uops"`), to be passed as argument for `mca-analysis/plot.py`
- parsing/plotting all parameters that MCA analysis outputs
- computing similarity of mca analyses outputs (`similarity.py`)

Moreover, all python scripts are `ruff` formatted and checked.

PR #1644, opened 2025-09-16
fix: reverse arguments when printing lists of arguments

luisacicolini:
This PR fixes the order in which arguments are printed by reversing it. 
in this way, the printing is consistent with the way arguments are parsed in Lean-MLIR.

PR 1/? from splitting #1643 

Co-authored-by: Sarah Kuhn <kuhnsa@student.ethz.ch>


tobiasgrosser:
Do we have any kind of testing story for this?

luisacicolini:
@tobiasgrosser : 
Let us consider a test file `test.mlir`
```
{
^bb0(%arg0: i64, %arg1: i64, %arg2: i1):
"llvm.return"(%arg0) : (i64) -> ()
}
 ```
 
On the current main, when we call `lake exe opt --passriscv64 test.mlir` we get
```
builtin.module { 
^bb0(%0 : i1, %1 : i64, %2 : i64):
  "llvm.return"(%0) : (i64) -> ()
 }
```
which is incorrect, because the arguments in the block are printed in the wrong order.

Instead, with this fix we get: 
```
builtin.module { 
^bb0(%0 : i64, %1 : i64, %2 : i1):
  "llvm.return"(%0) : (i64) -> ()
 }
 ```
which is correct.

tobiasgrosser:
> @tobiasgrosser : Let us consider a test file `test.mlir`
> 
> ```
> {
> ^bb0(%arg0: i64, %arg1: i64, %arg2: i1):
> "llvm.return"(%arg0) : (i64) -> ()
> }
> ```
> 
> On the current main, when we call `lake exe opt --passriscv64 test.mlir` we get
> 
> ```
> builtin.module { 
> ^bb0(%0 : i1, %1 : i64, %2 : i64):
>   "llvm.return"(%0) : (i64) -> ()
>  }
> ```
> 
> which is incorrect, because the arguments in the block are printed in the wrong order.
> 
> Instead, with this fix we get:
> 
> ```
> builtin.module { 
> ^bb0(%0 : i64, %1 : i64, %2 : i1):
>   "llvm.return"(%0) : (i64) -> ()
>  }
> ```
> 
> which is correct.

This is perfect. @luisacicolini @alexkeizer, where would we test this best? I feel this can be an easy eval + print + `msg_guard` test case somewhere in the repository, no?

tobiasgrosser:
@luisacicolini, the other test cases in this file still seem to be broken. Can you look into this: toFormat? If not, let me know and I can give it a shot.

luisacicolini:
@tobiasgrosser it all works for me locally in `InstCombine/Tests/Print.lean`, what error do you see exactly?

luisacicolini:
AFAIU, the other tests don't use the function that is impacted by the `reverse`

alexkeizer:
LGTM

PR #1645, opened 2025-09-16
chore: update RISCV64 to use `immediate` instead of `shamt`

luisacicolini:
We update the RISCV64 dialect to use `immediate` values instead of `shamt` when printing attributes. We do that such that XDSL (which we use further down our evaluation infrastructure) can parse the dialect. 

PR 2/? extracted from #1643 

Co-authored-by: Sarah Kuhn <kuhnsa@student.ethz.ch>


SSA/Projects/RISCV64/Base.lean
```diff
@@ -510,26 +510,26 @@ def attributesToPrint: RISCV64.Op → String
...
   | .lui (_imm : BitVec 20) => s!"\{immediate = { _imm.toInt} : i20 } "
-  | .auipc (_imm : BitVec 20) => s!"\{imm = { _imm.toInt} : si20 }" -- adding a s such that xdsl can parsee it, double-check when needed and when not
-  | .slliw (_shamt : BitVec 5) => s!"\{shamt = { _shamt.toInt} : i5 }"
-  | .srliw (_shamt : BitVec 5) => s!"\{shamt = { _shamt.toInt} : i5 }"
-  | .sraiw (_shamt : BitVec 5) => s!"\{shamt = { _shamt.toInt} :i5 }"
-  | .slli (_shamt : BitVec 6) => s!"\{shamt = { _shamt.toInt} : i6 }"
-  | .srli (_shamt : BitVec 6) => s!"\{shamt = { _shamt.toInt} : i6 }"
-  | .srai (_shamt : BitVec 6) => s!"\{shamt = { _shamt.toInt} : i6 }"
+  | .auipc (_imm : BitVec 20) => s!"\{immediate = { _imm.toInt} : si20 }" -- adding a s such that xdsl can parsee it, double-check when needed and when not
+  | .slliw (_imm : BitVec 5) => s!"\{immediate = { _imm.toInt} : i5 }"
```
alexkeizer, line 514: NIT (this was already wrong before this PR, so feel free to ignore):

The leading underscore in `_imm` usually indicates that the variable is unused. Here you do actually use it, so it's better to call the variable `imm`:

```suggestion
  | .auipc (imm : BitVec 20) => s!"\{immediate = { imm.toInt} : si20 }" -- adding a s such that xdsl can parsee it, double-check when needed and when not
  | .slliw (imm : BitVec 5) => s!"\{immediate = { imm.toInt} : i5 }"
```

And the same in the other branches

alexkeizer (commented):
From what I can tell, this changes the printing function, but not the parser, right?

Does the parser understand `immediate`? If not, we probably should keep those in sync, so that we maintain the invariant that the parser can correctly parse the output of `toString`/`toPrint`.

(But this is outside of core, so the bar for quality is a bit lower. Feel free to ignore if it's too much effort; it might be good to at least document the difference, though)

luisacicolini:
AFAIU it's the xDSL parser that can't parse `shamt`

SSA/Projects/RISCV64/Base.lean
tobiasgrosser: I think we can fix this in a separate PR. I will do this to unblock this PR.

tobiasgrosser:
Thank you, @alexkeizer @luisacicolini for the nice review ping-pong.

PR #1647, opened 2025-09-16
refactor: rename `{Ctxt,Valuation}.snoc` to `cons`

alexkeizer:
This PR changes contexts and valuations to be proper cons-lists. Previously, we'd named their "add-one-element" constructor `snoc`, but then we implemented this in terms of `List.cons` and generally were quite inconsistent in not relying on the underlying, e.g. in the GetElem instance or in the fact that we often build Ctxt literals as list literals with a Ctxt constructor.

I decided to stick with the snoc-list perspective, as we'd otherwise have to figure out a solid story for snoc-list literals, and AFAIR, the only reason for having a snoc-list is just for aesthetic purposes -- to match the convention that on pen-paper maths a context grows to the right. Furthermore, moving to the snoc-list perspective for Valuations brings it closer to HVectors, making it easier to unify those two concepts in the future.

The append instance did follow the snoc-list perspective, so I had to rework that.


SSA/Projects/Scf/ScfFunctor.lean
```diff
@@ -553,8 +553,8 @@ theorem Ctxt.Var.toSnoc (ty snocty : Arith.Ty) (Γ : Ctxt Arith.Ty)  (V : Ctxt.V
     {hvproof : Γ[v]? = some ty}
     {var : Γ.Var ty}
     (hvar : var = ⟨v, hvproof⟩) :
-    V var = (V ::ᵥ snocval) ⟨v+1, by simp [hvproof]⟩ := by
-  simp [Ctxt.Valuation.snoc, hvar]
+    V var = (snocval ::ᵥ V) ⟨v+1, by simp [hvproof]⟩ := by
```
bollu: Should the theorem be renamed to `Ctxt.Var.toCons` ?

bollu (commented):
Asked a naming question

SSA/Projects/Scf/ScfFunctor.lean
alexkeizer: This theorem turns out to not be used at all, so I dropped it

PR #1650, opened 2025-09-16
feat: initial setup of a MemorySSA evaluation with register reads/writes

alexkeizer:
This PR adds a very simplified ASL/ISL/Sail-inspired dialect which we can use to evaluate a "MemorySSA"-style approach to side-effectful rewrites.

It also tweaks the `def_signature` macro to allow definition of multiple return types by using a list literal (although the relevant check is a bit fragile, so that, e.g., `| regRead => (.regIndex, .regFile) -> [.bits registerWidth, .regFile]` works but any slight deviation, even just wrapping the list in parentheses -- as in `([.bits registerWidth, .regFile])` -- will cause errors.


.github/workflows/ci-tools.yml
```diff
@@ -65,3 +65,7 @@ jobs:
           lake build AliveExamples
           (cd SSA/Projects/InstCombine/; ./update_alive_statements.py)
           bash -c '! git diff | grep .'  # iff git diff is empty, 'grep .' fails, '!' inverts the failure, and in the forced bash
+
+      - name: Compile ISL
+        run: |
+          lake -R build ISL
```
bollu: @alexkeizer I wonder if we could already start consolidating stuff in `evaluation.yml` if possible?

alexkeizer:
I'm going to merge this so it doesn't bitrot (and because I want to use the generalization of `def_signature` elsewhere, too

PR #1651, opened 2025-09-16
chore: LLVM - first test case for argument order

tobiasgrosser:
Add a test case that confirms the order in which arguments in an LLVM program are printed. The order is seemingly wrong, but for now, we just want to test the status quo.

Open questions:

- This seems to be an issue in generic code. Is there a way to have a generic test case?
- Why are the newlines printed as `\n` instead of an actual newline?


alexkeizer:
> This seems to be an issue in generic code. Is there a way to have a generic test case?

The code is generic, but we still need to instantiate it with a specific dialect to exercise the code.
We could think about defining a simple example dialect in core (either in actual core, or just in Core/Tests), and moving this to the core tests using that dialect, but for now I've just moved the existing tests using LLVM to the InstCombine project.

> Why are the newlines printed as \n instead of an actual newline?

I added test cases for the three ways we can currently print `Com`s (toPrint, toString, and `repr`) and interestingly both `toString` and `toPrint` show the `\n`s but `repr` just prints with newlines. However, when doing `toString <| repr ...`, we also get the `\n`s, so I'm relatively sure this is just a choice of how `#eval` prints it output: for a `String` it will show newlines as `\n`, whereas `Format.line`s get printed as actual newlines.

I agree the `\n`s are a bit ugly, so I wrapped the toPrint and toString tests in a `String.toFormat`, which converts a String to a `Format` and gives the prettier test output/ (We might consider changing `toPrint` to return a `Format`, but that's a different discussion for a future PR).

alexkeizer:
@tobiasgrosser I've taken the liberty of adopting this PR and merging it!

PR #1652, opened 2025-09-17
feat: setup blase CI that runs parallel to other CI only when Blase changes

bollu:
This lets us run `Blase test` only when `Blase` changes, ensuring that our expensive decision procedure tests don't run un-necessarily

### DIFF
--- a/Blase/Blase.lean
+++ b/Blase/Blase.lean
@@ -3,3 +3,4 @@
 import Blase.Fast
 import Blase.SingleWidth.Tactic
 import Blase.MultiWidth.Tactic
+-- Jiggle.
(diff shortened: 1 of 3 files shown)


Blase/Blase.lean
```diff
@@ -3,3 +3,4 @@
 import Blase.Fast
 import Blase.SingleWidth.Tactic
 import Blase.MultiWidth.Tactic
+-- Jiggle.
```
alexkeizer, line 6: @bollu should we drop this jiggle comment?

bollu: @alexkeizer I added it to trigger a CI jiggle, happy to drop it subsquently

PR #1654, opened 2025-09-17
Split Medusa off to a lake project

ineol:


bollu:
This is an LGTM from me. We should also update CI with a separate build from Medusa. Could you do this as well, Leo?

bollu:
LGTM, thanks! Gives us a bit more parallelism in our build, and perhaps more importantly, vendors Medusa as a separate project which helps us develop it such a way that we know it's a library!

PR #1655, opened 2025-09-18
feat: enable multithreading for `run_mca.py`

luisacicolini:
This PR enables running the script `run_mca.py` for the evaluation of the RISCV backend verification multithread.

### DIFF
--- a/SSA/Projects/LLVMRiscV/Evaluation/mca-analysis/run_mca.py
+++ b/SSA/Projects/LLVMRiscV/Evaluation/mca-analysis/run_mca.py
@@ -34,10 +34,16 @@ LOGS_DIR = f"{ROOT_DIR}/SSA/Projects/LLVMRiscV/Evaluation/mca-analysis/results/l
 def create_missing_folders(): 
     if not os.path.exists(MCA_LEANMLIR_DIR):
         os.makedirs(MCA_LEANMLIR_DIR)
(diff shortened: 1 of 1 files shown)


alexkeizer:
This looks good to make use of all cores on a single machine, but it's not quite what is needed to scale over multiple machines. 

You can have a look at the changes I made in https://github.com/opencompl/lean-mlir/pull/1629, where I parallelized the LLVM evaluation using a stride/offset strategy: In that script there was a clear set of files to be processed, so we could parallelize it over multiple machines just by distributing the files to be processed.

To be more precise, each job (i.e., invocation of the script) was passed an `offset` CLI argument that ranged from `0` to `n-1` (inclusive) and a `stride` argument equal to `n`, where `n` is the number of jobs. Each job started with the same total list of all files to be processed, dropped the first `offset`-many files and then only processed every `n`th file, such that combined each file was processed by exactly one of the `n` jobs running.

AFAICT, here too we have a fixed set of files to be processed? So we ought to be able to follow a similar strategy. Does the above make sense? If not, let's jump on a call and pair on it!

alexkeizer:
Actually, I had a go at this myself, at #1688. That PR implements both multi-threading and multi-machine parallelization, so it would supersede this PR.

alexkeizer:
Superseded by #1668

PR #1668, opened 2025-09-19
feat: re-usable taskqueue for easily parallelizing evaluation scripts

alexkeizer:
I've noticed our evaluation scripts tend to have a pretty similar pattern where there is a bunch of files to be processed (usually by calling lean on said file), the results of each file is stored separately, and then we have a separate step which gathers all the data. 

This pattern is quite straightforward to parallelize, yet it seems re-implement our pararellization strategy again and again for each script. Thus, this PR adds a simple task-queue abstraction which we can re-use across all our evaluation scripts. I've tried to thoroughly document the relevant file (taskqueue.py), so do look there for more details!

Furthermore, I've refacted the run_mca.py script to showcase how the taskqueue can be used (although I can't actually run that script locally, so I'm not sure whether I messed that up or not).


evallib/taskqueue.py
```diff
@@ -0,0 +1,129 @@
...
+                for future in futures:
+                    result = future.result()
+                    if result is not None:
+                        print(result)
+            else:
+                with open(output_file, "w") as f_out:
+                    for future in futures:
+                        result = future.result()
+                        if result is not None:
+                            f_out.write(str(result) + '\n')
```
bollu: always use `f_out` and set `f_out` to `stdout` instead of duplicating this logic.

evallib/taskqueue.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def run_tasks(
+        self,
+        tasks: List[T],
+        *,
+        run: Callable[[T], Optional[str]],
+        output_file: Optional[str] = None,
+    ):
+        """
+        `run_tasks` is a light-weight, static distributed task-queue.
```
bollu, line 34: I've read the code, and this really is a *queue*, not a task DAG. I realized when using `snakemake` that the following kinds of jobs also show up:

1) "phony jobs", that need to run first / last (e.g. run `uv --version`, or `lake exe cache get`, or run `plot.py` to establish startup / teardown)
2) jobs that are parametrized by other data. For example, a job that is parametrized by the name of the tactic. I imagine that this in this framework, the parametrization is performed by the script, which creates tasks for each choice of parameters?


I currently don't have a use for dynamic tasks (and indeed, it may be an anti-pattern).

evallib/taskqueue.py
```diff
@@ -0,0 +1,129 @@
...
+T = TypeVar('T')
+
+class TaskQueue:
+    def __init__(self, max_threads: int, stride: int = 1, offset: int = 0, show_progress_bar: bool = True):
+        # Maximum number of threads to use on the current machine
+        self.max_threads = max_threads
+        # Number of tasks to skip between each processed task
+        self.stride = stride
+        # Starting index for processing tasks
+        self.offset = offset
```
bollu, line 21:
```suggestion
        self.offset = offset
        assert self.offset < self.stride
```

We should check that the offset is legal.

bollu (commented):
LGTM, modulo my comment about `stdout`.

evallib/taskqueue.py
alexkeizer: Yeah, the use-case vs snakemake is slightly different: this is not a complete task-runner system, this is basically intended to replace a single hot loop in a script with something that is easily parallelized.

1) If you need to do setup or teardown, just do so imperatively before calling into the taskqueue

2) The "task description" can be arbitrary data, including some object that stores the parameters. What you say is right, all the parametrization is done up-front when generating the task list. See the example in the `run_mca.py` script where we parametrize based on two directories. The assumption is that doing this parametrization is cheap.

In general, this is a pretty low-tech solution intended to easily parallelize our existing evaluation. I think if we run into a situation where these more complex scenarios do come up, going for a `snakemake`-based solution likely makes more sense (is there a straightforward way to do stride/offset based distribution in snakemake?).

alexkeizer:
@luisacicolini could I get an LGTM from you, too? Just to check that the changes made to `run_mca` are sensible

luisacicolini (approved):

PR #1676, opened 2025-09-20
ci: try a simple run on namespace

tobiasgrosser:
We add two setups, a pure namespace-cache-based one and one that uses GitHub Action caches to transfer data between jobs.

# Pure NS Cache Volumes

We currently observe that the second job (Blaze Test) does not reliably pick up the cache that the first job in a given run creates. We want to see if this becomes more robust over time as cache volumes get distributed in the cluster.

# GitHub Action Cache

This works reliably, but has a constant overhead. It would be good if this constant overhead could also disappear, at least when a key is used repeatedly. This might be worth checking with namespace.


.github/workflows/create_evaluation-namespace.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
```
tobiasgrosser:
```suggestion
#!/bin/sh


```

tobiasgrosser:
Let's merge this to get some more experience with namespace runners.

PR #1678, opened 2025-09-20
chore: do not download all of mathlib in docker CI

tobiasgrosser:
Before this change the docker layer for `lake exe cache get; lake build` was 5.98GB. After dropping `lake exe cache get`, we are at 1.84GB. This reduces the "initialize container" time from `1m23s` to `41s`. Given that the lean install itself has 2GB in size, it may be difficult to go a lot below this.


tobiasgrosser:
This has been resolved.

PR #1689, opened 2025-09-22
feat/blase:Convert Parameters into Index in Predicate

bollu:
This is step 1 to support binders (existential and universal quantifiers), for $$\exists (x : \texttt{BV}~w)$$ and associated quantifier elimination.


tobiasgrosser:
Not a single test case?

bollu:
Nothing changed, it's NFC

bollu:
that all the previous tests work without a change *is* the test case

bollu:
It only changes the logical representation in proofs (which changed)

PR #1690, opened 2025-09-22
feat: Blase NS multi-machine batch

bollu:
We will use `snakemake`'s kubernetes executor:
https://snakemake.github.io/snakemake-plugin-catalog/plugins/executor/kubernetes.html

along with namespace's kubernetes cluster support:
- https://namespace.so/docs/reference/github-actions/nscloud-cluster-action

to run our snakemake setup directly on snakemake.
The open question for me is where this will store the build artifact. Currently, our evaluation run uploads a ZIP file to Zulip. We can upload this ZIP file as a build artifact on GH.


bollu:
!blase

bollu:
!blase

PR #1691, opened 2025-09-22
refactor: put dependencies in a separate Docker layer

### LINKED ISSUE #1688: docker build should store mathlib in a separate layer
Currently, docker has to always reload the mathlib layer even if mathlib did not change.

alexkeizer:
This PR optimizes our Dockerfile to create separate layers for dependencies and our code, to optimize cache usage. Fixes #1688.

I initially investigated actually splitting up the build in separate `RUN` command to achieve this layering, but this was fairly complex. Instead, @bollu suggested the current approach:
* We still build everything in a single `RUN` command, but in a separate `builder` phase
* The final image is *not* based on the builder phase, but instead uses a series of `COPY` commands to copy the relevant build artifacts.

This way it's much easier to create layers with just the contents we want!


alexkeizer:
@bollu I've hit merge on this myself, but do have a look at the final version to see if it makes sense!

PR #1696, opened 2025-09-24
feat: fix example where blase preprocessing has bug

bollu:
This was understood by checking examples where we know we must succeed but do not when compared to `bv_normalize`. Note that comparing against `bv_normalize` this way is valuable, as it lets us catch subtle bugs during our reflection/ metaprogramming that would otherwise go unnoticed.

```lean
theorem t7_thm.extracted_1._2 :
    ∀ (x : BitVec 8) (x_1 : BitVec 1) (x_2 : BitVec 8), x_1 = 1#1 → x_2 - 0#8 = 0#8 + x_2 := by
  intros;
  bv_multi_width
```

fails with:

```
Application type mismatch: The argument
  MultiWidth.Term.var ⟨0, ⋯⟩
has type
  MultiWidth.Term
    ((((MultiWidth.Term.Ctx.empty 2).cons (MultiWidth.WidthExpr.var ⟨0, ⋯⟩)).cons
          (MultiWidth.WidthExpr.var ⟨0, ⋯⟩)).cons
      (MultiWidth.WidthExpr.var ⟨1, ⋯⟩))
    ((((MultiWidth.Term.Ctx.empty 2).cons (MultiWidth.WidthExpr.var ⟨0, ⋯⟩)).cons
          (MultiWidth.WidthExpr.var ⟨0, ⋯⟩)).cons
      (MultiWidth.WidthExpr.var ⟨1, ⋯⟩) ⟨0, ⋯⟩)
but is expected to have type
  MultiWidth.Term
    ((((MultiWidth.Term.Ctx.empty 2).cons (MultiWidth.WidthExpr.var ⟨0, ⋯⟩)).cons
          (MultiWidth.WidthExpr.var ⟨0, ⋯⟩)).cons
      (MultiWidth.WidthExpr.var ⟨1, ⋯⟩))
    (MultiWidth.WidthExpr.var ⟨0, ⋯⟩)
in the application
  MultiWidth.Predicate.binRel MultiWidth.BinaryRelationKind.ne (MultiWidth.WidthExpr.var ⟨0, ⋯⟩)
    (MultiWidth.Term.var ⟨0, ⋯⟩)
```

Clearly, this is a reflection bug.


bollu:
Fixed, was, as usual, a De Bruijn indexing bug.

bollu:
I consolidated the "side-channel" information of the BV locations directly into the `TotalOrder` structure that we use to maintain our indexing.

PR #1698, opened 2025-09-24
chore: mark check for changes in AliveStatements as continue-on-error

bollu:
CC @alexkeizer @tobiasgrosser 

This prevents the flakiness of the AliveStatements regression from taking down all of the CI


tobiasgrosser:
That's not the right approach. We want to see if any of this changes. Hiding errors does not achieve this. We should disable parallel evaluation in this file.

tobiasgrosser:
This has meanwhile been solved by disabling parallel evaluation.

PR #1699, opened 2025-09-24
Medusa: Add bitblastable LLVM semantics

ineol:


Medusa/Medusa/LLVM/Semantics.lean
```diff
@@ -0,0 +1,489 @@
...
+deriving Inhabited, DecidableEq
+
+instance: Monad PoisonOr where
+  pure x := ⟨x, false⟩
+  bind x f :=
+    let y := f x.val
+    ⟨y.val, x.poisonous || y.poisonous⟩
+
+def PoisonOr.value {α : Type} (x : α) : PoisonOr α := ⟨x, false⟩ 
+def PoisonOr.poison {α : Type} [Inhabited α] : PoisonOr α := ⟨default, false⟩
```
alexkeizer: This one should be `true`, no?

```suggestion
def PoisonOr.poison {α : Type} [Inhabited α] : PoisonOr α := ⟨default, true⟩ 
```

ineol: Oops! good catch

bollu:
LGTM. @ineol , can you add into the file the commit hash of lean-mlir we took the file at?

PR #1705, opened 2025-09-25
chore: setup Medusa in unified CI

bollu:
This replicates the setup for `Blase` for `Medusa`.


.github/workflows/evaluation.yml
```diff
@@ -140,6 +140,46 @@ jobs:
         run: |
           lake build BlaseTest
 
+  test-blase:
```
ineol: `test-medusa`?

.github/workflows/evaluation.yml
```diff
@@ -140,6 +140,46 @@ jobs:
...
+
+      - name: Load Lean-MLIR from Cache
+        id: cache-lake
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            .lake/packages
+            .lake/build
+            */.lake
+          key: ${{ runner.os }}-lake-packages-${{ hashFiles('lake-manifest.json') }}-${{ github.sha }}
```
ineol, line 169: Is this the right file for medusa?

bollu: We glob all the lakefiles, so.. it should be :D

.github/workflows/evaluation.yml
bollu: Thanks, good catch!

.github/workflows/evaluation.yml
ineol: Even in the `lake-manifest.json` hashing?

PR #1549, opened 2025-08-24
feat: snakemakeify hacker's delight

bollu:
This shows how we would use `snakemake` to gradually migrate our plotting pipeline for hacker's delight.

See that `snakemake` now subsumes the (hard) job of task running in `collect.py`.
It leaves the CSV wranging in `collect.py` and `plot.py` untouched, as these have been vetted by @luisacicolini .

This provides parallelism at the task level, and I would like to believe, is much more declarative on how the hacker's delight files are built.

PR #1706, opened 2025-09-25
feat: `#evaluation in ...` command for easier evaluation

alexkeizer:
This PR adds an `#evaluation in $cmd` command similar to `#guard_msgs in $cmd`, which collects all messages generated by an arbitrary command, and serializes them into a JSON object. The idea is that this will be a building block for any kind of evaluation we'd like to do, to easily have existing tactics output JSONL without having to modify those tactics.

I was originally planning to use this to convert the existing LLVM evaluation to JSONL, but it seems that evaluation has fundamentally changed. @bollu is the LLVM evaluation superseded by your evaluation?

TODO:
- [ ] add a CI job which builds & tests `EvaluationHarness`


alexkeizer:
@bollu I added the options you mentioned, could you have a look to see if this is usable in its current form, or whether we need more?

bollu:
@alexkeizer No, the LLVM evaluation has not been changed by me. What's in the repo is the "latest". I attempted to improve it using `snakemake` here #1549 , but that's only the hacker's delight portion, and not the `InstCombine` portion.

alexkeizer:
> @alexkeizer No, the LLVM evaluation has not been changed by me. What's in the repo is the "latest". I attempted to improve it using `snakemake` here #1549 , but that's only the hacker's delight portion, and not the `InstCombine` portion.

So when I open, say, [`gzext_proof.lean`](https://github.com/opencompl/lean-mlir/blob/main/SSA/Projects/InstCombine/tests/proofs/gzext_proof.lean), I see the following:
```lean
theorem test_sext_zext_thm (e : IntW 16) : sext 64 (zext 32 e) ⊑ zext 64 e := by
    simp_alive_undef
    simp_alive_ops
    simp_alive_case_bash
    simp_alive_split
    extract_goals
    all_goals sorry
```

I assumed the `extract_goals` there was your doing, and I wonder if that is what's broken the evaluation CI; notice how it now just reports `0` everywhere:

````quote
bv_decide solved 0 theorems.
bitwuzla solved 0 theorems.
bv_decide found 0 counterexamples.
bitwuzla found 0 counterexamples.
bv_decide only failed on 0 problems.
bitwuzla only failed on 0 problems.
````

PR #1707, opened 2025-09-25
feat: cache-mount all subproject .lake folders in Docker build

### LINKED ISSUE #1687: docker builds do not cache blase files
The following build is an example: https://github.com/opencompl/lean-mlir/actions/runs/17913400098/job/50929896442

alexkeizer:
This PR tweaks the Docker build to enable incremental caching (via cache-mount) of the .lake folder of all sub-projects as well.

In addition, we move the docker caching setup and teardown to a dedicated shell script, in a new `.dockerscripts` folder, to reduce caching-related noise in the Dockerfile.

This fixes #1687.

### DIFF
--- a/.dockerscripts/cache.sh
+++ b/.dockerscripts/cache.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 if [[ -z "$CACHE_MOUNT" ]]; then
-    echo "error: \$CACHE not set"
+    echo "error: \$CACHE_MOUNT not set"
     exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


.dockerscripts/cache.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+if [[ -z "$CACHE_MOUNT" ]]; then
+    echo "error: \$CACHE not set"
```
ineol: Should it be `$CACHE_MOUNT` or is this github magic?

ineol (commented):
Otherwise LGTM!

.dockerscripts/cache.sh
alexkeizer: In the error message? Yeah, that should've been CACHE_MOUNT, thanks for catching!

PR #1710, opened 2025-09-25
refactor: promote LLVM dialect to core

alexkeizer:
This PR moves the LLVM dialect (but **not** the InstCombine evaluation) to the LeanMLIR core project.

I've made the necessary changes to generated evaluation files via `sed`, but also updated the generating scripts so things should still be in sync.


bollu:
@alexkeizer thanks for fixing up the scripts. You have my LGTM on the goal extraction, provided the `extract_goals` CI passes!

PR #1711, opened 2025-09-25
chore: drop Experimental folder

alexkeizer:
All files in this folder were either MWEs for external bug reports or experimental files that haven't been touched in ages, have since bitrotted and no longer build on the Lean version we currently use. Thus, it seems cleaner to just delete these files.


alexkeizer (commented):
@bollu, can you confirm the Bits/Fast stuff is old and there's no need to keep it around. Those are the only files that seemed like they might still be relevant (besides the fact that they, too, failed to build on our current Lean version)

bollu:
Yes, confirmed.

PR #1712, opened 2025-09-26
feat: add `select_constant_cmp` rewrite pattern

osmanyasar05:
This PR adds a new rewrite pattern, [select_constant_cmp](https://github.com/llvm/llvm-project/blob/cae73be8a6c2d39e5a827fd31848676d52062a94/llvm/include/llvm/Target/GlobalISel/Combine.td#L518). 

The pattern rewrites cases where select is given constant inputs (_true_ or _false_), simplifying it to the first or second input value accordingly.

In the current test case, the expected output still contains the input given to the rewriter. Ideally, this should be eliminated by DCE, which is not applied in this PR.

Co-authored-by: Luisa Cicolini luisacicolini@gmail.com

### DIFF
--- a/SSA/Projects/LLVMRiscV/Pipeline/Combiners.lean
+++ b/SSA/Projects/LLVMRiscV/Pipeline/Combiners.lean
@@ -135,12 +135,10 @@ def select_same_val_self : LLVMPeepholeRewriteRefine 64 [Ty.llvm (.bitvec 1), Ty
 def select_same_val : List (Σ Γ, LLVMPeepholeRewriteRefine 64  Γ) :=
   [⟨_, select_same_val_self⟩]
 
-
 /-- ### select_constant_cmp
   (true ? x : y) -> x
   (false ? x : y) -> y
 -/
-
 def select_constant_cmp_true : LLVMPeepholeRewriteRefine 64 [Ty.llvm (.bitvec 64), Ty.llvm (.bitvec 64)] where
   lhs := [LV| {
   ^entry (%x: i64, %y: i64):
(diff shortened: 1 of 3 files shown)


luisacicolini (changes requested):
This looks great, thank you! Could you add the new `Combiners` test to `Tests` as well? (like `Pipeline` is imported)

luisacicolini:
Also, following @alexkeizer's suggestion: could we rewrite the example applying DCE?

osmanyasar05:
I added the DCE as suggested, now the issue is that the constant value is not eliminated by DCE. Could it be that it needs one more DCE pass? @alexkeizer

SSA/Projects/LLVMRiscV/Pipeline/Combiners.lean
```diff
@@ -135,6 +135,38 @@ def select_same_val_self : LLVMPeepholeRewriteRefine 64 [Ty.llvm (.bitvec 1), Ty
 def select_same_val : List (Σ Γ, LLVMPeepholeRewriteRefine 64  Γ) :=
   [⟨_, select_same_val_self⟩]
 
+/-- ### select_constant_cmp
+  (true ? x : y) -> x
+  (false ? x : y) -> y
+-/
```
alexkeizer: /-- indicates that this is a docstring, meaing that the text in this comment is associated with the next definition and shows up in documentation for that definition (e.g., when hovering over it in VSCode).

* We generally don't include headers (the line with `###`) in docstrings, as those are mostly relevant when reading the source file, but aren't very informative in a definition-hover. Thus, we generally put headers in a `/-! ... -/` comment, these ones still show up in generated docs for the file, but aren't associated with any definition.
* Try to explain the definition in prose that makes sense even when not looking at the source code (as the source code won't be shown in a hover).

In this case, it's just a test-case, so instead of polishing the docstring to the usual expectation, we could just downgrade the comment to a regular comment (/- ... -/), those won't show up any generated documentation.

```suggestion
/-! ### select_constant_cmp -/

/- 
Test the rewrite:
  (true ? x : y) -> x
  (false ? x : y) -> y
-/
```

luisacicolini: @alexkeizer my bad, I suggested using `/--` for consistency with the rest of the file, but you are right - I will open a PR to refactor all comments

luisacicolini (approved):

PR #1720, opened 2025-09-29
feat: prove maxu and minu

osmanyasar05:
This PR introduces proofs for theorems ZBB_RTYPE_pure_RISCV_MAXU_eq_ZBB_RTYPE_pure_RISCV_MAXU_bv and ZBB_RTYPE_pure_RISCV_MINU_eq_ZBB_RTYPE_pure_RISCV_MINU_bv.

### DIFF
--- a/SSA/Projects/RISCV64/Semantics.lean
+++ b/SSA/Projects/RISCV64/Semantics.lean
@@ -992,8 +992,9 @@ def ZBB_RTYPE_pure_RISCV_MINU_bv (rs2_val : BitVec 64) (rs1_val : BitVec 64) : B
 theorem ZBB_RTYPE_pure_RISCV_MINU_eq_ZBB_RTYPE_pure_RISCV_MINU_bv (rs2_val : BitVec 64) (rs1_val : BitVec 64) :
     ZBB_RTYPE_pure_RISCV_MINU rs2_val rs1_val = ZBB_RTYPE_pure_RISCV_MINU_bv rs2_val rs1_val := by
   unfold ZBB_RTYPE_pure_RISCV_MINU ZBB_RTYPE_pure_RISCV_MINU_bv
-  simp
-  sorry
+  simp only [min_def, ule_iff_toNat_le]
+  norm_cast
+  split_ifs <;> simp
 
 def ZBB_RTYPE_pure_RISCV_MAXU_bv (rs2_val : BitVec 64) (rs1_val : BitVec 64) : BitVec 64 :=
   BitVec.extractLsb' 0 64 (if BitVec.ult rs1_val rs2_val then rs2_val else rs1_val)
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
Nice. 👍

tobiasgrosser:
Let's add @salinhkuhn to the reviewers of such patches.

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -992,17 +992,25 @@ def ZBB_RTYPE_pure_RISCV_MINU_bv (rs2_val : BitVec 64) (rs1_val : BitVec 64) : B
...
     ZBB_RTYPE_pure_RISCV_MAXU rs2_val rs1_val =ZBB_RTYPE_pure_RISCV_MAXU_bv rs2_val rs1_val := by
   unfold ZBB_RTYPE_pure_RISCV_MAXU ZBB_RTYPE_pure_RISCV_MAXU_bv
-  simp
-  sorry
+  simp only [max_def, ult_iff_toNat_lt]
+  norm_cast
+  split_ifs <;> simp
+  case neg h1 h2 =>
+    have h_le : rs2_val.toNat ≤ rs1_val.toNat := Nat.ge_of_not_lt h2
+    rw[toNat_eq]
```
salinhkuhn: whitespace :) btw rw and theorems

salinhkuhn (commented):
just the whitespace else lgtm !

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -992,17 +992,25 @@ def ZBB_RTYPE_pure_RISCV_MINU_bv (rs2_val : BitVec 64) (rs1_val : BitVec 64) : B
...
 theorem ZBB_RTYPE_pure_RISCV_MAXU_eq_ZBB_RTYPE_pure_RISCV_MAXU_bv (rs2_val : BitVec 64) (rs1_val : BitVec 64) :
     ZBB_RTYPE_pure_RISCV_MAXU rs2_val rs1_val =ZBB_RTYPE_pure_RISCV_MAXU_bv rs2_val rs1_val := by
   unfold ZBB_RTYPE_pure_RISCV_MAXU ZBB_RTYPE_pure_RISCV_MAXU_bv
-  simp
-  sorry
+  simp only [max_def, ult_iff_toNat_lt]
+  norm_cast
+  split_ifs <;> simp
+  case neg h1 h2 =>
+    have h_le : rs2_val.toNat ≤ rs1_val.toNat := Nat.ge_of_not_lt h2
```
luisacicolini: This should not be necessary AFAICT, have you tried using `omega` after rewriting?

osmanyasar05: yep, it works. thanks for the suggestion!

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -992,17 +992,23 @@ def ZBB_RTYPE_pure_RISCV_MINU_bv (rs2_val : BitVec 64) (rs1_val : BitVec 64) : B
...
   unfold ZBB_RTYPE_pure_RISCV_MAXU ZBB_RTYPE_pure_RISCV_MAXU_bv
-  simp
-  sorry
+  simp only [max_def, ult_iff_toNat_lt]
+  norm_cast
+  split_ifs <;> simp
+  case neg h1 h2 =>
+    bv_omega
+  case pos h3 h4 =>
+    bv_omega
```
luisacicolini: you can also use `<;> bv_omega`, since the same tactic solves both cases. "`<;> tactic`" is basically saying "apply `tactic` to any goal you find here" (e.g. after `split_ifs`: `<;> simp` applies the `simp` tactic at every goal)

osmanyasar05: ah yes i missed that one. thank you!

SSA/Projects/RISCV64/Semantics.lean
```diff
@@ -992,17 +992,19 @@ def ZBB_RTYPE_pure_RISCV_MINU_bv (rs2_val : BitVec 64) (rs1_val : BitVec 64) : B
...
   BitVec.extractLsb' 0 64 (if BitVec.ult rs1_val rs2_val then rs2_val else rs1_val)
 
 theorem ZBB_RTYPE_pure_RISCV_MAXU_eq_ZBB_RTYPE_pure_RISCV_MAXU_bv (rs2_val : BitVec 64) (rs1_val : BitVec 64) :
     ZBB_RTYPE_pure_RISCV_MAXU rs2_val rs1_val =ZBB_RTYPE_pure_RISCV_MAXU_bv rs2_val rs1_val := by
   unfold ZBB_RTYPE_pure_RISCV_MAXU ZBB_RTYPE_pure_RISCV_MAXU_bv
-  simp
-  sorry
+  simp only [max_def, ult_iff_toNat_lt]
+  norm_cast
+  split_ifs <;> simp <;> bv_omega
```
luisacicolini: One last thing: having non-terminal `simp` makes our design brittle, and we should avoid that. The theorem that `simp` applies can change and include other theorems that might change the result of the rewrite and potentially make the subsequent tactic (in this case `bv_omega`) unable to close the goal. To avoid that, if the `simp` is not terminal in the proof, we want to specify _exactly_ which theorems it applies. You can see this by typing `simp?`: the infoview will show you the set applied and you'll have something like `simp [lemma1, lemma2, lemma3]` at the end. Because different branches of the if can use different theorems, you might need an specific branch for each of them (i.e., without `<;>`).

luisacicolini (approved):

PR #1722, opened 2025-09-30
refactor: unify the three ways to print a `Com`

alexkeizer:
This PR changes the `ToString` and `Repr` instances for `Com` to be defined in terms of the `DialectPrint` typeclass, re-using the implementation of `Com.print` for those instances.


SSA/Projects/PaperExamples/PaperExamples.lean
```diff
@@ -30,20 +31,18 @@ instance : TyDenote Ty where
...
   Ty := Ty
 
-instance : ToString Ty where
-  toString t := repr t |>.pretty
-
 instance : DialectToExpr Simple where
   toExprM := .const ``Id [0]
   toExprDialect := .const ``Simple []
 
+
```
alexkeizer:
```suggestion
```

PR #1725, opened 2025-09-30
feat/blase: Existential quantifier elimination for bitvector literals

bollu:
This PR adds existential variable quantifier elimination. The idea is to convert to a NFA, which is then determinised to a DFA. This gives us quantifier elimination, with the expected exponential blowup at the NFA -> DFA conversion

### DIFF
--- a/Blase/Blase/Fast/Circuit.lean
+++ b/Blase/Blase/Fast/Circuit.lean
@@ -552,7 +552,7 @@ def bAnd : ∀ (_s : List α) (_f : α → Circuit β), Circuit β
     rw [bAnd, eval_foldl_and]; simp
 
 /-- In a sum type, assign variables to the right of the circuit signature. -/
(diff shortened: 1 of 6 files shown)


bollu:
After working on this more, I come to the conclusion that I need to be able to manipulate a different encoding of automata, where the next state is not necessarily a function of the next state, but a mere relation. So, instead of encoding

```lean
structure FSM (arity : Type) : Type 1 where
  ...
  nextStateCirc : α → Circuit (α ⊕ arity) -- External part in lean, internal part in Circuit
```

We should instead have:

```lean
structure FSM (arity : Type) : Type 1 where
  ...
  nextStatesCirc :  Circuit (α ⊕ arity ⊕ α) -- This is now fully in Circuit.
```

This will me us to build an automata for nondeterministic state transition. But the key is that there is no longer anything at the Lean-level, but rather, the entire next state computation lives as a single circuit.

PR #1726, opened 2025-10-01
feat: prove max and min

osmanyasar05:
This PR introduces proofs for theorems ZBB_RTYPE_pure_RISCV_MAX_eq_ZBB_RTYPE_pure_RISCV_MAX_bv and ZBB_RTYPE_pure_RISCV_MIN_eq_ZBB_RTYPE_pure_RISCV_MIN_bv.

The max_bv implementation now uses `sle` instead of `slt`, since the [definition of max](https://github.com/leanprover-community/mathlib4-nightly-testing/blob/f4019d6363fa9d91cc93c77858a66a5535254a8c/Mathlib/Order/Defs/LinearOrder.lean#L138) in Mathlib is defined with "less than or equal to", rather than "less than":
`lemma max_def (a b : α) : max a b = if a ≤ b then b else a := by rw [LinearOrder.max_def a]`


luisacicolini:
@osmanyasar05 can you also add to the commit message the change you added wrt. `sle`/`slt` with ideally one example why this was wrong previously?

PR #1729, opened 2025-10-02
chore: change font size in plots

luisacicolini:
This PR updates the formatting of plots we produce (size of the font, shape of the plot), adds the plotting of a bar plot and the extraction of some useful numbers (average overhead etc) to use in the paper. Will merge without review as it only updates the plottin script.


luisacicolini:
merging without review since these are strict formatting changes.

PR #1732, opened 2025-10-02
feat/medusa: Use Fp.lean definitions in FpExpr.eval

bollu:
This really connects whatever Medusa has with the associated Fp.lean semantics. All hell will break loose when we start to use these, but that's the fun of it!


bollu:
Merging to allow debugging of the FP.lean crash in a separate PR.

PR #1742, opened 2025-10-07
feat: integrate `mlir-fuzz` in LLVMAndRiscV evaluation

luisacicolini:
This PR adds `mlir-fuzz` as a dependency for the evaluation of LLVMAndRiscV. We integrate it in the generation of the benchmarks to evaluate programs with a certain number of operations and update the rest of the evaluation pipeline accordingly.

In particular, the folder `MLIR_multi` containing the initially generated synthetic programs now contains one file per each size from #ops = 3 to #ops = 8, each containing 1000 programs (as per `generate_multi.py`). For example, `output_3.mlir` contains 1000 synthesized programs with size 3.


luisacicolini:
As usual, given that this PR only concerns the evaluation of LLVMAndRiscV I will merge once CI passes.

PR #1743, opened 2025-10-07
feat: update `mlir-fuzz` benchmarks

luisacicolini:
This PR updates the benchmarks generated with `mlir-fuzz` with `random` strategy and the re-generated benchmarks.
We also support specifying different optimization flags for LLVM.


luisacicolini:
If CI passes I will merge this, given that I'm only changing the python script and pushing a dataset from mlir-fuzz

PR #1744, opened 2025-10-08
feat: add `hoist_logic_op_with_same_opcode_hands` rewrite pattern

osmanyasar05:
This PR adds the rewrite pattern [hoist_logic_op_with_same_opcode_hands](https://github.com/llvm/llvm-project/blob/7e3080f0c197316ff06725c446bc59e67f80f069/llvm/include/llvm/Target/GlobalISel/Combine.td#L754).

These patterns require checking that arguments are used only once, which is currently not supported in lean-mlir. Without this check, using these rewrites in larger programs may increase the number of instructions.


luisacicolini (approved):
thank you for the pr, lgtm!

PR #1756, opened 2025-10-08
Blase: Add a tactic to genralize the widths of bitvectors

ineol:
It seems to be working on a small example. 

It needs to special case functions such as `zeroExtend` to recognize that some parameters are "width" parameters.

### DIFF
--- a/Blase/Blase/MultiWidth/Tactic.lean
+++ b/Blase/Blase/MultiWidth/Tactic.lean
@@ -1093,6 +1093,126 @@ def evalBvMultiWidth : Tactic := fun
     solveEntrypoint g cfg
 | _ => throwUnsupportedSyntax
 
+/-
+A tactic to generalize the width of BitVectors
+-/
+
+structure State where
+  /-- Maps fixed width to a new MVar for the generic width. -/
+  mapping : DiscrTree Expr
+  invMapping : Std.HashMap Expr Expr
+  deriving Inhabited
+
+abbrev GenM := StateT State TermElabM 
+
+def State.get? (e : Expr) : GenM (Option Expr) := do
+  let s ← get
+  match ← s.mapping.getMatch e with
+  | #[x] => return x
+  | #[] => return none
+  | _ => unreachable!
+
+def State.setMapping (e x : Expr) : GenM Unit := do
+  let s ← get
+  let m ← s.mapping.insert e x
+  set {s with mapping := m}
+
+/-- Get the generic width BV MVar corresponding to an existing BV width. -/
+def State.add? (e : Expr) : GenM Expr := do
+  match ← get? e with
+  | some x => pure x
+  | none =>
+    if e.isFVar || e.isBVar then pure e else
+    let x ← mkFreshExprMVar (some (.const ``Nat [])) (userName := `w)
+    setMapping e x
+    modify fun s => { s with invMapping := s.invMapping.insert x e }
+    pure x
+
+/--
+This table determines which arguments of important functions are bitwidths and
+should be generalized and which ones are normal parameters which should be
+recursively visited.
+-/
+def genTable : Std.HashMap Name (Array Bool) := Id.run do
+  let mut table := .emptyWithCapacity 16
+  table := table.insert ``BitVec #[true]
+  table := table.insert ``BitVec.zeroExtend #[true, true, false]
+  table := table.insert ``BitVec.signExtend #[true, true, false]
+  table := table.insert ``BitVec.instAdd #[true]
+  table := table.insert ``BitVec.instSub #[true]
+  table := table.insert ``BitVec.instMul #[true]
+  table := table.insert ``BitVec.instDiv #[true]
+  table
+
+partial def visit (t : Expr) : GenM Expr := do
+  let t ← instantiateMVars t
+  match t with
+  | .app _ _ =>
+    let f := t.getAppFn
+    let args := t.getAppArgs
+    let table := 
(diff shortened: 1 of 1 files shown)


ineol:
@bollu it seems to be working now

Blase/Blase/MultiWidth/Tactic.lean
```diff
@@ -1093,6 +1093,123 @@ def evalBvMultiWidth : Tactic := fun
     solveEntrypoint g cfg
 | _ => throwUnsupportedSyntax
 
+/-
+A tactic to generalize the width of BitVectors
+-/
+
+structure State where
+  mapping : Std.HashMap Expr Expr
```
bollu:
```suggestion
  /-- Maps fixed width BV to a new MVar of generic width. -/
  mapping : Std.HashMap Expr Expr
```

Blase/Blase/MultiWidth/Tactic.lean
```diff
@@ -1093,6 +1093,123 @@ def evalBvMultiWidth : Tactic := fun
...
+  deriving Inhabited
+
+abbrev GenM := StateT State TermElabM 
+
+def State.get? (e : Expr) : GenM (Option Expr) := do
```
bollu, line 1108:
```suggestion
/-- Get the generic width BV MVar corresponding to an existing BV. -/
def State.get? (e : Expr) : GenM (Option Expr) := do
```

Blase/Blase/MultiWidth/Tactic.lean
```diff
@@ -1093,6 +1093,123 @@ def evalBvMultiWidth : Tactic := fun
...
+  let s ← get
+  for (e', x) in s.mapping do
+    if ← isDefEq e e' then
+      return x
```
bollu:
```suggestion
  /-- TODO: Instead of using a HashMap, consider using a DiscrTree. -/
  for (e', x) in s.mapping do
    if ← isDefEq e e' then
      return x
```

Blase/Blase/MultiWidth/Tactic.lean
```diff
@@ -1093,6 +1093,123 @@ def evalBvMultiWidth : Tactic := fun
...
+    let bv? (n : Nat) :=
+      match table with
+      | .some xs => xs.getD n false
+      | .none => false
+    /- let f ← visit f -/
+    args.zipIdx.foldlM (init := f) fun res (arg, i) => do
+      let arg ← if bv? i then State.add? arg else visit arg
+      pure <| .app res arg
+  | .forallE n e₁ e₂ info =>
+    pure <| .forallE n (← visit e₁) (← visit e₂) info
```
bollu, line 1166: Why don't you use the combinators `forallTelescoping` and instead work with raw BVars? I guess in this case it's OK, but I do wonder why you prefer the approach :)

Blase/Blase/MultiWidth/Tactic.lean
```diff
@@ -1093,6 +1093,123 @@ def evalBvMultiWidth : Tactic := fun
...
+def genTable : Std.HashMap Name (Array Bool) := Id.run do
+  let mut table := .emptyWithCapacity 16
+  table := table.insert ``BitVec #[true]
+  table := table.insert ``BitVec.zeroExtend #[true, true, false]
+  table := table.insert ``BitVec.signExtend #[true, true, false]
+  table := table.insert ``BitVec.instAdd #[true]
+  table := table.insert ``BitVec.instSub #[true]
+  table := table.insert ``BitVec.instMul #[true]
+  table := table.insert ``BitVec.instDiv #[true]
+  table
```
bollu, line 1145: Don't we want to *always* generalize a BV variable? My intuition is that instead of having a table, we check if a value has type `BitVec w`, and if it is, we generalize it, with a possible exception for `BitVec 1`?

bollu (approved):
While I agree that this is a viable approach, I had a different algorithm in mind:

At each function application `f x1 ... xn`, infer the type of the function as `f : T1 -> T2 .. -> Tn -> O`. Inspect the `(xi, ti)`, and if the type `ti` is a `BitVec K`, we perform the generalization as you did. This avoids the need for a table, but I guess I'm now not sure about the tradeoffs between having a table and not having one.





Anyway, LGTM to me! I wrote down some comments, but none of them are blockers to merging.

Thanks muchly ^_^

Blase/Blase/MultiWidth/Tactic.lean
ineol: The problem is that we want to generalize `10` in `x.signExtend 10` but it's only because we know that the first parameter of `signExtend` is a width and not a random Nat.

A more rigorous approach would be to analyze at the type of `signExtend` and  see that the variable appears as a paramter of `BitVec` and recover the information in the table like that, but it seems complicated.

ineol: I just read your comment above, so we agree on the method :)

I think we can try this simple approach and see if it's sufficient for the evaluation.

Blase/Blase/MultiWidth/Tactic.lean
ineol: Hmm we want to recurse in `e1` so we'd need to first change the types in the telescope and then use `forallTelescope` right?

Blase/Blase/MultiWidth/Tactic.lean
ineol: Hmm is there a good way to recover the original expression from the keys of the tree?

PR #1763, opened 2025-10-10
feat: add `mulo_by_2` rewrite pattern

osmanyasar05:
This PR adds the rewrite  pattern [mulo_by_2](https://github.com/llvm/llvm-project/blob/1d3ad74667e86a1ba19285e58d14156cce3de89d/llvm/include/llvm/Target/GlobalISel/Combine.td#L1266).

### DIFF
--- a/SSA/Projects/LLVMRiscV/Pipeline/Combiners.lean
+++ b/SSA/Projects/LLVMRiscV/Pipeline/Combiners.lean
@@ -1004,7 +1004,7 @@ Test the rewrite:
   (G_UMULO x, 2) -> (G_UADDO x, x)
   (G_SMULO x, 2) -> (G_SADDO x, x)
 -/
-def mulo_by_2 : LLVMPeepholeRewriteRefine 64 [Ty.llvm (.bitvec 64)] where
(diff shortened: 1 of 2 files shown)


osmanyasar05:
@luisacicolini do we need 2 separate patterns for matching unsigned/signed overflow flags, or would it work like the current version as well? I mean, instead of this:
`%0 = llvm.mul %x, %c overflow<nsw, nuw> : i64`

Do we need two different patterns with these:
`%0 = llvm.mul %x, %c overflow<nsw> : i64`
`%0 = llvm.mul %x, %c overflow<nuw> : i64`

luisacicolini:
thank you for the PR! Could you please merge main and make sure it builds?
And yes, I would like to have a rewrite for each possible flag, thank you!

osmanyasar05:
@luisacicolini thanks for the review, I added the other flags as well.

osmanyasar05:
Rewrites for other flags and their corresponding test cases have been added, PR is ready @tobiasgrosser

tobiasgrosser:
Cool. Let's get this merged then!

PR #1766, opened 2025-10-12
fix: replace SSA.Core import with LeanMLIR import

luisacicolini:


alexkeizer:
@luisacicolini turns out we had an old import referring to `SSA.Core` left, which I've now fixed to be `LeanMLIR`. I co-opted this PR to fix it in, so that your MWE now turns into a regression test.

PR #1767, opened 2025-10-12
Add +specialize option to `bv_generalize`

ineol:
This allows to test with `bv_decide` whether a predicate is true in its generalize form with reasonably high confidence.

### DIFF
--- a/Blase/Blase/MultiWidth/Tactic.lean
+++ b/Blase/Blase/MultiWidth/Tactic.lean
@@ -1138,6 +1138,7 @@ def genTable : Std.HashMap Name (Array Bool) := Id.run do
   table := table.insert ``BitVec #[true]
   table := table.insert ``BitVec.zeroExtend #[true, true, false]
   table := table.insert ``BitVec.signExtend #[true, true, false]
+  table := table.insert ``BitVec.instOfNat #[true, false, false]
   table := table.insert ``BitVec.instAdd #[true]
   table := table.insert ``BitVec.instSub #[true]
   table := table.insert ``BitVec.instMul #[true]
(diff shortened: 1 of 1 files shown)


Blase/Blase/MultiWidth/Tactic.lean
```diff
@@ -1202,17 +1227,31 @@ def evalBvGeneralize : Tactic := fun
...
 theorem test_bv_generalize (x y : BitVec 32) (zs : List (BitVec 44)) (z : BitVec 10) (h : 52 + 10 = 42) (heq : x = y) : 
     x.zeroExtend 10 = y.zeroExtend 10 + 0 := by
   bv_generalize
   bv_multi_width
 
+theorem test_bv_generalize_spec (x y : BitVec 32) (zs : List (BitVec 44)) (z : BitVec 10) (h : 52 + 10 = 42) (heq : x = y) : 
+    x.zeroExtend 10 = y.zeroExtend 10 + 0 := by
+  bv_generalize +specialize
+  bv_decide
+
```
bollu, line 1255: Nice, thanks for the tets :D

Blase/Blase/MultiWidth/Tactic.lean
```diff
@@ -1183,6 +1184,29 @@ def doBvGeneralize (g : MVarId) : GenM (Expr × MVarId) := do
...
+def specializeGoal (g : MVarId) (lengthCount : Nat) : TacticM Unit := do
+  let t ← g.getType
+  let newT ← forallTelescope t fun xs t => do
+    let mut t := t
+    let mut substs := FVarSubst.empty
+    for i in [0:lengthCount] do
+      let n := 2 * i + 1
+      substs := substs.insert xs[i]!.fvarId! (mkNatLit n)
+    let ys := xs.drop lengthCount
+    let newT ← mkForallFVars ys t (binderInfoForMVars := .default)
```
bollu, line 1198: I don't really understand the invariants here: Does it assume that the `BitVec w` occur first, and is not interleaved with other `fvars`? What happens if I have `BitVec w, Int, BitVec w`?

bollu (commented):
I left a comment about a corner case in the algorithm that I don't undertsand. @ineol if you believe that this corner case doesn't occur / I missed an invariant that prevents this, then LGTM!

Blase/Blase/MultiWidth/Tactic.lean
ineol: Since the goal is generated by the tactic itself we know that the first `lengthCount` variables are the width variables

PR #1768, opened 2025-10-15
feat: add `llvm.freeze` instruction

osmanyasar05:
This PR adds the implementation of [llvm.freeze](https://llvm.org/docs/LangRef.html#i-freeze) instruction. llvm.freeze instruction stops the propagation of poison values. If the argument is poison, returns the default value, else returns the input argument.


LeanMLIR/LeanMLIR/Dialects/LLVM/Basic.lean
```diff
@@ -181,6 +182,7 @@ namespace MOp
 @[match_pattern] def neg    (w : Width φ) : MOp φ := .unary w .neg
 @[match_pattern] def not    (w : Width φ) : MOp φ := .unary w .not
 @[match_pattern] def copy   (w : Width φ) : MOp φ := .unary w .copy
+@[match_pattern] def freeze     (w : Width φ) : MOp φ := .unary w .freeze
```
tobiasgrosser:
```suggestion
@[match_pattern] def freeze (w : Width φ) : MOp φ := .unary w .freeze
```

luisacicolini (commented):
thanks for the PR! 
could you fix the build error? It seems like the `opt` tool can't be built.
also, I have a question for my own understanding: how do we pick the default value to return in case of poison?

LeanMLIR/LeanMLIR/Dialects/LLVM/Basic.lean
```diff
@@ -297,6 +301,7 @@ namespace LLVM.Op
 @[match_pattern] abbrev select : Nat → LLVM.Op := MOp.select ∘ .concrete
 @[match_pattern] abbrev neg    : Nat → LLVM.Op := MOp.neg    ∘ .concrete
 @[match_pattern] abbrev copy   : Nat → LLVM.Op := MOp.copy   ∘ .concrete
+@[match_pattern] abbrev freeze : Nat → LLVM.Op := MOp.freeze  ∘ .concrete
```
alexkeizer, line 304:
```suggestion
@[match_pattern] abbrev freeze : Nat → LLVM.Op := MOp.freeze ∘ .concrete
```

LeanMLIR/LeanMLIR/Dialects/LLVM/Semantics.lean
```diff
@@ -632,4 +632,10 @@ def sext {w: Nat} (w': Nat) (x: IntW w) : IntW w' := do
   let x' <- x
   sext? w' x'
 
+@[simp_llvm_option]
+def freeze (x: IntW w) : IntW w := do
+  match x with
```
alexkeizer, line 637: If you match directly on the argument of a function, there is a short-hand version where you move the argument to the right of the colon, and remove the `:=`, then you can write the function directly in terms of patternmatching, without writing `match _ with`.
```suggestion
def freeze : IntW w -> IntW w
| poison => value 0
| value a => value a
```
NOTE: you'll want to replace `->` with the unicode arrow, but I couldn't write that easily in GHs UI.

Also note that the brackets you wrote are redundant, so it's cleaner to leave them out

osmanyasar05:
The build issue is now fixed by adding unrealized conversion casts, and using the value `0` for frozen poison values. @tobiasgrosser

tobiasgrosser:
Thank you. I merged this.

alexkeizer (commented):
I see this got merged already. I still left some comments, could you address those in a follow-up PR?

Also, could you add some tests for this new freeze instruction in `LeanMLIR/LeanMLIR/Tests/Dialects/LLVM/Print.lean`? Doesn't have to be a lot of tests, just a simple program that has a single freeze instruction, have a look at the other operations being tested in that file!

LeanMLIR/LeanMLIR/Dialects/LLVM/Semantics.lean
alexkeizer, line 637: IIUC, the langref says that freeze returns an *arbitrary* value, meaning that these semantics of freeze here are not accurate: the real semantics would have to be non-deterministic! 

Now, I would like to understand why we need to add `freeze` before making a move to non-deterministic semantics, so for now this simplification is fine, but I would like to see a nice and big doc-string on this function that very explicitly makes clear that these semantics are not accurate w.r.t. the langref, that the real semantics ought to be non-deterministic, but we decided to simplify.

Bonus points if we can explain why the simplification is not problematic!

osmanyasar05: Thanks for the review!
Yes, at first, the implemented semantics were:
```
@[simp_llvm_option]
def freeze (x: IntW w) : IntW w := do
  match x with
    | poison => value (default)
    | value a => value (a)
```
so that we also have non-deterministic semantics, however, this caused issues in instruction lowering. What do you think would be the correct way to implement non-deterministic semantics?

alexkeizer: Note that using `default` rather than `0` doesn't change much, `default` is still deterministic!

In fact, Lean is a pure language, so any function `IntW w -> IntW w` is necessarily deterministic! Thus, the way to model non-determinism is to say that the semantics of `freeze` is *not* such a function, but instead a function that maps an `IntW w` to the *set* of all possible choices of resulting `IntW w`. Or, more likely, you'd change the definition of `IntW w` to be something like `PoisonOr (Set (BitVec w))` to say that the semantic model is a set of possible (non-deterministic) values.

Even that wouldn't be quite right, as in Lean a `Set` is not computable (and we do care about our semantics being computable) so we'd rather use a `FinSet` or something like that.

This would be a complex change with a lot of downstream ripple effects, so I'm certainly not suggesting we make this change at this point, but we should acknowledge that the current semantics of freeze are not accurate

PR #1774, opened 2025-10-20
feat: script for constant matching

osmanyasar05:
This PR introduces a new Python script, `generate_constants.py`, that automates the generation of rewrites requiring constant matching.

- The rewrite patterns that involve constant matching should now be defined in `generate_constants.py.`
- The script automatically generates the corresponding Lean rewrite rules for the specified range of constant values and outputs them to `ConstantMatching.lean`.
- Three example patterns (`sub_to_add`, `mul_to_shl`, `urem_pow2`) are currently implemented to showcase the implementations and demonstrate the script's output.
- The previously incomplete rewrite placeholders in `Combiner.lean` have been removed since their functionality is now covered by the new script.

The script takes no input argument, and can be run by:
`python3 generate_constants.py`


tobiasgrosser:
Nice.

PR #1782, opened 2025-10-24
feat: add cast_combines narrow_binops

osmanyasar05:
This PR adds rewrite patterns requiring constant matching from GICombineGroup [cast_combines](https://github.com/llvm/llvm-project/blob/26db21400d5bfe87e2b3d386c8589b56b965b158/llvm/include/llvm/Target/GlobalISel/Combine.td#L2020).


tobiasgrosser (approved):
LGTM.

luisacicolini:
Thank you for the PR! I see that the opt tool does not build in CI, can you check locally and potentially fix? And also if you could merge main would be great.

PR #1793, opened 2025-10-27
chore: fix argument in generation

luisacicolini:
This PR fixes the `default` argument for the generation of benchmarks and removes the `Os` argument (which is illegal with llc to the best of my knowledge). We also update the `README` accordingly.


luisacicolini:
Will merge once CI passes, given that it's a trivial change limited to a python script.

tobiasgrosser (approved):

PR #1796, opened 2025-10-27
chore: change formatting and range of constant matching in rewrite patterns

osmanyasar05:
This PR reduces the range of constants we generate rewrites for (to [-5, 5], original was [-50, 50]), in addition to formatting changes.


luisacicolini:
lgtm, thank you!

PR #1797, opened 2025-10-27
feat: add integer_reassoc_combines constants

osmanyasar05:
This PR adds the rewrites requiring constant matching in [integer_reassoc_combines](https://github.com/llvm/llvm-project/blob/d7e40f3e71654658e249ee481e28aeae0e0410b1/llvm/include/llvm/Target/GlobalISel/Combine.td#L1845) GICombineGroup. 

The PR also adds a new rewrite group, `GlobalISelPostLegalizerCombinerConstantFolding`, that we can group rewrites that require constant folding together, enabling us to easily exclude these rewrites if necessary.


SSA/Projects/LLVMRiscV/Pipeline/scripts/generate_constants.py
```diff
@@ -395,21 +395,174 @@ def generate_canonicalize_icmp(max_val: int) -> RewriteGroup:
...
+    comment = """
+/-! ### integer_reassoc_combines_constants -/
+
+/-
+Test the rewrite:
+ fold (A+C1)-C2 -> A+(C1-C2)
+ fold C2-(A+C1) -> (C2-C1)-A
+ fold (A-C1)-C2 -> A-(C1+C2)
+ fold (C1-A)-C2 -> (C1-C2)-A
+ fold ((A-C1)+C2) -> (A+(C2-C1))
```
luisacicolini, line 536: maybe a stupid question: aren't there hypotheses on c1 and c2? what happens in llvm if c2 > c1? are we doing the same thing?

osmanyasar05: there are no hypotheses on them. why do you think c2>c1 might cause any issues?

luisacicolini: so depending on whether we're interpreting the operations on natural or integers, Lean can behave differently. If we're in naturals and c2 - c1, the rewrite won't hold. 
and AFAIK Lean by default resorts to interpretinv bitvecs as nats (but I'm not sure about this). @tobiasgrosser could you take a look?

osmanyasar05: the LLVM integers use 2's complement representation, and the implemented LLVM semantics in lean-mlir support that. so I don't think any issues would arise in that case.

luisacicolini: okay that sounds reasonable. could you maybe write a test with c2 > c2? so we document this (I'm being a bit paranoid sorry, but better to test it now)

luisacicolini:
Thank you for the PR! can you please merge main and update the commit message?

luisacicolini:
I will also be very annoying, sorry, but I would like to split the PR and move to another PR all the changes relative to the evaluation, thank you!

osmanyasar05:
> I will also be very annoying, sorry, but I would like to split the PR and move to another PR all the changes relative to the evaluation, thank you!

removed it! @luisacicolini

SSA/Projects/InstCombine/LLVM/Opt.lean
```diff
@@ -52,6 +55,7 @@ def mainCmd := `[Cli|
       passriscv64; "Lowering pass to a RISC-V 64 SSA-assembly IR"
       riscv; "Allows to parse a file as a RISC-V 64 SSA-assembly IR"
       passriscv64_optimized; "Allows to parse a file as a RISC-V 64 SSA-assembly IR"
+      passriscv64_optimized_const; "Allows to parse a file as a RISC-V 64 SSA-assembly IR"
```
luisacicolini: could you add a more specific comment to this? explaining the difference with the others. (We should fix the other ones as well and make them more informative, but let's start with this one)

SSA/Projects/LLVMRiscV/Pipeline/InstructionLowering.lean
```diff
@@ -192,3 +193,52 @@ def selectionPipeFuelWithCSEWithOpt {Γl : List LLVMPlusRiscV.Ty} (fuel : Nat) (
...
+    GLobalISelPostLegalizerCombiner optimize_initial_1;
+  let loweredConst := multiRewritePeephole fuel
+    const_match optimize_initial_2;
+  let lowerPart1 := multiRewritePeephole fuel
+    rewritingPatterns1  loweredConst;
+  let lowerPart2 := multiRewritePeephole fuel
+    rewritingPatterns0 lowerPart1;
+  let postLoweringDCE := (DCE.dce' lowerPart2).val;
+  let postReconcileCast := multiRewritePeephole fuel (reconcile_cast_pass) postLoweringDCE;
+  let remove_dead_Cast1 := (DCE.dce' postReconcileCast).val;
```
luisacicolini: is there a reason why you're not using `repeatDce` here?

osmanyasar05: fixed it!

SSA/Projects/InstCombine/LLVM/Opt.lean
osmanyasar05: rewrote the comment mentioning it includes rewrites with constant matching

PR #1798, opened 2025-10-27
2025-10-27 lean nightly update

tobiasgrosser:
automatic update of mathlib + lean via GitHub action.


tobiasgrosser:
@bollu, this breaks in Blase.

ineol:
This should fix the build.

@alexkeizer I changed the `simp_peephole` tactic because they removed `Id.pure_eq` and `Id.bind_eq` so you make want to check I didn't make a mistake.

alexkeizer:
> This should fix the build.
> 
> @alexkeizer I changed the `simp_peephole` tactic because they removed `Id.pure_eq` and `Id.bind_eq` so you make want to check I didn't make a mistake.

I believe we copied these rewrites as `Id.pure_eq'` and `Id.bind_eq'`. We do need these rewrites in `simp_peephole`, I expect just removing them would break the evaluation!

EDIT: @ineol, did you push this change? I don't see it in the diff

ineol:
@alexkeizer somehow [this](https://github.com/opencompl/lean-mlir/compare/dd5ff9e732caa7df3013bff1c3a85e422ad25da7..0a77d925de149b230bc795de7103b0a7e07811a0) overrode my changes

ineol:
Repushed

ineol:
I think @tobiasgrosser's automated process is undoing all my changes 😅

PR #1799, opened 2025-10-27
Blase: preprocess equality of propositions

ineol:
@bollu that could help?

Is there a simple way to evaluate this change?


bollu:
@ineol Hit merge and we can rerun the eval :D

PR #1803, opened 2025-10-29
feat: evaluation of ratios and similarity

osmanyasar05:
This PR adds support for analyzing similarity between programs with the same number of instructions and visualization for the proportional difference between different methods. The script also outputs the numbers in LaTeX format.


luisacicolini:
thank you! could you add to the pr message or to a comment an example of the plot this PR helps generating? :)

osmanyasar05:
image

This graph shows the proportional difference between CertOptISel and GlobalISel. GlobalISel serves as the baseline and normalized to 1, and the bars represent CertOptISel's performance relative to it. For example, CertOptISel needed 39% more cycles in programs with 6 instructions. This graph is generated for all combinations of parameters and ISel options.

luisacicolini:
LGTM, thank you for the PR!

PR #1805, opened 2025-10-29
fix: use `func.func` for register allocation

luisacicolini:
This PR introduces a `create_func` script that converts the output of Lean-MLIR's instruction selection to a `func.func` module, that XDSL can correctly register-allocate. 
This ensure the correctness of the assembly we produce. 

All python scripts are ruff formatted and checked.


luisacicolini:
Will merge once CI passes.

PR #1808, opened 2025-10-29
feat: add mulh_to_lshr rewrite pattern

osmanyasar05:
This PR adds the rewrite pattern [mulh_to_lshr](https://github.com/llvm/llvm-project/blob/e65b36c640b781be724f587085056c2819c7ad00/llvm/include/llvm/Target/GlobalISel/Combine.td#L1289), which transforms a mulh by pow2 operation into a right shift operation. 

The PR also moves the RISCV Peephole rewrite structures into `PeepholeRefine.lean` file, enabling ConstantMatching.lean to use the definition. Without this change, both `ConstantMatching.lean` and `Combiners.lean` files imported each other, resulting in circular import dependency.

### DIFF
--- a/SSA/Projects/LLVMRiscV/Pipeline/Combiners.lean
+++ b/SSA/Projects/LLVMRiscV/Pipeline/Combiners.lean
@@ -14,20 +14,6 @@ open LLVMRiscV
   while some are target-dependent.
 -/
 
-@[simp_riscv] lemma toType_bv : TyDenote.toType (Ty.riscv (.bv)) = BitVec 64 := rfl
-@[simp_riscv] lemma id_eq1 {α : Type} (x y : α) :  @Eq (Id α) x y = (x = y):= by simp only
-
-structure RISCVPeepholeRewrite (Γ : List Ty) where
-  lhs : Com LLVMPlusRiscV Γ .pure [Ty.riscv .bv]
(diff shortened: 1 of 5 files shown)


SSA/Projects/LLVMRiscV/Pipeline/Combiners.lean
```diff
@@ -14,20 +14,6 @@ open LLVMRiscV
...
-  lhs : Com LLVMPlusRiscV Γ .pure [Ty.riscv .bv]
-  rhs : Com LLVMPlusRiscV Γ .pure [Ty.riscv .bv]
-  correct : lhs.denote = rhs.denote := by simp_lowering <;> bv_decide
-
-def RISCVPeepholeRewriteToRiscvPeephole (self : RISCVPeepholeRewrite Γ) :
-    PeepholeRewrite LLVMPlusRiscV Γ [Ty.riscv .bv] where
-  lhs := self.lhs
-  rhs := self.rhs
-  correct := self.correct
-
```
luisacicolini, line 30: I don't understand this change, why did it get moved?

osmanyasar05: answered in the pr message

luisacicolini:
thank you, lgtm, but I would appreciate if @salinhkuhn or @tobiasgrosser could also take a look and sanity check :)

luisacicolini (approved):
let's merge!

PR #1809, opened 2025-10-29
chore: add axiom for the correctness of refinement in rewrite patterns 

luisacicolini:
We add an axiom for the rewrites whose correctness is proven up-to-refinement.


tobiasgrosser:
This looks like a good idea.

PR #1816, opened 2025-10-31
feat: Add truncate to list of expressions that bv_generalize knows

bollu:
OK, bv_generalize is super flaky, let's not use it for the time being, I am afraid.


bollu:
Chore: superceded by https://github.com/opencompl/lean-mlir/pull/1821

PR #1818, opened 2025-11-01
feat: add new conversion patterns for sext/zext

osmanyasar05:
This PR extends the RISC-V lowering logic for LLVM `llvm.sext`, `llvm.zext` and `llvm.trunc` operations. It replaces manual bit manipulation with dedicated RISC-V instructions/pseudo-instructions (`sext.b`, `sext.h`, `sext.w`, `zext.b`, `zext.w`), and adds support for 32-to-64 bit extension cases, using lowerings from llc 21.1.0, the newest llc version.

The changes also introduce the new [zext.w ](https://five-embeddev.com/riscv-bitmanip/1.0.0/bitmanip.html#insns-add_uw) operation to the RISC-V semantics.

### DIFF
--- a/SSA/Projects/LLVMRiscV/Pipeline/trunc.lean
+++ b/SSA/Projects/LLVMRiscV/Pipeline/trunc.lean
@@ -9,6 +9,7 @@ open LLVMRiscV
   This file implements the lowering for the `llvm.trunc` instruction for truncation:
   - from i32 to i8
   - from i32 to i16
+  - from i64 to i1
   - from i64 to i32
 -/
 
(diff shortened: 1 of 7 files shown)


SSA/Projects/LLVMRiscV/Pipeline/trunc.lean
```diff
@@ -9,6 +9,7 @@ open LLVMRiscV
   This file implements the lowering for the `llvm.trunc` instruction for truncation:
   - from i32 to i8
   - from i32 to i16
+  - from i64 to i1
```
salinhkuhn: nit: Is there a reason we do not order it decreasingly or increasingly ,e.g. first to i1 up on till i32 ? Or align it with the order in which they appear bellow maybe ?

osmanyasar05: It seemed ugly to me as well, but the logic is that we first order by source type (i32 then i64), and then the destination type.

osmanyasar05: i put them in the order they appear in the code now.

PR #1824, opened 2025-11-03
feat: add SelectionDAG optimizations patterns

osmanyasar05:
This PR introduces a new RISC-V 64 lowering pass that applies SelectionDAG optimization patterns to both LLVM IR. The main changes include verification of optimization rewrites from SelectionDAG and the implementation of the corresponding pipeline.

List of implemented patterns:

- [visitADD](https://github.com/llvm/llvm-project/blob/bc0d0cf3ac9b5f1c2ec33d22aba8e7ece7d08ed2/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L3175)
- [visitAND](https://github.com/llvm/llvm-project/blob/bc0d0cf3ac9b5f1c2ec33d22aba8e7ece7d08ed2/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L7457)
- [visitMUL](https://github.com/llvm/llvm-project/blob/bc0d0cf3ac9b5f1c2ec33d22aba8e7ece7d08ed2/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L4702)
- [visitOR](https://github.com/llvm/llvm-project/blob/bc0d0cf3ac9b5f1c2ec33d22aba8e7ece7d08ed2/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L8391)
- [visitSDIV](https://github.com/llvm/llvm-project/blob/bc0d0cf3ac9b5f1c2ec33d22aba8e7ece7d08ed2/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L5100)
- [visitSELECT](https://github.com/llvm/llvm-project/blob/bc0d0cf3ac9b5f1c2ec33d22aba8e7ece7d08ed2/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L12354)
- [visitSUB](https://github.com/llvm/llvm-project/blob/bc0d0cf3ac9b5f1c2ec33d22aba8e7ece7d08ed2/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L4126)
- [visitSRA](https://github.com/llvm/llvm-project/blob/bc0d0cf3ac9b5f1c2ec33d22aba8e7ece7d08ed2/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L10926)
- [visitXOR](https://github.com/llvm/llvm-project/blob/bc0d0cf3ac9b5f1c2ec33d22aba8e7ece7d08ed2/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L9936)

### DIFF
--- a/SSA/Projects/InstCombine/LLVM/Opt.lean
+++ b/SSA/Projects/InstCombine/LLVM/Opt.lean
@@ -40,6 +40,9 @@ def runMainCmd (args : Cli.Parsed) : IO UInt32 := do
   if args.hasFlag "passriscv64_optimized" then -- lowering pass to a RISC-V 64 SSA-assembly IR
     let code ← passriscv64_optimized fileName
     return code
+  if args.hasFlag "passriscv64_selectiondag" then -- lowering pass to a RISC-V 64 SSA-assembly IR applying optimizations from SelectionDAG
+    let code ← passriscv64_selectiondag fileName
+    return code
   else
     let code ← wellformed fileName
     return code
@@ -52,6 +55,7 @@ def mainCmd := `[Cli|
       passriscv64; "Lowering pass to a RISC-V 64 SSA-assembly IR"
       riscv; "Allows to parse a file as a RISC-V 64 SSA-assembly IR"
       passriscv64_optimized; "Allows to parse a file as a RISC-V 64 SSA-assembly IR"
+      passriscv64_selectiondag; "Allows to parse a file as a RISC-V 64 SSA-assembly IR using SelectionDAG optimization patterns"
     ARGS:
       file: String; "Input filename"
     ]
--- a/SSA/Projects/InstCombine/LLVM/Opt.lean
+++ b/SSA/Projects/InstCombine/LLVM/Opt.lean
@@ -40,7 +40,7 @@ def runMainCmd (args : Cli.Parsed) : IO UInt32 := do
   if args.hasFlag "passriscv64_optimized" then -- lowering pass to a RISC-V 64 SSA-assembly IR
     let code ← passriscv64_optimized fileName
     return code
-  if args.hasFlag "passriscv64_selectiondag" then -- lowering pass to a RISC-V 64 SSA-assembly IR applying optimizations from SelectionDAG
+  if args.hasFlag "passriscv64_selectiondag" then -- lowering pass to RISC-V 64 SSA-assembly IR applying optimizations from SelectionDAG
     let code ← passriscv64_selectiondag fileName
     return code
   else
@@ -55,7 +55,7 @@ def mainCmd := `[Cli|
       passriscv64; "Lowering pass to a RISC-V 64 SSA-assembly IR"
       riscv; "Allows to parse a file as a RISC-V 64 SSA-assembly IR"
       passriscv64_optimized; "Allows to parse a file as a RISC-V 64 SSA-assembly IR"
-      passriscv64_selectiondag; "Allows to parse a file as a RISC-V 64 SSA-assembly IR using SelectionDAG optimization patterns"
+      passriscv64_selectiondag; "Lowering pass to RISC-V 64 SSA-assembly IR applying optimizations from SelectionDAG"
     ARGS:
       file: String; "Input filename"
     ]
--- a/SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
+++ b/SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
@@ -1,0 +1,957 @@
+import SSA.Projects.LLVMRiscV.PeepholeRefine
+import SSA.Projects.LLVMRiscV.Simpproc
+import SSA.Projects.RISCV64.Tactic.SimpRiscV
+import SSA.Projects.LLVMRiscV.Pipeline.mkRewrite
+import SSA.Projects.LLVMRiscV.Pipeline.ConstantMatching
+
+open LLVMRiscV
+
+/- This file implements `DAGCombiner` patterns extracted from the LLVM Risc-V backend.
+  First, we implement the Lean structure that implements the rewrite patterns and then we implement
+  optimizations for LLVM IR and RISC-V.
+  In particular, we implement the patterns supported by LLVM's `GlobalIsel` for RISC-V.
+  Because `GlobalIsel` is hybrid, some of these patterns regard generic IR,
+  while some are target-dependent.
+-/
+
+@[simp_riscv] lemma toType_bv : TyDenote.toType (Ty.riscv (.bv)) = BitVec 64 := rfl
+@[simp_riscv] lemma id_eq1 {α : Type} (x y : α) :  @Eq (Id α) x y = (x = y):= by simp only
+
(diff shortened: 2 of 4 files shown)


luisacicolini:
`opt` does not seem to build, does it work locally?

SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
```diff
@@ -0,0 +1,957 @@
...
+    ^entry (%x: i1):
+      %0 = llvm.not %x : i1
+      %1 = llvm.zext %0 : i1 to i64
+      llvm.return %1 : i64
+  }]
+
+/-
+  (and x, -1) -> x,
+-/
+def visitADD_minus1 : LLVMPeepholeRewriteRefine 64 [Ty.llvm (.bitvec 64)] where
```
luisacicolini: rather than `minus`, we usually use `neg` in lean for the nomenclature with neg numbers (unless it's called like this somewhere in llvm?)

luisacicolini: (same below)

SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
```diff
@@ -0,0 +1,957 @@
...
+    ^entry (%a: i64, %b: i64):
+      %0 = llvm.sub %b, %a : i64
+      llvm.return %0 : i64
+  }]
+
+/-
+Test the rewrite:
+  fold (A+(0-B)) -> A-B
+-/
+def APlusZeroMinusB : LLVMPeepholeRewriteRefine 64 [Ty.llvm (.bitvec 64), Ty.llvm (.bitvec 64)] where
```
luisacicolini: rather than plus/minus in Lean we use `add` and `sub`, so this sould be `add_zero_sub` or something like that (unless it's called `APlusZeroMinusB` somewhere in LLVM?)

luisacicolini:
I suggested some changes for the naming conventions, other than that looks good to me!

SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
osmanyasar05: yes, it is called APlusZeroMinusB in GlobalISel indeed. But it does not have a name in SelectionDAG. We should change this one regardless since that's causing the build to fail due to multiple defs with same name. Thanks!

SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
osmanyasar05: fixed

SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
```diff
@@ -0,0 +1,957 @@
...
+    ^entry (%a: i64, %b: i64, %c: i64):
+      %0 = llvm.sub %b, %c : i64
+      llvm.return %0 : i64
+  }]
+
+/-
+Test the rewrite:
+ fold (add (xor a, -1), 1) -> (sub 0, a)
+-/
+def visitADD_XorNeg1Plus1 : LLVMPeepholeRewriteRefine 64 [Ty.llvm (.bitvec 64)] where
```
luisacicolini, line 207: being nitpicky but let's be consistend, can we have the `_` in between the ops as you did for hte others? `xor_neg_1_plus_1`

osmanyasar05: I think we should do visitXXX_{unique_name} for all. as in keeping it like visitADD_XorNeg1Plus1.

 what do you think? or you prefer visitXXX_op1_op2_...? like visitADD_xor_neg_1_plus_1?

SSA/Projects/InstCombine/LLVM/Opt.lean
```diff
@@ -52,6 +55,7 @@ def mainCmd := `[Cli|
       passriscv64; "Lowering pass to a RISC-V 64 SSA-assembly IR"
       riscv; "Allows to parse a file as a RISC-V 64 SSA-assembly IR"
       passriscv64_optimized; "Allows to parse a file as a RISC-V 64 SSA-assembly IR"
+      passriscv64_selectiondag; "Allows to parse a file as a RISC-V 64 SSA-assembly IR using SelectionDAG optimization patterns"
```
salinhkuhn: NIT: I am not sure about the descriptiton. I would make it clear that you parse into a SSA-assembly and not parse an assembly itself because now I read it as if the input would be a Risc-V assembly already.

SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
salinhkuhn: Yes I agree with  @osmanyasar05  on  visitXXX_{unique_name}

SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
```diff
@@ -0,0 +1,939 @@
...
+    ^entry (%a: i64, %b: i64):
+      %0 = llvm.sub %b, %a : i64
+      llvm.return %0 : i64
+  }]
+
+/-
+Test the rewrite:
+ (x - y) + -1  ->  add (xor y, -1), x
+-/
+def visitSUBPlusNeg1 : LLVMPeepholeRewriteRefine 64 [Ty.llvm (.bitvec 64), Ty.llvm (.bitvec 64)] where
```
salinhkuhn: I think a _ is missing in the name

SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
```diff
@@ -0,0 +1,939 @@
...
+
+def visitSELECT : List (Σ Γ, LLVMPeepholeRewriteRefine 64 Γ) :=
+  [⟨_, visitSELECT_constant_not_cond⟩,
+   ⟨_, visitSELECT_constant_cmp_true⟩,
+   ⟨_, visitSELECT_constant_cmp_false⟩]
+
+ /-! ### Grouped patterns -/
+
+def SelectionDAG_combines : List (Σ Γ, LLVMPeepholeRewriteRefine 64 Γ) :=
+  visitADD ++ visitAND ++ visitSUB ++ visitMUL ++ visitSDIV ++ visitOR ++ visitXOR ++ visitSRA ++
```
salinhkuhn: nit: do we want to align visitXOR meaning also indent it to match the start of the upper line ?

SSA/Projects/LLVMRiscV/Pipeline/SelectionDAG.lean
luisacicolini: anything works for me, as long as the naming remains consistent

PR #1833, opened 2025-11-05
feat: update plotting scripts for evaluation of isel

osmanyasar05:
We update the plotting scripts to: 
- use a fixed version of xdsl
- only compare `O2` (default) optimization level


luisacicolini (approved):

luisacicolini:
Given that the change only touches printing and plotting, I will merge without review.

PR #1834, opened 2025-11-06
blase support for new ops

bollu:
- **feat: add shiftl, setWidth**
- **feat: add setWidth, shiftl to tactic infra**


bollu:
@ineol this should create much smaller goal states in the SMT-LIB printer.

PR #1839, opened 2025-11-06
chore: add +1, +k support for widths

bollu:
This will allow us to solve problems in ROVER that involve width `v >= w` which is lower to `v + 1 > w`.


bollu:
This should hopefully allow ROVER examples to succeed.

PR #1841, opened 2025-11-07
feat: retrieve `RISCV64` semantics from `riscv-lean`

luisacicolini:
We update the semantics of the `RISCV64` dialect to use the one derived from the Sail model (opencompl/riscv-lean)


tobiasgrosser:
Thank you, @luisacicolini.

PR #1845, opened 2025-11-09
chore: try to understand why we fail in ROVER

bollu:


bollu:
I did not understand subtle hardware semantics (all code to repro what I've written is in the gist below): 

https://gist.github.com/bollu/eab2dbaaf1fd68ab1d8a142b155a637e

--- 

It seemed to me that I had found a  counter-example to the ROVER rewrite:

```
#  {
#    "name": "add_assoc_1",
#    "preconditions": ["(>= q t)", "(>= u t)"],
#    "lhs": "(bw t ( + (bw u (+ (bw p a) (bw r b))) (bw s c)))",
#    "rhs": "(bw t ( + (bw p a) (bw q (+ (bw r b) (bw s c)))))"
#  },
```

where the counter-example is of the form:

```
FAIL: ((a:[1#1] + b:[1#1]).bw(u:4) + c:[15#4]).bw(t:2) != a:[1#1] + (b:[1#1] + c:[15#4]).bw(q:2).bw(t:2)
      a:[1#1] + b:[1#1] = [0#1]
      (a:[1#1] + b:[1#1]).bw(u:4) = [0#4]
      (a:[1#1] + b:[1#1]).bw(u:4) + c:[15#4] = [15#4]
      (a:[1#1] + b:[1#1]).bw(u:4) + c:[15#4]).bw(t:2) = [3#2]
  --
      b:[1#1] + c:[15#4] = [0#4]
      (b:[1#1] + c:[15#4]).bw(q:2) = [0#2]
      a:[1#1] + (b:[1#1] + c:[15#4]).bw(q:2) = [1#2]
      a:[1#1] + (b:[1#1] + c:[15#4]).bw(q:2).bw(t:2) = [1#2]
```

My dumdum was to assume that the semantics of BV add was to extend to the *max* of the bitwidths, and then perform the add. However, after analyzing the counter-example, this seems wrong. The intended semantics seems to be an extension to "as much bitwidth is necessary to ensure that *no overflow* occurs". So, for the case of addition, we need to extend to `max(a, b) + 1`. Indeed, this is confirmed by setting the flag `EXTEND_TO_ENSURE_NO_OVERFLOW = True`. 



--- 

I wonder, given the rewrite statement:

```
(bw t ( + (bw u (+ (bw p a) (bw r b))) (bw s c)))
```

what exactly should I interpret the semantics of `+`, `*`, etc? Is there some standard reference I can cite for these so I don't make an ouchie? I did skim the ROVER paper, but was unable to find a precise defn of this format. My guess are:

- `+`: extend to `max(v, w) + 1`
- `*`: extend to `max(v, w) * 2`.
- I haven't thought about the shifts.

bollu:
superceded by https://github.com/opencompl/lean-mlir/pull/1847

PR #1847, opened 2025-11-10
fix rover widths

bollu:
- **chore: try to understand why we fail to no success**
- **chore: set rover widths to max**
- **chore: fix rover widths**


bollu:
supercedes #1845 , as it takes the insight from there and makes a clean patch.

PR #1851, opened 2025-11-10
feat: rover automata state space creation

bollu:
This makes the FSM state space to be `Fin n`, which allows for easy hashing, and might allow ROVER to solve


bollu:
Nope, this does not help, sadly, so I give up on this approach. Our best solution maybe to have  https://github.com/opencompl/lean-mlir/pull/1853

PR #1879, opened 2025-11-17
blase unify sorts

bollu:
- **chore: add Prop sort**
- **chore: unify all sorts**
- **chore: unify**


bollu:
We've now merged `Predicate` into `Term`, which should let us describe `cast`, `ofBool`, `decide` in the core language.

PR #1899, opened 2025-11-21
feat: our first `seq` program

luisacicolini:
This PR adds the circuit implemented with @ymherklotz and @wkrozowski, with the lowering of `Handshake.fork`. 

Co-authored-by: @wkrozowski  
Co-authored-by: @ymherklotz


SSA/Projects/CIRCT/Handshake-to-HW/fork.lean
```diff
@@ -0,0 +1,218 @@
...
+    fun n =>
+      {toArray := [a n, b n].toArray, size_toArray := by simp}
+
+
+
+/--
+  loop in:
+  https://docs.google.com/drawings/d/1ymKZgQxT9D3O9SNsK_9duNYM88miSwUYrdpusvCB2hE/edit
+-/
+def loop
```
wkrozowski: I don't fully understand the signature of this function. So, we take feedback input as inputs, but we don't output an updated feedback? I am just struggling to see if `loop` implements the body of the loop and one would need to put it inside of a feedback operator, or is it the implementation of the whole loop (with the feedback behaviour).

luisacicolini: my understanding is that loop should implement the whole circuit, with all the input/output/feedback loops (the output being the result of %12)

wkrozowski: Makes sense to me, but why do we take feedback as an input then?

luisacicolini: oh I see your point. given that we only use the feedback loops inside the loop itself we don't need them, good point! I'll fix that

luisacicolini:
will merge without review once CI is done.

PR #1911, opened 2025-11-29
feat: add Valaig package for Aiger solving in Blase

### LINKED ISSUE #1906: blase/cav26: Allow invoking an AIGER solver (e.g. rIC3 from within Lean)

georgerennie:
Fixes #1906. This adds Valaig as a dependency and replaces the `debugDumpAiger` config option with `solveRIC3`. There are many sorries still on this code path and certificates aren't checked yet, but this should at least allow getting some performance numbers. I'm not particularly convinced yet that there aren't bugs here, so please let me know if you see weird behaviour!

This assumes you have `rIC3` installed an on your path, and defaults to solving with the single threaded `ic3` engine currently.


Blase/Blase/MultiWidth/Tactic.lean
```diff
@@ -1123,12 +1123,19 @@ def solve (gorig : MVarId) : SolverM Unit := do
...
+      let ric3 : Valaig.External.rIC3 := {}
+      let res ← Valaig.External.checkSafety ric3 aig
+
+      match res with
+      | .counterexample => throwError "CEX: rIC3 found a counter-example"
+      | .unknown => throwError "UNKNOWN: rIC3 returned unknown status code"
+      | .proof =>
+        let prf ← mkSorry (synthetic := true) (← g.getType)
+        let _ ← g.apply prf
+        return
```
bollu: Could this be an if-then-else please? It makes control flow easier to understand wrt tactics to have `if-then-else` in my opinion.

Blase/Blase/Fast/Aiger.lean
```diff
@@ -1,101 +1,74 @@
...
+    latches,
+    bads := #[{ lit := .ofRef bad }]
+    -- We only use cached constant constructor so there are no other falses
+    hfalse := sorry
+    -- We construct the inputs/latches arrays directly as we add them and only
+    -- use cached constructors so atoms only appear once
+    hinputstodecl := sorry
+    hdecltoinputs := sorry
+    hlatchestodecl := sorry
+    hdecltolatches := sorry
```
bollu: Could we change the `sorry`s to an axiom usage?

bollu:
I asked for two changes. Overall, this is fantastic! I want to use `if-then-else` instead of early returns to make control flow super duper obvious, and secondly, I want to not use `sorry`, and instead just use `axiom`s.

georgerennie:
I've pushed a new version that uses a load of changes to Valaig, the main differences being:

- There are now no sorries in the lowering to Aiger! This is just proofs of structural properties of the Aig however, the semantic preservation is still to come
- The conversion from Blase's representation to Valaig's is simpler, using a new `fromStd` method in Valaig that just requires Blase to provide information about which atoms are inputs/latches
- The config option has changed to `externalSolver` to enable an external solver, and `externalSolverConfig` to set the config. This currently defaults to rIC3's ic3 engine with a 5s timeout and requires rIC3 to be installed on the machine. `externalSolver` defaults to false so must be set to true to enable this

bollu:
Awesome! @georgerennie, hit merge :)

PR #1919, opened 2025-12-10
feat: add more default targets to make opt, LLVMRisCV, mlirnatural part of base build so this is cached

bollu:
Investigated by looking at CI runner log @
https://github.com/opencompl/lean-mlir/actions/runs/20103919721/job/57682362336.

This shows that we spend 3min + 6min + 1min building targets that ought
to be cached in the base layer.

To ensure caching, we instead add these as default targets, so the
toplevel `lake -R build` builds all of these.


tobiasgrosser:
This looks good to me.

tobiasgrosser (approved):
LGTM. I enabled automerge.

PR #1922, opened 2025-12-19
feat: Neater solver config option for external solvers in Blase

georgerennie:
From a suggestion from @bollu, this is a bit nicer way of selecting between the kinduction/external solver than having a separate enable and config field and lets us do `solver := .external` to get a default external config


bollu:
Solid, thanks muchly for the refactor. Please do hit merge!

PR #1928, opened 2026-01-15
feat: add `Dockerfile` for RISC-V ISA paper

luisacicolini:


alexkeizer:
We spent a lot of effort layer and cache-optimizing the base lean-mlir image, let's use that as a base for this image, instead of rebuilding lean-mlir from scratch!

See [lean-mlir-instcombine](https://github.com/opencompl/lean-mlir/blob/main/bv-evaluation/Dockerfile) for an example of a Docker image that builds upon the lean-mlir image and installs a bunch of extra dependencies without destroying the layering/caching.

luisacicolini:
thanks for the review @alexkeizer - actually I will close this PR as the last conclusion we came to with @tobiasgrosser was to have a separate repo for the artifact. I'll anyways use the file you suggested as a base in the right repo!

PR #1934, opened 2026-02-02
feat: Automate GlobalISel Pattern Translation 

osmanyasar05:
This PR adds a Python script (`tablegen_to_lean.py`) that automatically translates pure MIR pattern combine rules from `Combine.td` into Lean-MLIR.


SSA/Projects/LLVMRiscV/Pipeline/mir_combines.lean
```diff
@@ -0,0 +1,498 @@
+import SSA.Projects.LLVMRiscV.PeepholeRefine
+import SSA.Projects.LLVMRiscV.Simpproc
+import SSA.Projects.RISCV64.Tactic.SimpRiscV
+import SSA.Projects.LLVMRiscV.Pipeline.mkRewrite
+import SSA.Projects.LLVMRiscV.Pipeline.ConstantMatching
+
+open LLVMRiscV
+
```
luisacicolini, line 9: afaict this is generated by the script above, could you write it in a comment at the beginning of the file? ideally the generator will print a comment saying this

luisacicolini: and also maybe specifying what combine/gisel file it was taken from

luisacicolini (approved):
Thank you very much for the PR :) I have two very small comments but other than that LGTM

SSA/Projects/LLVMRiscV/Pipeline/mir_combines.lean
osmanyasar05: added these to the beginning of the file, thanks:)

PR #1941, opened 2026-02-13
feat: write multi-width to single-width pass

bollu:
We try to use the approach outlined yesterday, to automatically derive the corresponding single-width statement for a multi-width problem.


bollu:
Superceded by #1947

PR #1947, opened 2026-02-24
feat: multi -> single width [2/?]

bollu:
This PR adds support for the multi widh to single width reduction
that produces the predicate which encodes the reduction.


bollu:
CC @ineol : Does the overall structure make sense? It'd be nice to chat with you today about the approach. I've written the skeleton of the translation, as well as the theorem statements, and it would be nice to get a "yes I beleive this"

PR #1957, opened 2026-03-01
chore/blase: cover operations ashr, distinct, lhsr, udiv, urem, vshl 

bollu:
This adds the missing ops as requested by @georgerennie , which should let us run a complete eval against the PBV sources.


ineol (approved):
LGTM

PR #1968, opened 2026-03-10
Implement missing blase ops

bollu:
feat: implement extract, and add support for eliminating subtractions and if-then-elses


bollu:
Supereceded, implemented in Blase now

PR #1980, opened 2026-04-03
experimental: toSingleWidth alternate definition

digama0:
cf. discussion with @bollu


bollu:
@digama0 Thanks! I was wondering if you could write a sketch of the meta-code for this? I wanted a sense of how you'd write the "translate binder-at-a-time"

